PR #2, opened 2015-11-10
Add imagebuilding step to the README

benbp:
Making it more explicit how to generate the discovery images from scratch, and when in the process to do them.

@heckj @DavidjohnBlodgett @clintonskitson

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -6,14 +6,6 @@
 
   We expect the latest version of GIT, Vagrant, and Ansible installed onto the host system.
 
-  We expect static files to be located in the correct path from the parent directory of dev-tools:
-
-    i.e.
-        ~/<repos directory>/RackHD/on-http/static/http/common/
-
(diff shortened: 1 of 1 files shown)


example/README.md
```diff
@@ -36,14 +28,21 @@
...
+  Build the discovery images from the monorail server (or any debian-based system):
+
+    $ sudo apt-get install -y ansible
+    $ git clone https://github.com/rackhd/on-imagebuilder.git
+    $ cd on-imagebuilder
+    $ sudo ansible-playbook -i hosts all.yml
+
+  Place the discovery images into HTTP server's static file directory:
+
+    $ mv /tmp/on-imagebuilder/builds/* ~/<repos directory>/RackHD/on-http/static/http/common/
```
DavidjohnBlodgett: if this is build inside the VM then we actually want to move them to the on-http repo through /home/vagrant/src/on-http/static/http/common/

PR #5, opened 2015-11-13
adding file replication from builds into bintray

heckj:
NOTE: names of these files vary around a bit, we may need to massage the
default workflows to match the relevant filenames for the demo setup

Total "vagrant up" time with bintray downloads enable = 20 min (from my coffee shop & laptop)


heckj:
related PR in documentation: https://github.com/RackHD/docs/pull/147

PR #8, opened 2015-11-16
Rename ipxe images to match on-dhcp-proxy bootfile name

jlongever:
Include EFI images

@RackHD/corecommitters @RackHD/rackhd_dev

### DIFF
--- a/example/roles/repos/tasks/main.yml
+++ b/example/roles/repos/tasks/main.yml
@@ -60,5 +60,7 @@
   get_url: url=https://bintray.com/artifact/download/rackhd/binary/ipxe/{{ item }}
            dest=/home/vagrant/src/on-tftp/static/tftp/{{ item }}
   with_items:
-   - ipxe.pxe
+   - monorail.ipxe
+   - monorail-efi32-snponly.efi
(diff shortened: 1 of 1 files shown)


DavidjohnBlodgett:
could this be the root cause of your issues on discovery with our current example automation process? @heckj @benbp

benbp:
:+1: seems likely @DavidjohnBlodgett

heckj:
@benbp sorry, not likely because the ipxe bootloaders are checked into source control right now: (https://github.com/RackHD/on-tftp/tree/master/static/tftp)

DavidjohnBlodgett:
I double checked @rolandpoulter 's environment and he found his /common dir to be empty. I have added sudo args to all ansible steps added for static files and we're seeing if this resolves any write permission issues.

benbp:
If you ran with -vvv did you get any messages about why it wasn't copied? Should the run fail if it can't copy anyways?

benbp:
Oh sorry @heckj I was ignorant of the PR context, I though @DavidjohnBlodgett was talking about discovery images.

PR #9, opened 2015-11-17
fix for static files not being written to common dir

DavidjohnBlodgett:
quick fix for files not appearing in expected locations as seen in @rolandpoulter 's env.

### DIFF
--- a/example/roles/repos/tasks/main.yml
+++ b/example/roles/repos/tasks/main.yml
@@ -46,6 +46,7 @@
   shell: 'npm run apidoc'
   args:
     chdir: /home/vagrant/src/on-http
+  sudo: yes
 
 - name: retrieve microkernel and overlays from bintray
   get_url: url=https://bintray.com/artifact/download/rackhd/binary/builds/{{ item }}
@@ -55,6 +56,7 @@
    - discovery.overlay.cpio.gz
    - initrd.img-3.13.0-32-generic
    - vmlinuz-3.13.0-32-generic
+  sudo: yes
 
 - name: retrieve bootloaders from bintray
   get_url: url=https://bintray.com/artifact/download/rackhd/binary/ipxe/{{ item }}
@@ -64,3 +66,4 @@
    - monorail-efi32-snponly.efi
    - monorail-efi64-snponly.efi
    - undionly.kpxe
+  sudo: yes


DavidjohnBlodgett:
@benbp

benbp:
What are the permissions on that directory in the first place? Can't we just chmod it instead of having to use sudo?

DavidjohnBlodgett:
vagrant@vagrant-ubuntu-trusty-64:~$ ls -all
total 48
drwxr-xr-x 6 vagrant vagrant 4096 Nov 16 23:52 .
drwxr-xr-x 4 root    root    4096 Nov 16 22:30 ..
-rw------- 1 vagrant vagrant  100 Nov 16 23:52 .bash_history
-rw-r--r-- 1 vagrant vagrant  220 Apr  9  2014 .bash_logout
-rw-r--r-- 1 vagrant vagrant 3637 Apr  9  2014 .bashrc
drwx------ 2 vagrant vagrant 4096 Nov 16 22:30 .cache
-rw-r--r-- 1 root    root     176 Nov 16 22:57 .env
drwxr-xr-x 7 root    root    4096 Nov 16 22:51 n
-rw-r--r-- 1 root    root     640 Nov 16 22:57 Procfile
-rw-r--r-- 1 vagrant vagrant  675 Apr  9  2014 .profile
drwxrwxr-x 1 vagrant vagrant 4096 Nov 16 22:28 src
drwx------ 2 vagrant vagrant 4096 Nov 16 22:30 .ssh
vagrant@vagrant-ubuntu-trusty-64:~$


it looks like there should be no issue... but rolands system had nothing populating his common dir, and I got stuff to populate... to be honest at this time it's felt like ansible black magic... but I can continue to dig and try to provide a better answer

DavidjohnBlodgett:
for better context all commands are run from the monorail server and files are placed in on-http through /src

benbp:
Yeah I mean the directory permissions of the common directory, not `~`. Is is that common doesn't exist on his machine but does on yours?

DavidjohnBlodgett:
so I checked the /common dir created by the sudo command issued in the below role and it seems to give open door permissions

https://github.com/RackHD/RackHD/blob/master/example/roles/repos/tasks/main.yml#L41

The only thing I can think of is maybe the host system could some how block the vagrant user trying to write to that directory because it does reside on the host... and by switching to root it tricks the host into allowing write access? that's a far stretch but maybe.

DavidjohnBlodgett:
I could also add a mode flag to that step and try... depending on how deep down the rabbits whole we'd like to take it.

benbp:
makes sense

benbp:
@DavidjohnBlodgett yeah if you create the directory with `sudo: yes` then it's only going to be accessible by the root user.

benbp:
Closing because it seems the problem is caused by the directory being created by the root user in the first place. Feel free to re-open if I'm mistaken.

PR #14, opened 2015-11-20
adding coreos install media

heckj:
Adding ansible tasks to pull down and install CoreOS installation
media (since it's small) to provide an example OS installation flow
with RackHD. Maps to the location and flow from the install-coreos.ipxe
profile (in on-http) and the "Task.Os.Install.CoreOS" workflow task in
on-tasks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-dev-tools/.vagrant
-dev-tools/config/monorail_rack.cfg
-dev-tools/bin/pxe*
+example/.vagrant
+example/config/monorail_rack.cfg
+example/bin/pxe*
--- a/example/README.md
+++ b/example/README.md
@@ -1,62 +1,96 @@
 ## DOCUMENTATION
 
-  The monorail_rack setup script is an easy "one button push" script to deploy an virtual rack within virtualbox to emulate a monorail server and some number of virtualbox PXE-booting clients. The enviornment is tied together using a virtual network called closednet set to our defualt subnet of 172.31.128.x for servicing DHCP and TFTP to the PXE clients.
+The monorail_rack setup script is an easy "one button push" script to deploy
+a 'virtual rack' using virtualbox. This emulates a RackHD server and some number
+of virtual servers - using virtualbox PXE-booting VMs. Private virtual networks
+simulate the connections between servers that would otherwise be on a switch
+in a rack.
 
-## PRE-REQS
+The virtual network `closednet` is set to our default subnet of 172.31.128.x
+to connect DHCP and TFTP from RackHD to the PXE clients.
 
-  We expect the latest version of GIT, Vagrant, and Ansible installed onto the host system.
+## PRE-REQS / SCRIPT EXPECTATIONS
 
-  We expect static files to be located in the correct path from the parent directory of dev-tools:
+We expect the latest version of git, Vagrant, and Ansible installed onto your
+system in order to use this script.
 
-    i.e.
+We also rely on the this projects structure of submodules to link the source
+into the VM (through vagrant). The ansible roles are written to expect the
+source to be autoloaded on the virtual machine with directory mappings
+configured in Vagrantfile:
+
+for example:
+
         ~/<repos directory>/RackHD/on-http/static/http/common/
 
-  Our static files can be built locally using the tools found here:
-      https://github.com/RackHD/on-imagebuilder
+The static files that RackHD uses can be built locally using the tools found in
+the on-imagebuilder repository (https://github.com/RackHD/on-imagebuilder),
+and this script will download the latest built versions that are stored in
+bintray from that open source repository's outputs.
 
 ## SET UP INSTRUCTIONS
 
+Clone RackHD repo to your local git directory.
 
-  Clone RackHD repo to your local git directory.
-
-    i.e.
-        ~/<repos directory>/RackHD/
+    $ git clone https://github.com/RackHD/RackHD
+    $ cd RackHD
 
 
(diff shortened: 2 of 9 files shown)


example/roles/repos/tasks/main.yml
```diff
@@ -67,3 +67,14 @@
    - monorail-efi32-snponly.efi
    - monorail-efi64-snponly.efi
    - undionly.kpxe
+
+- name: Mark coreOS directory for hosting coreos demonstration booting
```
benbp: "CoreOS"

benbp:
:+1: with minor spelling nit.

DavidjohnBlodgett:
:+1:

heckj:
(yes, I'll squash after I get done screwing around with this)

heckj:
@DavidjohnBlodgett yeah, in general - when I was doing "stupid dev tricks", I wanted to add a remote for my own fork, and realized I was having to screw with github auth using HTTPS:// and my dual factor authentication thing, or I could forward the SSH credentials when I did a `vagrant ssh` to get into the VM. Not strictly needed for this effort, but nice to have.

DavidjohnBlodgett:
@heckj cool thanks!

DavidjohnBlodgett:
:+1:

heckj:
@DavidjohnBlodgett I ran through the script with the change, and it worked through brilliantly...

example/README.md
```diff
@@ -1,114 +1,176 @@
...
-# the monorail_rack setup script.
+    curl -H "Content-Type: application/json" \
+    -X PUT --data @samples/virtualbox_install_coreos.json \
+    http://localhost:9090/api/1.1/workflows
 
-# deployment variables
-pxe_count=1
-```
+We also include the sku definition. RackHD sku support includes a mechanism that
+will run a workflow when a SKU is identified. In this case, the example uses the
```
benbp: Inconsistent capitalization of 'SKU'.

Also, this language is a little misleading perhaps? "sku support includes a mechanism that will run a workflow when a SKU is identified." We don't run a workflow right when the SKU is identified, it has to be triggered by another workflow.

example/bin/monorail_rack
```diff
@@ -61,16 +54,20 @@ if [ $pxe_count ]
...
-        vboxmanage modifyvm $vmName --nic1 intnet --intnet1 closednet --nicpromisc1 allow-all;
-        vboxmanage modifyvm $vmName --nictype1 82540EM --macaddress1 auto;
-
+        if [[ ! -e $vmName.vdi ]]; then # check to see if PXE vm already exists
+            echo "deploying pxe: $i"
+            vboxmanage createvm --name $vmName --register;
+            vboxmanage createhd --filename $vmName --size 8192;
+            vboxmanage storagectl $vmName --name "SATA Controller" --add sata --controller IntelAHCI
+            vboxmanage storageattach $vmName --storagectl "SATA Controller" --port 0 --device 0 --type hdd --medium $vmName.vdi
+            vboxmanage modifyvm $vmName --ostype Ubuntu --boot1 net --memory 768;
```
benbp: Should we just bump memory to 1024?

example/roles/repos/tasks/main.yml
```diff
@@ -11,31 +11,29 @@
...
-- name: Checkout branch and set HEAD to most current
-  shell: git checkout {{ branch }}
+- name: Reset to the branch specified
+  shell: git reset --hard {{ branch }}
   when: branch is defined
   args:
     chdir: /home/vagrant/src/{{ item }}
   with_items:
     - on-core
+    - on-tasks
```
benbp: Ah

.gitignore
```diff
@@ -1,3 +1,3 @@
-dev-tools/.vagrant
-dev-tools/config/monorail_rack.cfg
-dev-tools/bin/pxe*
+example/.vagrant
+example/config/monorail_rack.cfg
+example/bin/pxe*
```
DavidjohnBlodgett: oh good catch!

example/bin/monorail_rack
heckj: I tried with both, and was able to get away with 768MB, so I thought I'd try and keep it as small as possible from a quick/dev setup perspective

benbp: k

benbp:
@DavidjohnBlodgett the casing issue of virtualboxmanage might vary depending on whether you have a case-sensitive filesystem or not.

benbp:
:+1: after minor nit fixes.

example/README.md
heckj: How's something like this:

This example includes a workflow that we'll use when we identify a "virtualbox"
SKU with RackHD. This workflow sets up no-op out of band management settings
for a demo and triggers an installation of CoreOS as a default flow to run
once the "virtualbox" SKU has been identified. We'll load it into our library
of workflows:

    cd ~/src/rackhd/example
    # make sure you're in the example directory to reference the sample JSON correctly

    curl -H "Content-Type: application/json" \
    -X PUT --data @samples/virtualbox_install_coreos.json \
    http://localhost:9090/api/1.1/workflows

To enable that workflow, we also need to include a SKU definition that includes
the option of another workflow to run once the SKU has been identified. This
takes advantage of the discoverySKU workflow, which will attempt to identify a
SKU and run another workflow if specified.

benbp: :+1:

benbp: "DiscoverySKU"?

DavidjohnBlodgett:
:+1:

rolandpoulter:
:+1:

PR #15, opened 2015-11-24
devscripts

heckj:
local dev scripts to make development (for me anyway) easier- created symbolic links for on-core and on-tasks and uses where-ever the git submodules are. Can be used to install locally on a mac or linux, and the ansible based Vagrant install uses it with this PR. Doesn't impact time to install materially, but does give you a "live edit" view.

`vagrant up` time:

    real 27m45.130s
    user  0m8.808s
    sys   0m3.552s

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 example/.vagrant
 example/config/monorail_rack.cfg
 example/bin/pxe*
+*.vdi
--- a/example/Vagrantfile
+++ b/example/Vagrantfile
@@ -18,7 +18,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             # If you wish to use a specific branch, enable the variable below
             # and the repos role will check out code on that branch across
             # the RackHD git repositories.
-            # ansible.extra_vars = { branch: "master" }
+            # ansible.extra_vars = { branch: "master", node_version: '0.10.40' }
 
             # if the playbook seems hung try uncommenting below to enable
             # debuging level output
--- a/example/roles/node/tasks/main.yml
+++ b/example/roles/node/tasks/main.yml
@@ -10,10 +10,6 @@
   when: download_n|changed
 
 - name: Install node
-  shell: /home/vagrant/n/bin/n {{ item }}
-  with_items:
-#    - 4.1.1
-#    - 0.12.7
-    - 0.10.40
+  shell: /home/vagrant/n/bin/n {{ node_version | default('0.10.40') }}
   sudo: yes
   when: download_n|changed
--- a/example/roles/repos/tasks/main.yml
+++ b/example/roles/repos/tasks/main.yml
@@ -26,15 +26,25 @@
     - on-syslog
   sudo: yes
 
-- name: Npm install Repos
-  npm: path=/home/vagrant/src/{{ item }}
-  with_items:
-    - on-syslog
-    - on-tftp
-    - on-dhcp-proxy
-    - on-taskgraph
-    - on-http
-  sudo: yes
+- name: clean any local installation
+  shell: './src/scripts/clean_all.bash'
+  args:
+    chdir: /home/vagrant
+
+- name: Link and install locally
+  shell: './src/scripts/link_install_locally.bash'
+  args:
+    chdir: /home/vagrant
+
+# - name: Npm install Repos
+#   npm: path=/home/vagrant/src/{{ item }}
+#   with_items:
+#     - on-syslog
+#     - on-tftp
+#     - on-dhcp-proxy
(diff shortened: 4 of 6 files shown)


DavidjohnBlodgett:
:+1: thanks! @heckj

heckj:
@benbp @jlongever @brianparry @yyscamper @anhou quick looksee when you have some time, please. 

The symbolic link mechanism didn't end up saving time because if you do it first, "npm install" just overruns it and replaces it, so you've got to do it second - meaning you've got all the overhead of the initial npm install anyway :-( Fortunately, `npm` caches aggressively, so that saves most of the heavy re-install lifting. It does, however, save you the fist banging on the keyboard you'd otherwise have to enable to try out live branches with an install.

jlongever:
@heckj @benbp Is this something that could satisfy both dev and deployment of rackhd with a shared link reference to core library?

jlongever:
+1

heckj:
@jlongever honestly probably not how you'd want to install on a production system, but will definitely suffice for where you want to install and hack on source (dev setups, test environment, etc)

benbp:
Do the symbolic links persist in the host file system since these are virtual box shared directories?

benbp:
Also, curious why `npm link` didn't work for you?

jlongever:
@heckj Perhaps not using symlinks,  I ask because this was a discussion out of the CC summit where we wanted to remove duplicated modules and point to a shared location (decreasing the deployed footprint).

heckj:
Re: symbolic links, I'll check my laptop when I get back to it. I believe so, but I didn't explicitly check after I ran the installs through, only looked from the OS that it was run on (either Mac or within Vagrant). I expect to see them.

I avoided `npm link` because in the past I've had issues with it depending on where and how NPM was installed, where `npm link` wanted to set the files into a possibly root-only owned area, resulting in needing to use sudo in some cases, and then getting wacky permissions all over the place.

heckj:
@jlongever sorry, i didn't understand your last question. The script I wrote uses symbolic links - are you asking why, or suggesting using something else?

jlongever:
@heckj I'm asking how we should handle production style deployments.  At the CC summit we discussed using npm link or symlinks (in production deployments) to avoid installing duplicated libraries shared across services (i.e. on-core).

jlongever:
maybe this decision is up to the consumer and how they wish to deploy.

heckj:
@jlongever Ah - production employment, yeah - I wouldn't expect this to be a means to do so. When we've made debian (or RPM) packages in the past from this code, we used "npm install && npm prune --production", which removes the "devDependencies" set from package.json, but includes all the relevant code for the platform on which it was built. In fact, we could fairly directly extend the travis stuff to make these packages and stash those in bintray if we wanted - a point for another time I guess.

Along with how the code is installed, I'd expect someone running in production to use the moral equiv of an upstart or SYSV daemon script - be that systemd, upstart, or an init script - to invoke the code, capture any local logs, etc. As you say though, exactly how someone packages this into their solution (for example, a docker-based solution would be different that what I'm suggesting) is really up to the packager

heckj:
@benbp I checked the symbolic links, and they reference the host upon which they're run - so if you run within vagrant and then look in your local directory on a mac, you'll see a symbolic link to `/home/vagrant/...`. That said, I think that might be for the best - as the NPM install will have made any C library extensions under node_modules for the relevant OS, so what's there is in no way guaranteed to work with another OS.

I could change these to use local references (aka `ln -s ../something node_modules/`) so they'll "work" to point to the right location regardless of OS due to the structure of the submodules, but if it fails due to C library extensions made for another OS, that will be far less clear.

anhou:
Using below steps to avoid root or sudo operation.
1. touch .npmrc in home directory, and adding 
export PATH=~/npm/bin:$PATH

then global installed modules will be installed to ~/npm/bin/ directory
2. create 'npm' directory in home directory
$ mkdir ~/npm

then we can run  npm link without root access or sudo
3. add 'export PATH=~/npm/bin:$PATH' to .bash_profile

heckj:
@anhou and @benbp - would you two prefer me to redo these scripts using NPM link and set up .npmrc extensions with the vagrant setup so that the modules are all linked locally? I'm good either way, wasn't sure what you intended with the feedback - offering an alernative or asking that we use that alternative.

anhou:
@heckj  just a offer an alternative way :), yes, that link is the same way. I re-think about this way, it will add a 'npm' directory and also modify the .bash_profile,  it will help a lot when manually creating link by 'npm link' instead of 'ln -s' , but it's not so clean that create a directory and change .bash_profile comparing this PR. so I think the way of this PR is good.
:+1:

PR #17, opened 2015-12-03
Fix coreos login IP and instructions in README

benbp:


benbp:
@heckj

benbp:
@DavidjohnBlodgett

example/README.md
```diff
@@ -140,11 +140,9 @@ installed. The installation workflow included in the RackHD system installs
 with a default SSH key that's included in our repositories. From the `example`
 directory, you should be able to log in using:
 
-`vagrant ssh`
+`vagrant ssh`:
 
-    cp ~/src/on-http/data/rackhd_rsa ~/.ssh/id_rsa
-    chmod 400 ~/.ssh/id_rsa
-    ssh -i ~/src/on-http/data/rackhd_rsa core@172.32.128.2
+    ssh -i ~/src/on-http/data/rackhd_rsa core@172.31.128.2
```
heckj: Unfortunately, I don't think this'll work - SSH won't let you use the key without more constrained permissions, and git checks it out as "644" default. So at the minimu, you need to `chmod 400 ~/src/on-http/data/rackhd_rsa`

tldavies:
+1

example/README.md
benbp: So, should we just do the first two steps and then `ssh core@...` instead of `ssh -i` ?

heckj:
Moving into a local .ssh/ may make other demo steps easier later. Could even do as part of an ansible vagrant setup role too

benbp:
Sounds good, will make the change.

heckj:
+1 (thanks @benbp )

PR #18, opened 2015-12-04
removed all but one apt-get update from roles

DavidjohnBlodgett:
It was suggested to only run apt-get update on the first role we run to elevate repetitive actions. I have made the fix. @heckj @benbp

### DIFF
--- a/example/dev.yml
+++ b/example/dev.yml
@@ -3,6 +3,7 @@
   hosts: all
   sudo: no
   roles:
+    - apt
     - build
     - git
     - rabbitmq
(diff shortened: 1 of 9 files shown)


heckj:
+1

benbp:
I think it makes more sense to have the update be its own role, rather than just keeping it in the build-essential role.

benbp:
But yeah, this should shave off several minutes.

tldavies:
+1

benbp:
:+1:

PR #20, opened 2015-12-04
initial work in progress port of internal packer configs

heckj:
# WIP pull request - please do not merge

    export ATLAS_TOKEN=...
    export ATLAS_USERNAME=rackhd
    export ATLAS_NAME=rackhd
    packer build --only=virtualbox-iso template.json


example/roles/overlays/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
+---
```
benbp: Probably makes more sense to name the role 'images' ? Since you're getting all the pxe/coreOS stuff too.

heckj: updated in next commit

PR #30, opened 2015-12-10
Filter logs

heckj:
adding default filtering to "notice" level to calm down the demonstration log output


heckj:
@benbp @dalebremner @brianparry adding filtering to the demo setup - will make a corresponding PR to take the low-level notices about blocking the event loop from notice to debug in on-core

benbp:
info level perhaps better in light of https://github.com/RackHD/RackHD/issues/34 ?

heckj:
yeah, thinking so - when we merge in any log level changes, I'll swing back by and update this

benbp:
:+1:

PR #58, opened 2015-12-23
Update client library

jlongever:
- Update client library to python-on-http
- Add nose package to requirement

@RackHD/corecommitters @RackHD/rackhd_dev 

https://github.com/RackHD/on-http/pull/77 should be merged first.


heckj:
:+1:

test/README.md
```diff
@@ -5,6 +5,10 @@
 
     virtualenv .venv
     sudo pip install -r requirements.txt
+    
+    echo "deb https://dl.bintray.com/rackhd/debian trusty main" | sudo tee -a /etc/apt/sources.list.d/rackhd-trusty.list
```
brianparry: It didn't occur to me when I reviewed https://github.com/RackHD/on-http/pull/76, but I think we should be packaging this as a tarball or wheel that can be installed using pip so that the package can be installed in the virtualenv rather than in dist-packages.

brianparry:
:+1:  Changes look good, but I think we should move away from debian packaging for the generated library.

jlongever:
@brianparry We already had debian deployment in place but we can setup pypi deployment in another PR.

PR #61, opened 2015-12-30
Validate nodes APIs

uppalk1:


uppalk1:
@keedya @jlongever @tannoa2 @zyoung51 @benbp

keedya:
@RackHD/corecommitters

test/tests/nodes_tests.py
```diff
@@ -57,20 +57,40 @@ def __init__(self):
...
+        for n in nodes:
+            if n.get('type') == 'compute':
+                uuid = n.get('id')
+                Nodes().api1_1_nodes_identifier_obm_identify_post(uuid, data)
+                rsp = self.__client.last_response
+                codes.append(rsp)
+        assert_not_equal(0, len(codes), message='Failed to find compute node Ids')
+        for c in codes:
+            assert_equal(200, c.status, message=c.reason)
+        assert_raises(rest.ApiException, Nodes().api1_1_nodes_identifier_get, 'fooey')
```
heckj: Not critical, but I might suggest this last test as a separate test case because it's testing a different scenario than the previous bits of code creating nodes. If this broke, but not the rest of this method - it would be a tad more difficult to dig out, because we're testing the corner case.

(This is totally my opinion on structuring tests - take it as a suggestion only, not a requirement to change for merging)

heckj:
I had some suggestions inline for breaking down the tests based on testing fwd/happy path vs. corner cases, but as a suggestion only. +1 from me for getting some testing there and operational!

jlongever:
Agree with @heckj.  Create a post test and a get test, then you can make the post test a dependent test.  +0

keedya:
There is some cases where post tests depend on other post tests(create nodes->post I made settings). Also in workflows the sequence of the test is important ex: post workflow -> get workflow active -> delete workflow active.

uppalk1:
After looking at this further we did find a way to do the segregation without causing any dependency issues within the tests. However looks like it needs some more scrubbing and testing before we can make the changes and be confident. For now we'll work on further improving as we work on more tests and then come back to enhance these files. 
I fixed the other issue Joe Heck highlighted. 
Can we continue to merge this file and work on further enhancements later based on what we learn on other test cases ?

keedya:
:+1:

heckj:
:+1:

PR #62, opened 2015-12-30
Validate Profile APIs

tannoa2:


keedya:
@jlongever @benbp @uppalk1

tannoa2:
@keedya

keedya:
@RackHD/corecommitters

heckj:
+1 - love the use of the swagger generated client in the tests.

jlongever:
Cool.   We should make sure we run a valid profile end-to-end and don't limit the test to just checking the HTTP response.  +1

jlongever:
I also realize some tests can be more simple, but tests like workflows and profiles become slightly more complex.

keedya:
:+1:

PR #64, opened 2016-01-06
Fix OBM tests

jlongever:
- Delete redundant test_node_id_obm_create()
- Move OBM identify test into OBMTests() and set check-obm dependency

@RackHD/corecommitters @zyoung51 @tannoa2 @keedya @uppalk1


heckj:
:+1:

zyoung51:
+1

PR #70, opened 2016-01-14
Fix ESXi installation chainloader from ipxe to syslinux

anhou:
undionly.kpxe from ipxe is not the original gPXE's undionly.kkpxe from syslinux  which has been validated for installing ESXi on multiple platforms, correct it here. This will fix the chainload failure issue when installing ESXi OS
related PR: 
https://github.com/RackHD/on-imagebuilder/pull/17
https://github.com/RackHD/on-http/pull/91


heckj:
:+1:

benbp:
:+1:

PR #74, opened 2016-01-15
Added AMQP connection error handler.

jlongever:


jlongever:
@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2

test/modules/amqp.py
```diff
@@ -40,6 +44,14 @@ def on_message(self, body, message):
     def on_conn_retry(self):
         LOG.error('Retrying connection for {0}'.format(self.__amqp_url))
 
+    def on_connection_error(self, exc, interval):
```
zyoung51: If this is called to re-establish a connection, should the __max_error be reset once the connection is established and good again?

jlongever: Good point.  Maybe I can call ensure_connection() and reset __max_error if the connection good?

jlongever: nvm, ensure_connection takes the on_error callback.

jlongever: @zyoung51 The connection retry should be reset by ensure_connection when the connection is reestablished (ensure_connection maintains it's own retry count and interval.)   The __max_error is a global threshold for the number of ensure_connection attempts.

benbp:
:+1:

keedya:
:+1:

Issue #96, opened 2016-02-17
packer build failing...


heckj:
I was running a new packer build process from master this morning, and found that packer build wasn't completing:

https://atlas.hashicorp.com/heckj/build-configurations/rackhd/builds/5?expanded=virtualbox-iso

The error was placing `monorail.json` from updated configuration names

Issue #94, opened 2016-02-16
http proxy doesn't work with master branch and existing configuration in vagrant setup


heckj:
While debugging https://github.com/RackHD/RackHD/issues/90, I found that the latest code from master on all branches as of today (2/16/2016) fails with the existing proxy configuration as it's created by Packer/Ansible (config file at https://github.com/RackHD/RackHD/blob/master/packer/ansible/roles/monorail/files/config.json). 

To reproduce this I spun up the Vagrant demonstration and then updated the code to the latest master branches with:

```cd ~/src
./scripts/clean_all.bash && ./scripts/reset_submodules.bash && ./scripts/link_install_locally.bash
cd ~
sudo nf start```

And thereafter the tests I was doing to debug the proxy issue in https://github.com/RackHD/RackHD/issues/90 failed consistently with the response `Cannot GET /test/test1` to the requests for any files defined in the proxy configuration.

**Comment by heckj** [comment]
@anhou @pengz1 @yyscamper - wanted to let you know about this - I suspect (but haven't verified) that the API auth and split north/south interfaces means that we need a different configuration for the Vagrant instance in order to allow this to work.

**Comment by anhou** [comment]
@heckj You're right! and I also confirmed with @WangWinson, For http proxy, the latest 'on-http' should work with the latest 'config.json' which is updated in config.json change PR https://github.com/RackHD/RackHD/pull/84/files and endpoint splitting PR https://github.com/RackHD/RackHD/pull/83. so http proxy could work in vagrant@rackhd-demo with following commands before 'sudo nf start'
```
cd ~/src
git fetch
git rebase origin/master
sudo mkdir /opt/monorail
sudo cp ./packer/ansible/roles/monorail/files/config.json /opt/monorail/
cd ~
sudo nf start
```
And could you help update a new virtualbox.box so that new vagrant virtualbox include the new config.json? 

below is my steps to verify http proxy function in vagrant@rackhd-demo 
$ sudo vi /opt/monorail/config.json

    "httpProxies": [{
        "localPath": "/coreos",
        "server": "http://172.31.128.1:9080",
        "remotePath": "/common/"
    }],
$ sudo nf start
$ curl 172.31.128.1:9080/coreos/base.trusty.3.13.0-32-generic.squashfs.img -o test
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #97, opened 2016-02-17
Step forward submodules, packer rebuild

heckj:
Stepping forward the modules to rebuild a latest vagrant instance that leverages the most recent code.

resolves https://github.com/RackHD/RackHD/issues/96
believed will also resolve https://github.com/RackHD/RackHD/issues/94

### DIFF
--- a/packer/ansible/roles/monorail/tasks/main.yml
+++ b/packer/ansible/roles/monorail/tasks/main.yml
@@ -4,5 +4,5 @@
   sudo: yes
 
 - name: Load monorail conf onto guest
-  copy: src=monorail.json dest=/opt/monorail/config.json
+  copy: src=config.json dest=/opt/monorail/config.json
   sudo: yes


benbp:
Ah! :+1: 

FYI @DavidjohnBlodgett looks like I gave you an incomplete list of changes that needed to be made to support the config path update it seems?

DavidjohnBlodgett:
oh dear, I thought I double checked what packer was using for the config name~ Sorry about that. :+1:

DavidjohnBlodgett:
does this mean our on-core is broken? https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L46

DavidjohnBlodgett:
@benbp ^^

anhou:
:+1:

anhou:
@DavidjohnBlodgett let me answer instead of @benbp  :-)  I think on-core is not broken, in my test yesterday, the old monorail.json will take effect, if no config.json, and if config.json exist, it will be used instead of monorail.json. it's just a problem of config.json for packer here.

PR #106, opened 2016-02-24
Fix workflow related API tests

benbp:
See https://github.com/RackHD/on-http/pull/145 for more information. Requires https://github.com/RackHD/on-http/pull/145/files as well.

A lot of these tests have been returning false positives because the on-http server was incorrectly returning 200 status codes when it should have been returning 400 and 404 status codes. Recent updates to on-http now return proper error status codes, which is now breaking these tests.

@RackHD/corecommitters @VulpesArtificem @heckj


jlongever:
+1

anhou:
:+1:

yyscamper:
+1

heckj:
:+1:

jlongever:
Functional scripts pass with all dependencies locally.
+1

PR #111, opened 2016-02-25
Add initial test for auth enabled

WangWinson:
- Add config.json for smoke test 
- Add basic smoke test when auth enabled

@RackHD/corecommitters @cgx027

Dependency PR https://github.com/RackHD/on-http/pull/143 for API python client update


WangWinson:
Note: Jenkins will not pass until the PR to update Vagrantfile in on-build-config repo merged

jlongever:
@WangWinson Do you plan on submitting the PR to on-build-config?  I don't see one.
https://github.com/jlongever/on-build-config

jlongever:
@WangWinson nevermind,  I see the PR.  Could you submit it here: https://github.com/jlongever/on-build-config.   I moved on-build-config to public repo.

jlongever:
@WangWinson better yet, just forked the build tools into RackHD..   submit the PR here: https://github.com/RackHD/on-build-config.

test/tests/api/v1_1/__init__.py
```diff
@@ -16,5 +17,6 @@
     'profiles.tests'
     'config.tests',
     'workflowTasks.tests',
-    'workflows.tests'
+    'workflows.tests',
+    'auth.tests'
```
jlongever: this will always run auth.tests as part of the v1.1 group.   Does this still work when config.json disables authentication?

WangWinson: It won't work. It should work with the specific config.json for test

jlongever: So we should define a new group for testing with auth enabled.

WangWinson: You mean add something like below in https://github.com/RackHD/RackHD/blob/master/test/run.py
```python
register(groups=['api-v1.1-withauth'], depends_on_groups=api_1_1.tests_withauth)
```
But the problem is whether this was enabled by default. If the auth disabled in config file. This tests withauth will still fail.

WangWinson:
Note: Since we are now switch to `monorail.yml` to generate python client. So Jenkins will also require Joe's update https://github.com/jlongever/RackHD/commit/8ed616cfba8c46c6f1099ef588cbcee800ff76bc to pass all previous tests

packer/ansible/roles/monorail/files/auth_config/config.json
```diff
@@ -0,0 +1,65 @@
+{
+    "CIDRNet": "172.31.128.0/22",
```
jlongever: @WangWinson  Since this configuration is backwards compatible I would just overwrite the files/config.json with the new config.

@heckj  fyi once this goes in we'll need to generate a new vagrant box and update Atlas.

example/config/monorail_rack.cfg.example
```diff
@@ -16,3 +16,5 @@ PXE_COUNT=1
 # Select single repo to be synchronized with the vagrant instance (optional, requires WORKSPACE to be set).
 # REPO=on-http
 
+# Select directory for customized config file (optional, requires WORKSPACE to be set).
+# CONFIG_DIR=packer\ansible\roles\monorail\files\auth_config
```
jlongever: Since we can overwrite the config.json,  just set the example path to ```packer\ansible\roles\monorail\files```.  The option to override the config is still good :+1:

jlongever:
@WangWinson verified the changes locally :+1: after Jenkins check passes.   I  went ahead committed the new config.json to on-build-config so we can go ahead and test this please :)

test/tests/api/v1_1/auth_tests.py
```diff
@@ -0,0 +1,34 @@
...
+        """setup test environment"""
+        Auth.enable()
+
+    @after_class(always_run=True)
+    def teardown(self):
+        """ restore test environment """
+        Auth.disable()
+
+    @test(groups=['test-nodes-withauth'])
+    def test_nodes(self):
```
jlongever: I would really prefer we set this up so we can reuse the other test methods instead of rewriting the same code for authentication testing.  The only difference should be the api_client.

jlongever:
Dependent PR: https://github.com/RackHD/RackHD/pull/112

jlongever:
test this please

test/tests/api/v1_1/auth_tests.py
WangWinson: Agree. The existing tests should be reused. Maybe run twice of those tests. 1st time no auth. 2nd time with auth. The `modules/auth.py` is simply handle this with `Auth.enable()` and `Auth.disable()`. I'll figure out how to apply this to existing tests later. 
For this PR, just try to run a simple case to make sure auth works.

WangWinson:
resolve conflict and re-submit commits

WangWinson:
test this please

WangWinson:
Ooops. Not sure what happen. The error did not happen verify on my local PC. Trying to workaround the error 
`Reason: SSLError
[Errno 8] _ssl.c:510: EOF occurred in violation of protocol`
seem like some urllib3 issue. See
http://stackoverflow.com/questions/14102416/python-requests-requests-exceptions-sslerror-errno-8-ssl-c504-eof-occurred

jlongever:
test this please

anhou:
test this please

jlongever:
test this please

jlongever:
test this please

jlongever:
test this please

jlongever:
+1

jlongever:
@WangWinson up to you,  but you probably can remove those SSL/TLS workarounds.

WangWinson:
@jlongever  Yes, I'm doing it now.

example/Vagrantfile
```diff
@@ -30,6 +30,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
           else
             target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/#{ENV['REPO']}"
           end
+          if ENV['CONFIG_DIR']
+            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['CONFIG_DIR']}", "/home/vagrant/opt/monorail/"
```
jlongever: We should include the legacy /opt/onrack/etc location here as well.

WangWinson:
test this please

WangWinson:
test this please

WangWinson:
Jenkins finally pass. :+1:

anhou:
:+1: it's great that jenkins tests pass finally :-)

PR #119, opened 2016-03-04
Footprint benchmark initial checkin

iceiilin:
This PR implements the process of running benchmark test.
https://github.com/RackHD/RackHD/wiki/proposal-footprint-benchmarks

Users issue the test by
```
~/src/RackHD/test$ python benchmark.py
```
To run the test in API v2, please use "--api_version=2". (cases in api v2 has not been implemented yet)
API v1.1 is used by default.
 
The log data is in ~/benchmark/<timestamp>
Related info will be added to README later.

The first time users run this script, he will be asked about user/pwd in the test machine (local) and RackHD. This is used in Ansible to setup test environment both in local and RackHD.

It will include several typical footprint scenarios (like running poller, discovery and bootstrap), and aims to get footprint data from several tools (atop and mongo shell) and display the result.

This PR as an initial checkin, focuses on the scenario of **running poller** using api v1.1,  fetch data and parse it to js file for display purpse.

The working stream is as follows:

Step 1 and 4 in the above graph use ansible Python API to run playbooks and communicate with RackHD. You can find playbooks in tests/benchmark/utils/ansible/. Data collection interval is 1s by default

@RackHD/corecommitters @cgx027 @heckj @wuy1 @chenge3

### DIFF
--- a/test/config/benchmark_config.py
+++ b/test/config/benchmark_config.py
@@ -1,0 +1,41 @@
+from settings import *
+
+ANSIBLE_AUTH_FILE = '.ansible_auth'
+
+# Obfuscate credentials
+def set_ansible_auth(local_user, local_pwd, host_ssh_port, host_ssh_user, host_ssh_pwd):
+    lu = b64encode(local_user)
+    lp = b64encode(local_pwd)
+    rport = b64encode(host_ssh_port)
+    ru = b64encode(host_ssh_user)
+    rp = b64encode(host_ssh_pwd)
+
+    with open(ANSIBLE_AUTH_FILE,'w+') as file:
+        out = \
+        'local_user="{0}"\n'.format(lu) + \
+        'local_pwd="{0}"\n'.format(lp) + \
+        'host_ssh_port="{0}"\n'.format(rport) + \
+        'host_ssh_user="{0}"\n'.format(ru) + \
+        'host_ssh_pwd="{0}"\n'.format(rp)
+        file.write(out)
+
+# Unobfuscate credentials
+def get_ansible_auth():
+    ansible_auth = load_source('creds',ANSIBLE_AUTH_FILE)
+    return b64decode(ansible_auth.local_user), b64decode(ansible_auth.local_pwd), \
+        b64decode(ansible_auth.host_ssh_port), b64decode(ansible_auth.host_ssh_user), \
+        b64decode(ansible_auth.host_ssh_pwd)
+
+# Initial ansible auth file if it doesn't exist
+if os.path.isfile(ANSIBLE_AUTH_FILE) is False:
+    local_user = raw_input('localhost username: ')
+    local_pwd = getpass('localhost sudo password: ')
+    host_ssh_port = raw_input('RackHD ssh port (22): ')
+    host_ssh_user = raw_input('RackHD ssh username (onrack): ')
+    host_ssh_pwd = getpass('RackHD ssh password (onrack): ')
+
+    host_ssh_port = "22" if host_ssh_port == "" else host_ssh_port
+    host_ssh_user = "onrack" if host_ssh_user == "" else host_ssh_user
+    host_ssh_pwd = "onrack" if host_ssh_pwd == "" else host_ssh_pwd
+
+    set_ansible_auth(local_user, local_pwd, host_ssh_port, host_ssh_user, host_ssh_pwd)
--- a/test/config/benchmark_config.py
+++ b/test/config/benchmark_config.py
@@ -1,33 +1,17 @@
 from settings import *
 
-ANSIBLE_AUTH_FILE = '.ansible_auth'
-
-# Obfuscate credentials
 def set_ansible_auth(local_user, local_pwd, host_ssh_port, host_ssh_user, host_ssh_pwd):
-    lu = b64encode(local_user)
-    lp = b64encode(local_pwd)
-    rport = b64encode(host_ssh_port)
-    ru = b64encode(host_ssh_user)
-    rp = b64encode(host_ssh_pwd)
+    set_b64_cred({"LOCAL_USER":local_user, "LOCAL_PWD":local_pwd, "HOST_SSH_PORT":host_ssh_port, \
+                  "HOST_SSH_USER": host_ssh_user, "HOST_SSH_PWD": host_ssh_pwd})
 
-    with open(ANSIBLE_AUTH_FILE,'w+') as file:
-        out = \
-        'local_user="{0}"\n'.format(lu) + \
-        'local_pwd="{0}"\n'.format(lp) + \
-        'host_ssh_port="{0}"\n'.format(rport) + \
-        'host_ssh_user="{0}"\n'.format(ru) + \
-        'host_ssh_pwd="{0}"\n'.format(rp)
-        file.write(out)
-
-# Unobfuscate credentials
 def get_ansible_auth():
-    ansible_auth = load_source('creds',ANSIBLE_AUTH_FILE)
-    return b64decode(ansible_auth.local_user), b64decode(ansible_auth.local_pwd), \
-        b64decode(ansible_auth.host_ssh_port), b64decode(ansible_auth.host_ssh_user), \
-        b64decode(ansible_auth.host_ssh_pwd)
+    return get_b64_cred(['LOCAL_USER', 'LOCAL_PWD', 'HOST_SSH_PORT', \
+                        'HOST_SSH_USER', 'HOST_SSH_PWD'])
 
-# Initial ansible auth file if it doesn't exist
-if os.path.isfile(ANSIBLE_AUTH_FILE) is False:
+# Initial ansible auth information if it doesn't exist
+try:
+    get_ansible_auth()
+except:
     local_user = raw_input('localhost username: ')
     local_pwd = getpass('localhost sudo password: ')
     host_ssh_port = raw_input('RackHD ssh port (22): ')
--- a/test/tests/api/v1_1/poller_tests.py
+++ b/test/tests/api/v1_1/poller_tests.py
@@ -1,0 +1,60 @@
+from config.api1_1_config import *
+from config.amqp import *
+from modules.logger import Log
+from on_http_api1_1 import NodesApi as Nodes
+from on_http_api1_1 import PollersApi as Pollers
+from on_http_api1_1 import TemplatesApi as Templates
+# from on_http_api1_1 import rest
+# from datetime import datetime
+from proboscis.asserts import assert_equal
+# from proboscis.asserts import assert_false
+# from proboscis.asserts import assert_raises
+from proboscis.asserts import assert_not_equal
+# from proboscis.asserts import assert_is_not_none
+# from proboscis.asserts import assert_true
+from proboscis import SkipTest
+from proboscis import test
+from json import loads
+import time
+
+LOG = Log(__name__)
+
+@test(groups=['poller.tests'])
+class PollerTests(object):
+
+    def __init__(self):
+        self.__client = config.api_client
+
+    @test(groups=['test-node-poller'], depends_on_groups=['test-nodes'])
+    def test_node_pollers(self):
+        """ Test /nodes/:id/pollers are running """
+        Nodes().nodes_get()
+        nodes = loads(self.__client.last_response.data)
+        LOG.debug(nodes, json=True)
+        samples = []
+        valid = False
+
+        for n in nodes:
+            LOG.info(n)
+            if n.get('type') == 'compute':
+                uuid = n.get('id')
+                Nodes().nodes_identifier_pollers_get(uuid)
+                rsp = self.__client.last_response
+                data = loads(rsp.data)
+                assert_equal(200, rsp.status, message=rsp.reason)
+                assert_not_equal(0, len(data), \
+                        message='Failed to find poller for nodes {0}'.format(n.get('id')))
+                samples.append(data[0])
+
+        for sample in samples:
+            count = 18 # Wait for 2 mins (poller interval is 1 min)
+            while valid == False:
+                try:
+                    Templates().pollers_identifier_data_get(sample.get('id'))
+                    valid = True
+                except Exception, e:
+                    LOG.warning('Poller {0} doesn\'t work normally'.format(sample.get('id')))
+                    time.sleep(10)
+                    count -= 1
+                    assert_not_equal(0, count, \
+                            message='Poller {0} failed to get data'.format(sample.get('id')))
--- a/test/tests/api/v1_1/poller_tests.py
+++ b/test/tests/api/v1_1/poller_tests.py
@@ -4,14 +4,8 @@ from modules.logger import Log
 from on_http_api1_1 import NodesApi as Nodes
 from on_http_api1_1 import PollersApi as Pollers
 from on_http_api1_1 import TemplatesApi as Templates
-# from on_http_api1_1 import rest
-# from datetime import datetime
 from proboscis.asserts import assert_equal
-# from proboscis.asserts import assert_false
-# from proboscis.asserts import assert_raises
 from proboscis.asserts import assert_not_equal
-# from proboscis.asserts import assert_is_not_none
-# from proboscis.asserts import assert_true
 from proboscis import SkipTest
 from proboscis import test
 from json import loads
@@ -47,7 +41,7 @@ class PollerTests(object):
                 samples.append(data[0])
 
         for sample in samples:
-            count = 18 # Wait for 2 mins (poller interval is 1 min)
+            count = 18 # Wait for 3 mins (poller interval is 1 min)
             while valid == False:
                 try:
                     Templates().pollers_identifier_data_get(sample.get('id'))
--- a/test/benchmark.py
+++ b/test/benchmark.py
@@ -1,0 +1,44 @@
+import sys
+import getopt
+from proboscis import register
+from proboscis import TestProgram
+
+def run_tests(api_ver, selected):
+
+    if api_ver == '2':
+        import benchmark.api_v2_0 as benchmark
+    else:
+        import benchmark.api_v1_1 as benchmark
+
+    register(groups=['poller'], depends_on_groups=benchmark.benchmark_poller_tests)
+    register(groups=['discovery'], depends_on_groups=benchmark.benchmark_discovery_tests)
+    register(groups=['bootstrap'], depends_on_groups=benchmark.benchmark_bootstrap_tests)
+
+    if selected == False:
+        TestProgram(groups=['poller','discovery','bootstrap']).run_and_exit()
+    else:
+        TestProgram().run_and_exit()
+
+
+
+if __name__ == '__main__':
+
+    api_version = "1"
+    group_selected = False
+
+    try:
+        opts, args = getopt.getopt(sys.argv[1:], "h", ["api_version=", "group="])
+        for op, value in opts:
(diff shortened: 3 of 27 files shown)


yyscamper:
I see the benchmark will run those special designed test cases, this is OK for init design. But as a follow up I think you should think about how to leverage the existing test cases? So we can avoid to write duplicated test cases.

test/tests/benchmark/ansible/group_vars/benchmark
```diff
@@ -0,0 +1,3 @@
+test_machine_log_path: "{{ lookup('env', 'HOME') }}/benchmark/20160304-045310"
```
heckj: I suspect this hardcoded datestring after `/benchmark/` isn't quite intended...

test/tests/benchmark/ansible/start_daemon.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Restart rsyncd server
+      command: /etc/init.d/rsync restart
+      sudo: yes
+
+    - name: Install collectd
+      apt: name=collectd state=present
+      sudo: yes
+
+    - name: Copy config file to collectd
+      copy: src=files/test.cfg dest=/etc/collectd/
```
heckj: the file `files/test.cfg` looks blank in this PR, is that intended?

heckj:
Couple of questions in the PR specifically, looks good as an initial cut.

In the diagram above, it looks like you're intending to capture the footprints around each individual test case, which I didn't think you'd actually want to do  - but instead do a "set up" at the beginning, run the relevant benchmarks, and then down a completion/teardown where you captured the data. What's the level of granularity you're intending to capture data around?

Since I couldn't see a collectd configuration file, I wasn't sure what this was capturing right now, but it seems like a good start.

Do you have any sense yet on what aggregate measures might be the most valuable, or are you leading into that and intend to run a few rounds and then pick what's most valuable?

test/tests/benchmark/ansible/group_vars/benchmark
iceiilin: This timestamp is written to this file at runtime when starting to run the test group. Test cases within this group will use this variable afterwards and write logs to this specific path. So this variable changes every time the group runs. I can find a way to restore this variable after test group ends, but I don't think it is necessary.

test/tests/benchmark/ansible/start_daemon.yml
iceiilin: Right, it is a placeholder supposed to be modified

iceiilin:
@yyscamper Great suggestion. I will consider to make the footprint benchmark as a decorator thus every cases can leverage it if they want.

iceiilin:
@heckj You are right: the environment setup would be executed only once at the begining of this test group; daemon and data capture will be run per test case (test scenario). I will move the playbook execution. 
With @cgx027 's great investigation, we will use **atop and db.stats** instead of collectd for data capture. The granularity will be **5s** (atop default) or **10s** (statsd default), since they are commonly used (default value in popular tools), and also it can change if having trouble during our test.
For every test scenario, we will run this scenario once, and log data when runtime, as well as a summary of every entry in footprint benchmark during the whole test case. The final result for every footprint benchmark will include:
* runtime graph
* min, max and average value of this graph
* Summary of the resource consumption in this period

test/config/benchmark_config.py
```diff
@@ -0,0 +1,11 @@
+from settings import * 
+from on_http import Configuration, ApiClient
```
WangWinson: if you are targeting using 1.1 API python client. The latest lib name  changed from `on_http` to `on_http_api1_1`.
Also, any special reason why not leverage existing `api1_1_config.py`

test/tests/benchmark/runtime_tests.py
```diff
@@ -0,0 +1,118 @@
...
+from on_http import NodesApi as Nodes
+from on_http import WorkflowsApi as Workflows
+from on_http import rest
+from datetime import datetime
+from proboscis.asserts import assert_equal
+from proboscis.asserts import assert_false
+from proboscis.asserts import assert_raises
+from proboscis.asserts import assert_not_equal
+from proboscis.asserts import assert_is_not_none
+from proboscis.asserts import assert_true
```
WangWinson: Remove unused `import`

test/tests/benchmark/runtime_tests.py
```diff
@@ -0,0 +1,118 @@
+from config.benchmark_config import *
+from config.amqp import *
+from modules.logger import Log
+from modules.amqp import AMQPWorker
+from on_http import NodesApi as Nodes
```
WangWinson: lib name `on_http` updated. You need to generate the python client code with the latest on-http code

test/tests/benchmark/ansible_control.py
```diff
@@ -0,0 +1,111 @@
...
+from modules.logger import Log
+
+from config.benchmark_config import *
+from ansible.playbook import PlayBook
+from ansible.inventory import Inventory
+from ansible import callbacks
+from ansible import utils
+from ansible import constants
+
+import jinja2
```
WangWinson: Does the `jinja2` need a pre installation? If yes, you may add to test/requirements.txt

test/tests/benchmark/runtime_tests.py
```diff
@@ -0,0 +1,118 @@
...
+        nodes = loads(self.__client.last_response.data)
+        count = 0
+        for n in nodes:
+            type = n.get('type')
+            if type == 'compute':
+                count += 1
+        return count
+
+    @test(groups=['runtime.tests', 'check_precondition'])
+    def test_check_precondition(self):
```
WangWinson: Could this leverage existing `nodes.tests`?

test/tests/benchmark/data_display.py
```diff
@@ -0,0 +1,43 @@
...
+        rrdcmd_template = jinja2.Template(cmd)
+        rrdcmd = rrdcmd_template.render({
+            'log_path': log_path,
+            'name': self.__testname
+        })
+
+        os.system(rrdcmd)
+
+    def generate_comparative_graph(self):
+        return
```
WangWinson: Looks like just a place holder function here. You may use `pass` instead of `return`

test/run.py
```diff
@@ -3,13 +3,15 @@
 
 def run_tests():
 
-    import tests.api.v1_1 as api_1_1 
+    import tests.api.v1_1 as api_1_1
     import tests.api.v2_0 as api_2_0
     import tests.api.redfish_1_0 as api_redfish_1_0
+    import tests.benchmark as benchmark
```
WangWinson: This means benchmark test will always run at the end of Jenkins test. Does it by design ?

test/config/benchmark_config.py
jlongever: `pip install requirements.txt` will install the necessary libraries.

test/tests/benchmark/__init__.py
```diff
@@ -0,0 +1,31 @@
+# import tests
+import time
+import os
+
+from runtime_tests import RuntimeTests
+tests = [
+    'runtime.tests'
+]
+
+def create_log_path():
```
jlongever: Is there a specific reason for not reusing the logger module?

test/tests/benchmark/runtime_tests.py
jlongever: Run `pip install -r requirements.txt` and `from on_http_api1_1 import NodesApi as Nodes`

test/tests/benchmark/runtime_tests.py
```diff
@@ -0,0 +1,118 @@
...
+from ansible_control import ansibleControl
+import json
+import time
+
+from data_display import DataDisplay
+
+LOG = Log(__name__)
+
+@test(groups=['runtime.tests'])
+class RuntimeTests(object):
```
jlongever: How come these are reimplementing tests already defined in the other test groups? We should be running those.

test/tests/benchmark/__init__.py
```diff
@@ -0,0 +1,31 @@
+# import tests
+import time
+import os
+
+from runtime_tests import RuntimeTests
+tests = [
+    'runtime.tests'
```
jlongever: This should probably be `benchmark.tests` that have test group dependencies.

test/tests/benchmark/runtime_tests.py
```diff
@@ -0,0 +1,118 @@
...
+        Nodes().api1_1_nodes_get()
+        nodes = loads(self.__client.last_response.data)
+        count = 0
+        for n in nodes:
+            type = n.get('type')
+            if type == 'compute':
+                count += 1
+        return count
+
+    @test(groups=['runtime.tests', 'check_precondition'])
```
jlongever: Instead of reimplementing test cases here, add a `depends_on_groups=['api-v1.1', 'api-v2.0', 'api-redfish-1.0']` or set a subset of test groups that you want to benchmark.

test/run.py
jlongever: We can specify the groups to run during smoke testing,  but I'd prefer you create a `benchmark.py` that registers this group and can be executed independently of `run.py`.

iceiilin:
@WangWinson Jinja2 has been installed without adding this entry to requirements.txt. Ansible package should add this dependency when it is required.

iceiilin:
@WangWinson @jlongever You mentioned leveraging existing methods for footprint test scenario. I thought about it but would prefer to implement it separately at current time to decouple their dependencies. Further we may think about abstracting common methods to a shared library.

jlongever:
@linlynn Adding current test dependencies should be trivial.  It's a :-1:  for me if we duplicate the same tests.

iceiilin:
@jlongever it's not the same tests. In benchmark group, the tests need to discover/bootstrap **12** nodes and **measure footprint data during this period**. while smoke test is based on **api level** verification and doesn't cover this scenario. 

The only thing in common I can think of is some methods (like handle_graph_finish()), but I would prefer not to import modules in smoke test to use these methods since it brings dependencies, and the latter might change with smoke test.

For these common method, I propose to move them to a common library when refactor in the future thus no code duplication then.

WangWinson:
I agree @iceiilin if it's for different test scenario. :+1: with a story to refactor the common method.

jlongever:
@iceiilin It would cover that scenario...  Your test_check_precondition is identical to test_nodes_discovery, you can reuse 'nodes.discovery.test' to bootstrap/discover/benchmark your 12 nodes (or reuse it in your poller test). I'm not sure what you mean by _"while smoke test is based on api level verification"_.  These aren't just single purpose "smoke tests", they are all functional based.  What is your plan for running functional tests without invoking RackHD APIs?   If you're plan is to create something completely independent from these RackHD integration test, then maybe benchmark belongs somewhere else?  Maybe RackHD/benchmark?

iceiilin:
@jlongever I would not prefer to add dependency between tests, and vote for creating common files in modules/ that can be imported whenever it is needed. Currently the relationship is simple, and poller_test in benchmark can use nodeTests for its precondition check, so I changed to import this package in my poller_test.py. Please help to review.

I will add delete-rediscovery process for benchmark, and as far as I know, @anhou is adding bootstrap case for feature test. If these three test suites (I call it suite here for api/, benchmark/ and feature/ if there will be one) need to use tests from other suite, it might not be quite convenient to import that package and might possibly create "import loop". 

I will continue to write other cases for benchmark/ and compare with api/. If they share the common method, I can move it to modules/. What's your opinion about that?

jlongever:
@linlynn I think there is confusion about the /tests/api directory.  That directory is there to just isolate API versions that the functional tests run against;  it contains all the test logic.  I'm assuming when you want to benchmark or profile the system you would invoke an API (e.g. kick off a workflow against multiple nodes, etc) and monitor the system in parallel.  I created a worker class in /modules/worker.py so you can spawn worker threads (i.e. performance monitors, AMQP listeners, etc). 

The /modules directory is just a location to store common classes/methods that are not tied to specific API versions or test logic.  IMHO we shouldn't want to deviate from the normal integration test suite when profiling the system; keep the test cases consistent while profiling;  as functional tests evolve so do the benchmark metrics.  Any new functional tests that are needed for your benchmarks but are not available today should be developed under the /api test classes.  The good thing about Python proboscis is we can create group hierarchies and dependencies.  So we can create a "smoke.tests" group that only run specific test cases;  and create "benchmark.tests" group that  run specific test cases for your benchmarks;  same for regression "regression.tests" that would run everything.

cgx027:
There are some great email discuss between @jlongever and @iceiilin , just re-post to github for the community awareness that may help understand the PR. 

------------------This is a line breaker------------------------------

Hello Joel,

I read your comments in benchmark test PR and read through codes. It is great to know your design purpose about RackHD/test.

Please correct me if my understanding is correct: All test cases can be added in api/, and we assign test groups to these cases so that we can run them selectively. We can use modules/worker to create a thread for benchmark.

I tried this solution today, and find some questions that need your help.

1.	How to define different “depend groups” for one test case.
For example, the case below will be run in two groups with different dependent group. But assigning the second dependency causes both two groups to run it. I can try to dig further in Proboscis docs, but till now no related info found.
                [JL]  Is delete_all_active_workflows a test case or just a common utility method?  
I don’t think you can get logic out of the proboscis test decorator.  Perhaps you should define a common delete_all_active_workflows() method outside of a test class to be called explicitly from other test cases that depend on it?  Then you can set those test cases’ unique dependency there: 

    ```
      @test(groups=[‘testcase.test’], depends_on_groups=[‘benchmark.poller.precondition’]
      def testcase1():
Workflows().delete_all_active_workflows()
                  ... 

     @test(groups=[‘testcase.test’], depends_on_groups=[‘some.other.dependency’]
     def testcase2():
     Workflows().
    ```
image
2.	How to communicate between benchmark thread and test running thread
When test is running, benchmark thread needs to log some critical time point (like begin/end of discovery). Though these two can be monitored through AMQP queue, a general method that can be easily extended might be better for future usage.
3.	How to run post-test operation in benchmark thread
After the completion of benchmark test case, benchmark thread needs to collect footprint data from RackHD. But TestProgram().run_and_exit() exits the whole process directly, and there is no method like run() (without exit) in proboscis/case.py.
[JL] Maybe creating a tests/benchmark/benchmark_tests.py would be better—to avoid duplicating it in each api/<version>/ directory?   For each benchmark test case you can do your setup work, execute selected test cases (imported from /api test module),  and then capture your results at the end.  Thoughts?

If there is no convenient solution for question 2. and 3, what do you think if I create some test cases in api/ that is used for benchmark only (meaning adding benchmark-related operation in the test case)? 

Thanks,
Lyne

test/tests/benchmark/utils/time_logger.py
```diff
@@ -0,0 +1,61 @@
+import os
+import time
+import json
+
+"""
+Class to manage time logger file
+"""
+
+class timeLogger(object):
```
jlongever: This probably belongs in /modules/logger.py as an extension of the Log() class?

test/tests/benchmark/utils/ansible/files/collect_data.sh
```diff
@@ -0,0 +1,97 @@
...
+
+function stop_atop_collection()
+{
+    terminate_process $atop_procid
+
+    # Get resource summary at last
+    atop 1 1 -oa | grep "elapsed\|Kbps" | grep -v "NET" >> ${log_path}/summary.log
+}
+
+function start_mongod_collection()
```
jlongever: Keep in mind that we may need to support different DB services in the future.

test/tests/benchmark/utils/time_logger.py
iceiilin: I will think about that. It is a single file about benchmark case information (perhaps I need to change the class name to make it more clear), about the interval of data collection, every time point that needs to be logged for benchmark event, which is put in the same folder of footprint data collected from RackHD machine (~/benchmark/<timestamp>/<case name>/ in the machine running "python benchmark.py", by default). I prefer to isolate this method in benchmark/ so that it would not confuse other users. Thoughts?

test/tests/benchmark/utils/ansible/files/collect_data.sh
iceiilin: I am aware of this discussion in google group. I will change the name to make it more extensible like "start_db_collection". Once the data base changed, I could only update the subsequent shell scripts. Thanks.

test/tests/benchmark/utils/time_logger.py
jlongever: This seems like common functionality; although being introduced for benchmarking it could potentially be used in other places?  Either way, I would move this into /modules/logger.py and look at leveraging the Python Logging library handler objects: https://docs.python.org/2/library/logging.html#logging.Logger.addHandler.

iceiilin: I read logging manual and tried on my side, and it turns out it contains many customized operations for benchmark. I think it would be better to keep it inside benchmark/. 

The case info file created by timeLogger would be:
```
{
    "time marker": {
        "start": "2016/03/25 05:13:36",
        "end": "2016/03/25 05:13:45"
    },
    "interval": 1,
    "case name": "poller",
    "log path": "20160325-051329"
}
```
The reasons not to move it to logger is:
1. logging.handler, formatter and filter cannot write the log in this format.
I did experiments and the log can only be one json object per log item, and cannot be piled as one json object. Though changing benchmark parser to handle a different format is viable, the efforts exceed that in combining timeLogger to modules/logger.
2, write_interval(), write_log_path() and others are heavily customized for benchmark, and some of them are executed when initializing the log file. It might not be suitable to put in modules/logger as common methods.

test/tests/api/v1_1/poller_tests.py
```diff
@@ -0,0 +1,60 @@
+from config.api1_1_config import *
+from config.amqp import *
+from modules.logger import Log
+from on_http_api1_1 import NodesApi as Nodes
+from on_http_api1_1 import PollersApi as Pollers
+from on_http_api1_1 import TemplatesApi as Templates
+# from on_http_api1_1 import rest
```
jlongever: remove unused imports.

test/config/benchmark_config.py
```diff
@@ -0,0 +1,41 @@
+from settings import *
+
+ANSIBLE_AUTH_FILE = '.ansible_auth'
+
+# Obfuscate credentials
+def set_ansible_auth(local_user, local_pwd, host_ssh_port, host_ssh_user, host_ssh_pwd):
```
jlongever: You could reuse the settings.py b64 methods:
https://github.com/RackHD/RackHD/blob/master/test/config/settings.py#L26

Maybe rename them to something generic like `set_b64_creds()`.

test/config/benchmark_config.py
```diff
@@ -0,0 +1,41 @@
...
+        out = \
+        'local_user="{0}"\n'.format(lu) + \
+        'local_pwd="{0}"\n'.format(lp) + \
+        'host_ssh_port="{0}"\n'.format(rport) + \
+        'host_ssh_user="{0}"\n'.format(ru) + \
+        'host_ssh_pwd="{0}"\n'.format(rp)
+        file.write(out)
+
+# Unobfuscate credentials
+def get_ansible_auth():
```
jlongever: Same,  reuse settings.py method: https://github.com/RackHD/RackHD/blob/master/test/config/settings.py#L36

jlongever:
Jenkins, test this please

jlongever:
some minor comments but otherwise +1

iceiilin:
Jenkins, test this please

PR #131, opened 2016-03-09
adding maven and swagger to vagrant install

heckj:
- including maven and JDK into vagrant instance
 - added 'swagger' role that installs swagger code gen tools
 - extending local install script to autogen the python code libraries


anhou:
:+1: just need resolve some code conflicts

heckj:
damnit, I destroyed all my work.. Will redo and add again later...

PR #133, opened 2016-03-09
Assert 400 error on duplicate lookup entry creation

benbp:
This is a bad request, not an internal server error. Now that we properly handle the error, return the right response code.

Requires https://github.com/RackHD/on-core/pull/99, passing locally.

@RackHD/corecommitters @zyoung51 @tldavies

### DIFF
--- a/test/tests/api/v1_1/lookups_tests.py
+++ b/test/tests/api/v1_1/lookups_tests.py
@@ -105,7 +105,7 @@ class LookupsTests(object):
         try:
             Lookups().lookups_post(self.lookup)
         except Exception,e:
-           assert_equal(500,e.status, message = 'status should be 505')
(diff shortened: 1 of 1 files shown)


heckj:
May consider `409` instead of 400 here for better semantic meaning. 

:+1:

heckj:
(reference: http://www.restapitutorial.com/httpstatuscodes.html)

benbp:
@heckj reading through that description, it specifies conflicts with the current state of a resource, whereas I take this route/code to mean "create a new resource, no you can't" rather than "update this resource, no there's a conflict" ?

DavidjohnBlodgett:
:+1:

anhou:
:+1:

anhou:
merge this PR to let Jenkins pass

PR #135, opened 2016-03-11
python, maven and swagger additions

heckj:
adding in python virtualenv, maven and swagger components to VM for local dev efforts


heckj:
@anhou reset from the branch that I previous destroyed...

DavidjohnBlodgett:
Thanks for the addition of instructions for the python bits^^ :+1:

anhou:
:+1:

PR #153, opened 2016-03-29
Fetch undionly.kpxe from ipxe

anhou:
related PR:
https://github.com/RackHD/on-imagebuilder/pull/32
https://github.com/RackHD/on-dhcp-proxy/pull/30

@RackHD/rackhd_dev @RackHD/corecommitters

### DIFF
--- a/docker/files/Dockerfile
+++ b/docker/files/Dockerfile
@@ -8,6 +8,8 @@ ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail-efi64-snpo
     /RackHD/downloads/monorail-efi64-snponly.efi
 ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail.ipxe \
     /RackHD/downloads/monorail.ipxe
+ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail-undionly.kpxe \
+    /RackHD/downloads/monorail-undionly.kpxe
 
 ADD https://bintray.com/artifact/download/rackhd/binary/syslinux/undionly.kkpxe \
     /RackHD/downloads/undionly.kkpxe
--- a/packer/ansible/roles/images/tasks/main.yml
+++ b/packer/ansible/roles/images/tasks/main.yml
@@ -25,6 +25,7 @@
            force=yes
   with_items:
    - monorail.ipxe
+   - monorail-undionly.kpxe
    - monorail-efi32-snponly.efi
    - monorail-efi64-snponly.efi
 


yyscamper:
:+1:

stuart-stanley:
I guess I have the same question as JoeH over in the on-imagebuilder PR: what's the story on undionly.kpxe and undionly.kkpxe? I know we tripped over the two actually being different recently, but I still don't know *why* :)

heckj:
fwiw: we learned empirically that the ESXi bootloaders will work with syslinux undionly.kkpxe, but not with the ipxe one. There's implications/conventions to the suffixes that are relevant. From http://forum.ipxe.org/showthread.php?tid=6989 (http://etherboot.org/wiki/gpxe_imagetypes):

- `.pxe` is an image designed to be chain loaded, unloading both the underlying PXE and UNDI code sections
- `.kpxe` is a PXE image that keeps UNDI loaded and unloads PXE
- `.kkpxe` is a PXE image that keeps PXE+UNDI loaded and return to PXE (instead of int 18h). From here

And there is also .kkkpxe - "The ipxelinux.0 build process should therefore now use undionly.kkkpxe instead of undionly.kkpxe."

Most often, you will want either .pxe, .kkpxe (for when you are using cached credentials, iirc), and .kkkpxe (usually the last two are in combination with pxelinux).

stuart-stanley:
I don't even know what to think about ipxe's and syslinux's thought process in naming images with subtly different behaviors (and more importantly, the hidden application of those subtle differences!) with such, um, descriptive file names... kkkpxe? Egads. :)

More relevant, I think this means we should have at least the ones we have been known to use (kkpxe and kpxe) available in RackHD, so +1 from me.

Note: it would be nice to merge this before https://github.com/RackHD/on-imagebuilder/pull/31, so this can be sucked up into versioned packaging at the same time instead of later in an extra pass. Not required by any stretch of imagination, but it would shave an hour of boring work out of my personal schedule :)

anhou:
Thanks @heckj  describing the difference of '.kpxe' '.kkpxe' '.kkkpxe'  , 
 @stuart-stanley and more, it will be crazy when we see both syslinux and ipxe have undionly.kpxe, but their behavior are different sometime. for example, from previous experience of debugging chainloading mboot.c32 when installing ESXi, even 'mboot.c32's boot options are same, syslinux succeed, but ipxe fail. it looks like they have different environment variables to boot mboot.c32 after some code hacking.  A little complex here, so agree with you that it should have the ones at least used by RackHD now.

johren:
+1

kostew:
Current docker images do not contain 'undionly.kpxe' which is required for a HW configuration under test in MN to work.  Would like to see this merged to unblock my test configuration.

benbp:
@rolandpoulter see above comment from @kostew ^^

benbp:
:+1:

Issue #151, opened 2016-03-28
404 response on delete active workflow.


jlongever:
Observed in both 1.1 and 2.0 API's;  following the workflow engine redesign the integration tests started reporting random 404 errors on a DELETE to /node/:id/workflows/active after starting a discovery graph workflow.  Checks to monitor the status following a POST to workflows indicate the workflow is either 'pending' or 'running';  when a DELETE is sent the following error is returned:

```
ApiException: (404)
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Content-Length': '62', 'X-Powered-By': 'Express', 'Connection': 'keep-alive', 'ETag': 'W/"3e-uVr3eF51jj1og38xQuRGng"', 'Date': 'Tue, 22 Mar 2016 07:28:50 GMT', 'Access-Control-Allow-Origin': '*', 'Content-Type': 'application/json; charset=utf-8'})
HTTP response body: {"message":"NotFoundError: No active workflow found for node"}
```

This can be reproduced locally after multiple attempts.   

Looking at the workflow API service in on-http;  there isn't much to distinguish between the two error scenerios in the promise chain.  But after some digging the second .then() in the chain is what's being hit indicating the cancel request is randomly responding with an unfulfilled graph:
```
        .then(function(graphId) {
            if (!graphId) {
                throw new Errors.NotFoundError('No active workflow found for node');
            } else {
                return;
            }
        });
```

A [commit] (https://github.com/RackHD/RackHD/commit/4c7e4b2daf030b9b41128d70868e3adb05157730)  was made to help mitigate these random failures in the functional integration tests;  but some additional evaluation around this is probably in order to ensure everything is behaving as designed.

@RackHD/corecommitters FYI

jlongever:
Put up this PR https://github.com/RackHD/on-http/pull/192/files to help provide some more detail when a cancel fails.

benbp:
That route code was added to handle possible race conditions where the workflow was completed after the API call/node query but before the delete call. Perhaps there is a bug with that handling, OR the workflow in the test is actually completing?

Discussion originally at https://github.com/RackHD/on-http/pull/112#discussion_r52049893

jlongever:
@benbp The functional test ensures the state is running or pending prior to requesting the cancel.  The time between status checking and cancellong is immediate. How long does the Graph.Discovery workflow take to complete?  Is it possible the findAndModify() isn't updating the document on subsequent workflow request?

benbp:
[...]

PR #162, opened 2016-04-04
Fix for delete workflow test failures

brianparry:
Possible fix for https://github.com/RackHD/RackHD/issues/151

### DIFF
--- a/test/tests/api/v2_0/nodes_tests.py
+++ b/test/tests/api/v2_0/nodes_tests.py
@@ -89,10 +89,10 @@ class NodesTests(object):
                 assert_is_not_none(status)
         return status
         
-    def __post_workflow(self, id, graph_name, data):
+    def __post_workflow(self, id, graph_name):
         status = self.__get_workflow_status(id)
         if status != 'pending' and status != 'running':
-            Api().nodes_post_workflow_by_id(identifier=id, name=graph_name, body=data)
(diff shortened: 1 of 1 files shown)


brianparry:
Tests should pass now that https://github.com/RackHD/on-http/pull/203 is merged, so test this please

brianparry:
Looks like test was aborted.  test this please

test/tests/api/v2_0/nodes_tests.py
```diff
@@ -89,10 +89,10 @@ def __get_workflow_status(self, id):
                 assert_is_not_none(status)
         return status
         
-    def __post_workflow(self, id, graph_name, data):
+    def __post_workflow(self, id, graph_name):
         status = self.__get_workflow_status(id)
         if status != 'pending' and status != 'running':
-            Api().nodes_post_workflow_by_id(identifier=id, name=graph_name, body=data)
+            Api().nodes_post_workflow_by_id(identifier=id, name=graph_name, body={'name': graph_name})
```
benbp: Do we need name in both the query param and data body?

brianparry: @benbp We should only need the body, but the generated client library complained when I omitted the name keyword argument.  We need to rebuild the client library and make a minor change to the swagger file to fix this.  I put this PR up as-is to try and get the tests running again, since they are failing most of the time now.

benbp:
:+1:

dalebremner:
+1

PR #204, opened 2016-04-28
Add benchmark node discovery case

iceiilin:
Add node discovery case for footprint benchmark. Modification and test is still ongoing.
@jlongever, please help to review the structure first in this PR.

Some notices:

1. Add function tests in v1_1/workflows_tests.py and nodes_tests.py
In current group-name convention, these tests will also run in Jenkins CI tests, which is not expected (there are node deletion in the newly-added case, https://github.com/RackHD/RackHD/pull/204/files#diff-2d1662f00dd71c100c268af11a72f1ccR399).
If you agree, I will add group names like "jenkins test" to distinguish those running in CI tests.

2. Some APIs is inevitably used in benchmark/api_v1_1_tests.py
Though both of us agree to constrain function tests in tests/ repo, benchmark/ repo still need to use API for monitoring and logging some status for benchmark data, like the number of nodes (https://github.com/RackHD/RackHD/pull/204/files#diff-25464a03b247127ac3d853933f715647R39), and the finishing time for every node's discovery (through AMQP monitoring, https://github.com/RackHD/RackHD/pull/204/files#diff-25464a03b247127ac3d853933f715647R109).

3. I plan to run TestProgram.run_and_exit() more than once to ensure the test case sequence.
The expected sequence of the two benchmark tests is:
(poller_running -> benchmark_poller) -> (poller_running -> **reboot_node -> delete_node** -> benchmark discovery)
But actually it is:
poller_running -> **reboot_node -> delete_node** -> benchmark_poller -> benchmark discovery
I could add dependencies between benchmark_poller and benchmark_discovery to avoid it, but it would lose the flexibility to run the discovery case along. Separate TestProgram process is a feasible choice.


jlongever:
@iceiilin  I'd prefer to not add environment specific tags such as 'jenkins' to the framework.    At this point we should probably go ahead and create the 'smoke-test', 'regression-test' and 'benchmark-test' groups and specify the tests we want those groups to include.  This should be a prerequisite to your benchmark work.

iceiilin:
@jlongever I am afraid that adding group names such as "smoke-test" or "integration-test" in case's decorator will be an easy way to tell proboscis which cases to run for a specific need. 

In your previous comment, do you mean that we may create a group and include cases that are needed to it? I tried today the following structure, and it seems more efforts than expected:
in *api/.../__init__.py*:
```
nodes_smoke_tests = ['test-nodes', 'create-node', ...]
obm_smoke_tests = [...]
...

smoke_tests = nodes_smoke_tests + obm_smoke_tests + ...
```
in *run.py*,
```
 register(groups=['api-v1.1'], depends_on_groups=api_1_1.smoke_tests)
```
It can designates which cases to run for smoke test, but other cases (newly added for verifying) in nodes_tests that are not needed for smoke test will still be run since obm smoke tests depends on the whole group of "nodes.test". The same scenario can be applied to other group dependencies. We can definitely decouple the dependencies to a set of cases rather than another whole group, but it might bring more efforts to figure it out.

So I would prefer to add group names such as "smoke-test" or "integration-test" in case's decorator. What do you think about that?

jlongever:
@iceiilin You could move your nodes test out of `nodes.test` group into a different class/group and then set the specific `nodes.test` dependencies needed using `depends_on_groups=[]` on you new class groups.   Alternatively we could define each specific test to be part of a group of groups with the top group being `smoke.tests`, `regression.tests` and `benchmark.test` and the subgroup being the specific tests need to be run.  https://pythonhosted.org/proboscis/#groups-of-groups

iceiilin:
test this please

iceiilin:
@jlongever I agree with you about creating new classes/groups for non-smoke test. I created a discovery_test in api/v1_1/ that will be used in benchmark. Please help to review and approve if possible.

iceiilin:
test this please

iceiilin:
test this please

iceiilin:
test this please

test/.passwd
```diff
@@ -1,2 +1,7 @@
-BMC_USER="YWRtaW4="
-BMC_PASS="YWRtaW4="
\ No newline at end of file
+BMC_USER=""
```
jlongever: Why was the default BMC user/password deleted?

jlongever: This will cause issues in the smoke tests.

iceiilin: Right, I found this problem too by examining Jenkins console output. Thanks.

iceiilin:
test this please

iceiilin:
test this please

test/benchmark.py
```diff
@@ -26,19 +37,12 @@ def run_tests(api_ver, selected):
...
-                group_selected = True
-            if op == "-h":
-                print "Usage: benchmark.py [--api_version|--group]"
-                exit()
-    except getopt.GetoptError:
-        sys.exit("option or arg is not supported")
+    for arg in sys.argv[1:]:
+        if arg[:14] == '--api_version=':
+            # Remove this arg from array to prevent TestProgram processing it
+            sys.argv = filter(lambda x: x != op+'='+value, sys.argv)
```
WangWinson: where is `op` and `value` defined ?

test/benchmark/api_v1_1_tests.py
```diff
@@ -0,0 +1,137 @@
...
+
+class BenchmarkTests(object):
+    def __init__(self, name):
+        ansible_ctl.render_case_name(name)
+        self.__data_path = ansible_ctl.get_data_path_per_case()
+        self.case_recorder = caseRecorder(self.__data_path)
+        self.client = config.api_client
+        self.node_count = 0
+
+    def prepare_case_env(self):
```
WangWinson: Best practice: 
Looks like `BenchmarkTests` is the base class. The real test case class will extend it , right ?
So the base function which is for inherit using need to be start with single underscore '_' to imply it's `protected` function

test/benchmark/api_v1_1_tests.py
```diff
@@ -0,0 +1,137 @@
...
+        self.node_count = self.check_compute_count()
+        self.case_recorder.write_interval(ansible_ctl.get_data_interval())
+        self.case_recorder.write_start()
+        self.case_recorder.write_node_number(self.node_count)
+
+        assert_equal(True, ansible_ctl.start_daemon(), \
+                    message='Failed to start data collection daemon!')
+
+    def collect_case_data(self):
```
WangWinson: Same as above

test/benchmark/api_v1_1_tests.py
```diff
@@ -0,0 +1,137 @@
...
+        assert_equal(True, ansible_ctl.collect_data(), message='Failed to collect footprint data!')
+        self.case_recorder.write_end()
+
+        # parser.parse(self.__data_path)
+
+    def check_compute_count(self):
```
WangWinson: `private` function ?

test/benchmark/api_v1_1_tests.py
```diff
@@ -0,0 +1,137 @@
...
+            type = n.get('type')
+            if type == 'compute':
+                count += 1
+        return count
+
+
+@test(groups=["benchmark.poller"])
+class BenchmarkPollerTests(BenchmarkTests):
+    def __init__(self):
+        BenchmarkTests.__init__(self,'poller')
```
WangWinson: code style :  space needed after comma.
Do we run pylint ?

test/benchmark/api_v1_1_tests.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @test(groups=["test-bm-discovery"],
+            depends_on_groups=["test-bm-discovery-prepare", "test_discovery_delete_node"])
+    def test_discovery(self):
+        """ Wait for discovery finished """
+        self.case_recorder.write_event('start all discovery')
+        self.__task = WorkerThread(AMQPWorker(queue=QUEUE_GRAPH_FINISH, \
+                                              callbacks=[self.handle_discovery_finish]), \
+                                   'benchmark discovery')
+        def start(worker, id):
```
WangWinson: `id` not using ?

test/benchmark/api_v1_1_tests.py
```diff
@@ -0,0 +1,137 @@
...
+            worker.start()
+        tasks = WorkerTasks(tasks=[self.__task], func=start)
```
WangWinson: Does it possible to replace `func=start` with `func = lambda w: w.start()` , so that we could omit the definition `def start(worker, id):`

test/tests/api/v1_1/discovery_tests.py
```diff
@@ -0,0 +1,76 @@
...
+            if n.get('type') == 'compute':
+                id = n.get('id')
+                assert_not_equal(id,None)
+                status = self.__get_workflow_status(id)
+                timeout = 20
+                while status == 'pending' and timeout != 0:
+                    LOG.warning('Workflow status for Node {0} (status={1},timeout={2})'.format(id,status,timeout))
+                    status = self.__get_workflow_status(id)
+                    sleep(5)
+                    timeout -= 1
```
WangWinson: so this means `retry` times ?

WangWinson:
:+1: after comments addressed

test/benchmark/api_v1_1_tests.py
iceiilin: Two arguments are required in WorkerTask, and I would prefer to use "def start()" to make it more human readable than lambda.

WangWinson: OK, make sense

iceiilin:
@WangWinson great comments, addressed.

yyscamper:
+1

PR #206, opened 2016-04-29
Benchmark/add html pages

cgx027:
Add html files used to show footprint benchmark data. 

summary.html is the portal page. 

A follow PR will be filed for copying the html files into the report folder after test runs.


cgx027:
@RackHD/corecommitters  @yyscamper @anhou @WangWinson @wuy1 @rolandpoulter

cgx027:
test this please

jlongever:
I think this web content belongs in it's own repository or maybe on-tools?  The test location is reserved for functional test content.

heckj:
@jlongever has a good point - @cgx027 what's your thought/plan for how and where this information will be displayed? Can the presentation of the information be separated from the tests themselves so that we don't end up cluttering the benchmark test directories with presentation details?

cgx027:
@jlongever @heckj , You guys have a good point. Those HTMLs are supposed to be copied to the output file where benchmark log data resides, so that it can take those log data and show to the users. I would agree on-tools seems like a better place. 

I will close this PR and create a new PR against on-tools to host those HTMLs.

cgx027:
PR https://github.com/RackHD/on-tools/pull/22 created hence retire this PR.

PR #212, opened 2016-05-03
Modify footprint benchmark parser.py

cgx027:
Modify footprint benchmark parser.py to: 

1. further scan log folder to show not only time stamp but also case name
2. Write process list to log file for use by front end html pages

@RackHD/corecommitters @iceiilin @WangWinson


anhou:
+1

yyscamper:
:+1

WangWinson:
:+1:

PR #215, opened 2016-05-03
Add AMQP listener mode utility

jlongever:
Pass --amqp option with exchange/queue parameters to listen continuously on exchange

Optional tool.  Take it or leave it.

@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2 @geoff-reid @keedya

### DIFF
--- a/test/config/amqp.py
+++ b/test/config/amqp.py
@@ -14,9 +14,15 @@ QUEUE_SDR_RESULT        = Queue('ipmi.command.sdr.result',
 QUEUE_CHASSIS_RESULT    = Queue('ipmi.command.chassis.result',
                                 EXCHANGE_TASK,
                                 routing_key='ipmi.command.chassis.result.*')
-
+                                
 # Event exchange queues
 EXCHANGE_EVENT          = Exchange('on.events', type='topic')
 QUEUE_GRAPH_FINISH      = Queue('graph.finished',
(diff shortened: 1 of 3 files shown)


tannoa2:
test this please

zyoung51:
Should optional utilities/tools be in the on-tools repo?  +1 from me either way.

jlongever:
@zyoung51 probably, but since it reuses classes defined in test I just made it an option.

benbp:
@jlongever yeah we have a similar utility at https://github.com/RackHD/on-tools/blob/master/dev_tools/sniff.js, though there seems to be a bit more functionality here. Might be worth a README mention if you're going to add this?

👍

iceiilin:
+1 This tool will be useful

heckj:
👍 looks great, thinking maybe we should kill the tooling in on-tools/dev_tools/sniff.js - especially where this could be leveraged to get better insight into the system for testing, which I suspect is some of the desire/drive here (at least that's my guess based on it being in `tests` - was that part of the intent @jlongever ?)

jlongever:
@heckj tbh I didn't put too much thought into it,  this was something I used while testing out new poller alerts and figured it would be useful under the integration test framework.  Same with the  `--httpd` option as well,  was used for validating Redfish EventService webhook functionality.

PR #223, opened 2016-05-12
[redfish] RackHD Manager support

zyoung51:
The RackHD ethernet interface list returns the os named interfaces
instead of an index number.  Change the test to use the enumerated
values returned by the list interface command rather than the hard
coded '0' value.


zyoung51:
@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @jlongever

zyoung51:
Test this please

PR #224, opened 2016-05-12
Create smoke-test and regression test groups

jlongever:
Pass --group option:
  `--group=regression-tests`
  `--group=smoke-tests`
With no group option specified defaults to smoke-tests

@RackHD/corecommitters @uppalk1 @zyoung51 @keedya @tannoa2


keedya:
:+1:

brianparry:
:+1:

PR #232, opened 2016-05-17
add benchmark bootstrap case; add benchmark operation in readme

iceiilin:
Add benchmark bootstrap case for installing CentOS 7 in os_install_tests.py
Add benchmark operation in README.

@jlongever @WangWinson 

The image address is prepared as a proxy server in Ansible playbook prepare_bootstrap_env.yml. In this playbook, the config.json is copied into RackHD and the service is restarted before bootstrap case. This configuration will be restored after the case finished. @WangWinson suggested to add this config.json into the Vagrantfile directly, but benchmark would be used in any RackHD, and it might be more flexible to config it runtime.

The tests in os_install_tests (only post install OS workflow without tracking its status) is sufficient now for benchmark case, since benchmark would monitor the status itself. For regression test, we could add the result checking tests in the future, without interfering benchmark test scenario.


test/benchmark.py
```diff
@@ -13,18 +13,21 @@ def run_tests(api_ver, selected):
 
     register(groups=['poller'], depends_on_groups=['benchmark.poller'])
     register(groups=['discovery'], depends_on_groups=['benchmark.discovery'])
+    register(groups=['bootstrap'], depends_on_groups=['benchmark.bootstrap'])
 
     if selected == False:
         # Three test groups need to run sequentially,
         # while proboscis schedules tests in different group at a mixed manner.
-        # Adding dependencies among groups is not prefered since they also can be executed along.
+        # Adding dependencies among groups is not prefered since they also can be executed separatly.
```
WangWinson: typo: separately

jlongever:
@iceiilin @leoyjchang @amymullins We are working on a story to add OS install functional tests as part of regression test effort: https://www.pivotaltracker.com/story/show/119308071

@iceiilin Do you have any plans on adding other OS installer workflows?

test/tests/api/v1_1/os_install_tests.py
```diff
@@ -1,27 +1,68 @@
...
+                while status != 204 and retries != 0:
+                    LOG.warning('Workflow status for Node {0} (status={1},retries={2})'.format(id,status,retries))
+                    sleep(1)
+                    retries -= 1
+                    Nodes().nodes_identifier_workflows_active_get(id)
+                    status = self.__client.last_response.status
+
+                assert_equal(204, status, message = 'status should be 204')
+
+                Nodes().nodes_identifier_workflows_post(id,name=graph_name,body=body)
```
jlongever: Do you intend to listen on the AMQP event `graph.finished` exchange and wait for completion of the graph?  How do you know the graph was successful?

jlongever: Nevermind I see you do it here: https://github.com/RackHD/RackHD/pull/232/files#diff-25464a03b247127ac3d853933f715647R158

So what about other test groups that may want to reuse the OS install test?   Probably want to keep the AMQP exchange setup/handler under the same test case. You could always add a function parameter to pass in different handler functions if necessary.

test/README.md
```diff
@@ -1,76 +1,135 @@
...
+
+Footprint benchmark collects system data when running poller, node discovery and bootstrap.
+Details can be found in WIKI page:
+[proposal-footprint-benchmarks](https://github.com/RackHD/RackHD/wiki/proposal-footprint-benchmarks)
+
+###Precondition
+
+In RackHD, compute nodes have been discovered, and pollers are running
+
+No external AMQP queue with the name "graph.finish" is listening RackHD
```
jlongever: Why is this a precondition?

test/README.md
```diff
@@ -1,76 +1,135 @@
...
+
+###Settings
+
+Aside from Optional settings in the section above,
+following parameters are also required at the first time user issuing the test,
+and stored in .passwd
+
+    localhost username and password: for the machine running the test
+    RackHD ssh port, username and password
+    image server: remote address of CentOS image
```
jlongever: This should probably be an env variable passed in?

amymullins:
@iceiilin  and @leoyjchang , As @jlongever mentions above, there is a story currently in development creating regression tests for OS installs (ESXi, CentOS, Suse, and Ubuntu).  

We can continue this effort if these benchmark tests can't be expanded to cover the additional OSs or if they can't be updated such that other test groups can reuse the OS Install tests.  Reference some recent work to create sub - test groups in the functional test suite:  https://www.pivotaltracker.com/projects/1492890/stories/119308107

leoyjchang:
@amymullins we only planned this CentOS bootstrap test used for benchmark testing. You can continue developing bootstrap regression test for all supported OS. I would suggest to look into whether any code here can be reused in your bootstrap test work.  BTW, I am glad to see full bootstrap regression test development is underway, which is one important task planned in my mind especially we encountered a bunch of bootstrap ODRs.

test/tests/api/v1_1/os_install_tests.py
iceiilin: @jlongever Do you plan to install OS in nodes one by one (meaning install OS in one node, listen to AMQP until it finishes, and then install OS in another node)? If you agree to begin installing OS at the same time, you may add another case focusing on listening AMQP, which depends on the previous issuing workflow test.

test/README.md
iceiilin: ENV variable is another choice to specify this setting, and I view these two options of implementations as equal. What's your opinion?

test/tests/api/v1_1/os_install_tests.py
iceiilin: @jlongever I didn't add AMQP listener here in os-install-test since benchmark would do some special operation in handle_graph_finish, like https://github.com/RackHD/RackHD/pull/232/files#diff-25464a03b247127ac3d853933f715647R85. I thought adding AMQP listener separately in benchamark test and os install test (not implemented yet) in the following test would decouple these two test suite effectively.

test/README.md
jlongever: Env is consistent with other variables used in this framework. BMC credentials was an exception to hide their values.

test/tests/api/v1_1/os_install_tests.py
jlongever: I run them on all nodes at the same time and listen on AMQP for each node.  Feel free to review my branch https://github.com/jlongever/RackHD/commit/9c152139a0f31e792e7e1bc564a0ff1f58c4bd63

jlongever: You could easily pass in different AMQP callback handlers as a parameter to the POST workflow function.

jlongever:
Could you follow the same pattern for defining OS install methods that can be shared such as https://github.com/jlongever/RackHD/commit/9c152139a0f31e792e7e1bc564a0ff1f58c4bd63#diff-0548e5992aaf08d4ac9129f0bf93ff83R39.   Or wait for this change before adding benchmark tests for OS installs?  I will add a parameter to specify custom AMQP callbacks.

WangWinson:
@iceiilin this is the [config.json ](https://github.com/RackHD/on-build-config/blob/master/vagrant/config/config.json) I mentioned for test using. There will be no bad impact  to update it directly. Thus, you may save the reboot service effort.

test/README.md
iceiilin: It comes from the OVA case, where the conductor is also consuming "graph.finished" queue. If proboscis is also listening to this queue at the same time, "active consumers connected to a queue receive messages from it in a round-robin fashion." [from here](https://www.rabbitmq.com/consumer-priority.html). Some messages will be lost in this case for the AMQP listener here. So I listed here as a precondition to ensure no other "graph fisnihed" queue exists. We can change the queue name in config/amqp.py, but it is still valid to  list it as a precondition to avoid conflict with the new name.

iceiilin:
@WangWinson Thanks for telling me about this config.json for Jenkins test. As we talked, the scenario here is a little different: it hasn't been integrated in Jenkins, and aims to let developers run the benchmark test in their own machine. The image server address in the proxy is an input from user, and only can be uploaded to RackHD at run time.

iceiilin:
@jlongever Thanks for your information. I read your code, and found it doesn't match my test steps.
it uses post_workflows, which manipulate nodes in a serial manner:
```
post workflow to one node -> wait for it to finish -> post workflow to the next node -> ...
```
While in my case, I would prefer to manipulate them in parallel to measure the footprint consumption during this period:
```
post workflows to all nodes -> wait for their finishing
```
Will you adopt the parallel manner? If not, I would consider moving centOS installation for benchmark to another file without mixing the process with regression test.

amymullins:
@leoyjchang , sounds great! We'll continue on with the story as planned in the current sprint to cover the OS Install regression tests for CentOS, ESXi, Ubuntu and Suse.

jlongever:
@iceiilin I'd prefer to you move the CentOS install code into your benchmark class.

test/tests/api/v1_1/os_install_tests.py
iceiilin: I tried today but the handler in benchmark test needs to stop tasks that is created in post_workflow, and I didn't come up with a good solution to work around it. So an AMQP listener is still created in benchmark test to monitor the whole process.

test/README.md
iceiilin: Changed image path to ENV. I read from your code that you use a serial of ENV to set the image repo. It is convenient in Jenkins test, where the network environment is known and stable. In the benchmark case designed for all developers, the network conditions are diverses: nodes might not access the external repo. So I use http proxy to setup the repo with the IMAGE_SERVER env from users.

iceiilin:
@jlongever please help to review and approve if possible.

Moved CentOS install code to benchmark class
Copied post_workflow from your working branch, added verify logic to wait for the connection being stable without retry. https://github.com/RackHD/RackHD/pull/232/files#diff-d618aea7eb216b7b27d43c843a01227cR149

test/README.md
jlongever: I'm not sure what you mean, the env settings in my branch can use a local or proxy address.  I'm actually using a proxy repo in my testing,  nothing about it is Jenkins specific.

jlongever: The only difference is you called yours IMAGE_SERVER and I called mine by OS name I.e. RACKHD_CENTOS_REPO_PATH.

test/tests/api/v1_1/os_install_tests.py
```diff
@@ -1,27 +1,68 @@
...
-    @test(groups=['suse-install.v1.1.test'])
-    def test_install_suse(self):
-        """ Testing SUSE OS Installer Workflow """
-        pass
+    @test(groups=['centos7-install.v1.1.test'],
+            depends_on_groups=["test-node-poller", "test-bm-bootstrap-prepare"])
+    def test_install_centos7(self):
+        """ Testing CentOS 7 Installer Workflow """
+        body = {
+            "options": {
```
jlongever: Why not pass the `repo` in as an option here and override it?  Particularly if you want this to be flexible for external users.  Otherwise the user will be required to put their OS files in the hard coded path defined in the install graph.

https://github.com/jlongever/RackHD/commit/bdfa99a62beef078a92459eb3d6e8e6ecf8ae758#diff-0548e5992aaf08d4ac9129f0bf93ff83R50

iceiilin: I am considering the case where the node cannot access external repo. So I use IMAGE_SERVER env to let users put their external repo address, and render it in http proxy in config.json (https://github.com/RackHD/RackHD/pull/232/files#diff-9f377a64dbf4c6443324e8d98f5bf34aR33), so that nodes will ask the RackHD about files for installation, and RackHD works as a proxy to fetch external files.

I write in the config.json about files path for nodes, which is the same as default value in workflows, and don't need to overwrite it here in the test case.

jlongever: What if a user wants to use their own config.json?

iceiilin:
Ah, it seems I haven't reset os_install_tests.py after moving my logic into benchmark class. I will add that.

iceiilin:
@jlongever I use the same repo env as in your PR. Please help to review and approve if possible.

I will leave to you the update of env in README and post_workflow in workflows_tests to avoid some merge conflicts. It will not impact benchmark functions.

test/tests/api/v1_1/discovery_tests.py
```diff
@@ -35,29 +36,29 @@ def __get_workflow_status(self, id):
     @test(groups=['test_discovery_post_reboot'], depends_on_groups=["test-node-poller"])
     def test_node_workflows_post_reboot(self):
```
jlongever: What is the purpose of this test? it's only initializing a `workflow` variable and never POSTs it.

test/benchmark/api_v1_1_tests.py
```diff
@@ -93,50 +133,77 @@ def test_prepare_discovery(self):
...
-        BenchmarkTests.__init__(self,'bootstrap')
-        self.__task_worker = None
-        self.__discovered = 0
+        BenchmarkTests.__init__(self, 'bootstrap')
+        self.__base = os.getenv('RACKHD_BASE_REPO_URL', \
+             'http://{0}:{1}'.format(HOST_IP, HOST_PORT))
+        self.__os_repo = os.getenv('RACKHD_CENTOS_REPO_PATH', \
+            self.__base + '/repo/centos/7')
+
+    def __post_workflows(self, graph_name, body):
```
jlongever: Could this be moved into the workflows_tests.py module so other workflow related tests can pick up the retry/status checks?

jlongever: I suppose I could pull this into my PR once this PR has been merged.

test/tests/api/v1_1/discovery_tests.py
iceiilin: It's here to post this new workflow
https://github.com/RackHD/RackHD/pull/232/files#diff-ecb0f933ac478b58493da50b899c920aR58

jlongever: Oh never mind, I read the diff wrong.  thanks.

jlongever:
+1

test/benchmark/api_v1_1_tests.py
iceiilin: Great. Thanks.

yyscamper:
:+1:

PR #233, opened 2016-05-18
an new ansiable playbook: to install RackHD throughly from latest source( build deb/microkernel)

panpan0000:
comparing with the rackhd_package.yml, it will do below special: 
(1) build deb package from source code, and dpkg install it
(2) build micro-kernel/ipxe image from source code(/on-imagebuilder/), instead of downloading from bintray

===================
*Things Left before merge*  update /RackHD/on-imagebuilder pointer to latest head of on-imagebuilder repo

===================
Test -> I download a 14.04 Ubuntu vagrant box, and run this yml with below. it runs well.
`sudo  ansible-playbook -i "local," -c local rackhd_install_from_src.yml`

### DIFF
--- a/packer/ansible/rackhd_install_from_src.yml
+++ b/packer/ansible/rackhd_install_from_src.yml
@@ -1,0 +1,18 @@
+---
+- name: RackHD Throughly Installation Server
+  hosts: local
+  sudo: no
+  roles:
+    - apt
+    - build
+    - git
+    - rabbitmq
+    - mongodb
+    - node-legacy
+    - rackhd-deb-from_src
+    - isc-dhcp-server
+    - monorail
+    - snmptool
+    - ipmitool
+    - images_from_src
+    - integrationtest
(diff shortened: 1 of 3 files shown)


anhou:
@panpan0000 some concerns about this PR for its purpose and technical points. 1) why build package and dpkg install, not fetch latest codes directly. so I'm not sure the purpose.  2) when build and deploy vagrant box.  I suggest not include compiling steps. just separate compiling environment and deployment environment, it has the case that compiling in on-imagebuilder on Ubuntu 14.04 pass but 16.04 fail, and when it could be deployed to both 14.04 and 16.04. I mean these steps could be separately maintained is better. IMHO.

panpan0000:
Thanks, Andrew.
it's great idea to separate build script and deploy script .

heckj:
Let me beg that question - why is it better to have a build script vs a deploy script. I made a version to drive using the Debian packages as an alternative, but it's certainly not intended to be a formal release mechanism.

Get Outlook for iOS<https://aka.ms/o0ukef>

_____________________________

panpan0000:
Close PR.
will reconsider the motivation.

PR #234, opened 2016-05-19
Add OS installer regression tests

jlongever:
- Adds OS installer test cases for CentOS, Ubuntu, ESXi, and SUSE OS installer workflows for 1.1 API.  Runs concurrently across all discovered nodes.
- Adds workflow to remove disk MBR and partitions on each node (runs concurrently)
- Some cleanup around group naming and priorities (mainly in API1.1 AMQP and OBM tests). 
- Some POST workflow refactoring under workflows_test module.

@RackHD/corecommitters @keedya @zyoung51 @tannoa2 @uppalk1 @iceiilin

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -8,7 +8,7 @@
     virtualenv .venv
     source .venv/bin/activate
     sudo pip install -r requirements.txt
-
+    
 ## Running the tests
 
 Run Vagrant environment
@@ -36,15 +36,15 @@ Run regression tests
 
     python run.py --group=regression-tests
 
-## Optional settings
+## Optional Environment Variables
 
 Log levels
-
+    
     NOTE: CRITICAL < ERROR < WARNING < INFO < DEBUG
 
     export RACKHD_TEST_LOGLVL=[CRITICAL|ERROR|WARNING|INFO|DEBUG default=WARNING]
 
-API host/port
+API host/port 
 
     export RACKHD_HOST=[host ip default=localhost]
     export RACKHD_PORT=[host port default=9090]
@@ -53,7 +53,26 @@ AMQP URL
 
     export RACKHD_AMQP_URL=[amqp://<url>:<port> default=amqp://localhost:9091]
 
-Specify test groups
+OS Installer locations
+
+    Specify specific OS repository path:
+
+    export RACKHD_CENTOS_REPO_PATH=[location to http mirror for all CentOS images]
+    export RACKHD_ESXI_REPO_PATH=[location to http mirror for all ESXi images]
+    export RACKHD_UBUNTU_REPO_PATH=[location to http mirror for all Ubuntu images]
+    export RACKHD_SUSE_REPO_PATH=[location to http mirror for all SUSE images]
+
+    Example: export RACKHD_CENTOS_REPO_PATH=http://ip:port/repo/centos
+
+HTTP proxy configuration (optional):
+
+To define HTTP proxies for accessing remote OS image repositories, see the [RackHD Configuration howto](http://rackhd.readthedocs.io/en/latest/rackhd/configuration.html?highlight=httpProxies#rackhd-configuration)
+
+Specifying test groups
+    
+    To display the entire test plan and available test groups run:
+
+    python run.py --show-plan
 
     Use the nosetest --group option to test a specific group:
 
@@ -66,7 +85,7 @@ Specify test groups
     Run only Redfish 1.0 compliant API related tests:
     python run.py --group=api-redfish-1.0
 
-To reset the default target BMC user/password
+To reset the default target BMC user/password 
 
     NOTE: only prompts for user/password when .passwd file is missing
 
--- a/test/tests/api/v1_1/os_install_tests.py
+++ b/test/tests/api/v1_1/os_install_tests.py
@@ -1,27 +1,162 @@
 from config.api1_1_config import *
-from obm_settings import obmSettings
+from config.settings import *
+from modules.logger import Log
 from on_http_api1_1 import NodesApi as Nodes
 from on_http_api1_1 import rest
-from modules.logger import Log
+from workflows_tests import WorkflowsTests as workflows
 from datetime import datetime
 from proboscis.asserts import assert_equal
 from proboscis.asserts import assert_false
 from proboscis.asserts import assert_raises
 from proboscis.asserts import assert_true
 from proboscis.asserts import assert_not_equal
+from proboscis.asserts import assert_is_not_none
 from proboscis import SkipTest
 from proboscis import test
+from proboscis import after_class
+from proboscis import before_class
 from json import dumps, loads
+import os
 
 LOG = Log(__name__)
 
-@test(groups=['os-install.v1.1.tests'])
+@test(groups=['os-install.v1.1.tests'], \
+    depends_on_groups=['amqp.tests'])
 class OSInstallTests(object):
 
     def __init__(self):
-        self.__client = config.api_client 
-
-    @test(groups=['suse-install.v1.1.test'])
-    def test_install_suse(self):
-        """ Testing SUSE OS Installer Workflow """
+        self.__client = config.api_client
+        self.__base = os.getenv('RACKHD_BASE_REPO_URL', \
+            'http://{0}:{1}'.format(HOST_IP, HOST_PORT))
+            
+    @before_class()
+    def setup(self):
         pass
+        
+    @after_class(always_run=True)
+    def teardown(self):
+        self.__format_drives()  
+    
+    def __get_data(self):
+        return loads(self.__client.last_response.data)
+    
+    def __post_workflow(self, graph_name, nodes, body):
+        workflows().post_workflows(graph_name, timeout_sec=3600, nodes=nodes, data=body)         
+
+    def __format_drives(self):
+        # Clear disk MBR and partitions
+        command = 'for disk in `lsblk | grep disk | awk \'{print $1}\'`; do '
+        command = command + 'sudo dd if=/dev/zero of=/dev/$disk bs=512 count=1 ; done'
+        body = {
+            'options': {
+                'shell-commands': {
+                    'commands': [ 
+                        { 'command': command } 
+                    ]
+                }
+            }
+        }
+        self.__post_workflow('Graph.ShellCommands', [], body) 
+
+    @test(enabled=False, groups=['format-drives.v1.1.test'])
+    def test_format_drives(self):
+        """ Drive Format Test """
+        self.__format_drives()  
+        
+    def install_centos(self, version, nodes=[], options=None):
+        graph_name = 'Graph.InstallCentOS'
+        os_repo = os.getenv('RACKHD_CENTOS_REPO_PATH', \
+            self.__base + '/repo/centos/{0}'.format(version))
+        body = options
+        if body == None:
+            body = {
+                'options': {
+                    'defaults': {
+                        'kvm': 'undefined', 
+                        'version':version,
+                        'repo': os_repo
+                    }
+                }
+            } 
+        self.__post_workflow(graph_name, nodes, body)
+        
+    def install_esxi(self, version, nodes=[], options=None):
+        graph_name = 'Graph.InstallEsx'
+        os_repo = os.getenv('RACKHD_ESXI_REPO_PATH', \
+            self.__base + '/repo/esxi/{0}'.format(version))
+        body = options
+        if body == None:
+            body = {
+                'options': {
+                    'defaults': {
+                        'installDisk': 'firstdisk',
+                        'version': version, 
+                        'repo': os_repo
+                    }
+                }
+            }
+        self.__post_workflow(graph_name, nodes, body)  
+        
+    def install_suse(self, version, nodes=[], options=None):
+        graph_name = 'Graph.InstallSUSE'
+        os_repo = os.getenv('RACKHD_SUSE_REPO_PATH', \
+            self.__base + '/repo/suse/{0}/'.format(version))
+        body = options
+        if body == None:
+            body = {
+                'options': {
+                    'defaults': {
+                        'kvm': 'undefined', 
+                        'version':version,
+                        'repo': os_repo
+                    }
+                }
+            }
+        self.__post_workflow(graph_name, nodes, body)
+        
+    def install_ubuntu(self, version, nodes=[], options=None):
+        graph_name = 'Graph.InstallUbuntu'
+        os_repo = os.getenv('RACKHD_UBUNTU_REPO_PATH', \
+            self.__base + '/repo/ubuntu')
+        body = options
+        if body == None:
+            body = {
+                'options': {
+                    'defaults': {
+                        'kvm': 'undefined', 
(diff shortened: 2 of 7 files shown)


yyscamper:
Invite @pengz1 to review

test/tests/api/v1_1/os_install_tests.py
```diff
@@ -1,27 +1,162 @@
...
+    def install_centos(self, version, nodes=[], options=None):
+        graph_name = 'Graph.InstallCentOS'
+        os_repo = os.getenv('RACKHD_CENTOS_REPO_PATH', \
+            self.__base + '/repo/centos/{0}'.format(version))
+        body = options
+        if body == None:
+            body = {
+                'options': {
+                    'defaults': {
+                        'kvm': 'undefined',
```
yyscamper: `kvm` is a boolean value

test/tests/api/v1_1/os_install_tests.py
```diff
@@ -1,27 +1,162 @@
...
+    def teardown(self):
+        self.__format_drives()  
+    
+    def __get_data(self):
+        return loads(self.__client.last_response.data)
+    
+    def __post_workflow(self, graph_name, nodes, body):
+        workflows().post_workflows(graph_name, timeout_sec=3600, nodes=nodes, data=body)         
+
+    def __format_drives(self):
```
yyscamper: Why need format drive? I think the OS installer will auto do this.

test/tests/api/v1_1/os_install_tests.py
```diff
@@ -1,27 +1,162 @@
...
+        
+    @test(groups=['centos-7-install.v1.1.test'])
+    def test_install_centos_7(self, nodes=[], options=None):
+        """ Testing CentOS 7 Installer Workflow """
+        self.install_centos('7.0')
+
+    @test(groups=['ubuntu-install.v1.1.test'])
+    def test_install_ubuntu(self, nodes=[], options=None):
+        """ Testing Ubuntu 14.04 Installer Workflow """
+        self.install_ubuntu('14.04')
```
yyscamper: usually our testing set the ubuntu version to `trusty`, and `trusty` is also the name that the official chooses.

test/README.md
```diff
@@ -53,8 +53,27 @@ AMQP URL
...
+    export RACKHD_CENTOS_REPO_PATH=[location to http mirror for all CentOS images]
+    export RACKHD_ESXI_REPO_PATH=[location to http mirror for all ESXi images]
+    export RACKHD_UBUNTU_REPO_PATH=[location to http mirror for all Ubuntu images]
+    export RACKHD_SUSE_REPO_PATH=[location to http mirror for all SUSE images]
+
+    Example: export RACKHD_CENTOS_REPO_PATH=http://ip:port/repo/centos
+
+HTTP proxy configuration (optional):
+
+To define HTTP proxies for accessing remote OS image repositories, see the [RackHD Configuration howto](http://rackhd.readthedocs.io/en/latest/rackhd/configuration.html?highlight=httpProxies#rackhd-configuration)
```
yyscamper: The httpProxies doc is only a general description, I suggest giving some concrete example for each OS here or in httpProxies doc, because I realized most people doesn't understand how to set a correct repo address.

test/tests/api/v1_1/os_install_tests.py
anhou: yeah, it's recored here. http://rackhd.readthedocs.io/en/latest/rackhd/index.html#os-installation-workflow-payload 
version	String	required	The version number of target OS that needs to install. NOTE: For Ubuntu, version should be the codename, not numbers, for example, it should be “trusty”, not “14.04”

test/README.md
```diff
@@ -53,8 +53,27 @@ AMQP URL
...
 
-Specify test groups
+OS Installer locations
+
+    Specify specific OS repository path:
+
+    export RACKHD_CENTOS_REPO_PATH=[location to http mirror for all CentOS images]
+    export RACKHD_ESXI_REPO_PATH=[location to http mirror for all ESXi images]
+    export RACKHD_UBUNTU_REPO_PATH=[location to http mirror for all Ubuntu images]
+    export RACKHD_SUSE_REPO_PATH=[location to http mirror for all SUSE images]
```
yyscamper: How about define these parameters to a configure file rather than environment?

test/tests/api/v1_1/os_install_tests.py
```diff
@@ -1,27 +1,162 @@
...
+                    'defaults': {
+                        'kvm': 'undefined', 
+                        'version':version,
+                        'repo': os_repo
+                    }
+                }
+            } 
+        self.__post_workflow(graph_name, nodes, body)
+        
+    def install_esxi(self, version, nodes=[], options=None):
```
yyscamper: Due to the `ssh-validation` task is appended at each OS bootstrap, and some enhancement for ESXi design (see ODR-682). I am sure the bootstrap will always fail if not explicitly set a valid static IP address.

test/tests/api/v1_1/os_install_tests.py
```diff
@@ -1,27 +1,162 @@
...
 from proboscis import test
+from proboscis import after_class
+from proboscis import before_class
 from json import dumps, loads
+import os
 
 LOG = Log(__name__)
 
-@test(groups=['os-install.v1.1.tests'])
+@test(groups=['os-install.v1.1.tests'], \
```
yyscamper: I think you missed the `CoreOS`

yyscamper:
@jlongever : the test case only trigger the install OS workflow, do you have further plan to validate whether each OS parameter are correct set in target OS?

test/tests/api/v1_1/os_install_tests.py
jlongever: Boolean doesn't work, the ks checks for undefined. Setting this to true or false will always evaluate to false allowing the ks to always install kvm. : https://github.com/RackHD/on-http/blob/master/data/templates/centos-ks#L58

test/README.md
jlongever: My preference is to not manage more config files and more documentation. But feel free to open an enhancement issue if you'd like to see these moved to another config file.

test/README.md
jlongever: This readme isnt the official RackHD documentation.  Maybe the readthedocs should be updated with some concrete examples?

test/tests/api/v1_1/os_install_tests.py
jlongever: The story didn't include CoreOS. @amymullins should the story include CoreOS as well?

test/tests/api/v1_1/os_install_tests.py
jlongever: This is required on a fresh deployment of RackHD when the target nodes have already been provisioned with an OS,  otherwise the node will always boot off the disk.  We need to wipe the MBR after running these tests so they run deterministically.

test/tests/api/v1_1/os_install_tests.py
jlongever: That is why I have the esxi tests disabled on line 154 and on line 159 (enabled=False).  I will enable these once the fix has been committed to master.

test/tests/api/v1_1/os_install_tests.py
jlongever: How does this functionally affect the test? 14.04 is trusty,  its semantics.

jlongever:
@yyscamper the scope of this PR is to execute the OS install workflows and validate the graph runs successfully. Adding post OS install tests can be adressed later.

test/tests/api/v1_1/os_install_tests.py
yyscamper: Actually there is a special in install-os job ( see https://github.com/RackHD/on-tasks/blob/master/lib/jobs/install-os.js#L207), so a boolean false should work well.

If not, this is a bug.

test/tests/api/v1_1/os_install_tests.py
yyscamper: So this is due to the pxe boot is not the first boot order? and it cannot support set pxe boot order via ipmi?
If so, I think the HW-simulator can improve this.
Of course, the `format-drive` is reasonable here.

test/tests/api/v1_1/os_install_tests.py
jlongever: It didn't work.

test/tests/api/v1_1/os_install_tests.py
yyscamper: It will not affect the testing, just the user need a special step to rename the `trusty` to `14.04`

test/README.md
yyscamper: Agree

test/tests/api/v1_1/os_install_tests.py
jlongever: Yes tge simulator supports the  ipmi bootdev command, but how does RackHD run that command before the node is discovered and sets the OBM settings? Default boot order is local disk first, then network boot.

test/README.md
yyscamper: There will be more and more such configuration entries as RackHD tests grow up, a single file to manage all these configuration will be more reasonable to me. 
I'm OK to move the test config file as a standalone story to track.

test/tests/api/v1_1/os_install_tests.py
jlongever: The user can set the explicit repo URL using the env variable.

test/tests/api/v1_1/os_install_tests.py
yyscamper: I didn't realize there is a `enabled=False` flag. seems the CentOS/RHEL still deserve the same flag, as it also has been appended the `ssh-validation` task.

test/tests/api/v1_1/os_install_tests.py
iceiilin: How will the format drive workflow be send before the node is discovered?

test/tests/api/v1_1/os_install_tests.py
yyscamper: @jlongever : have you  tested whether 14.04 works? I remembered the ubuntu installer uses `trusty` in many place, so I am afraid the 14.04 even doesn't work well

yyscamper: Image user just want to set a official ubuntu repo (such as http://archive.ubuntu.com/ubuntu/), the repo uses the name `trusty` not `14.04`. So 14.04 will not found the related installing file.
14.04 is hard-coded in the test case, so usually user doesn't change it.

WangWinson: @jlongever  I'm afraid '14.04' as a `version` will not works. Because the `version` param will be render in `ubuntu_preseed`, and it's require the release name 'trusty' instead of '14.04'

test/tests/api/v1_1/workflows_tests.py
```diff
@@ -117,26 +119,54 @@ def test_workflows_library_identifier_get(self):
...
+                    Nodes().nodes_identifier_workflows_active_delete(node)
+                    Nodes().nodes_identifier_workflows_active_get(node)
+                    if self.__client.last_response.status != 204:
+                        retry -= 1
+                        time.sleep(1)
+                    else:
+                        break
+                except Exception, e:
+                    assert_equal(404, e.status, message = 'Expected 404 response status')
+                    break
```
iceiilin: From my test, after nodes_identifier_workflows_active_delete(), no matter it returns 404 or 204, the API follows it will  be retired, even if the actual workflow has been started at the first call:
```
2016-05-19 22:37:26,642 WARNING Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', BadStatusLine("''",))': /api/1.1/nodes/573d9acc1525e6593247fb79/workflows?name=Graph.PXE.Reboot
2016-05-19 22:37:26,642:urllib3.connectionpool:WARNING - Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', BadStatusLine("''",))': /api/1.1/nodes/573d9acc1525e6593247fb79/workflows?name=Graph.PXE.Reboot
```
From
```
curl http://localhost:8080/api/common/workflows/*
```
We can see that two same workflows are started. It might be caused by the untimely release of network resources when it is not a GET operation.

I tested that the API after nodes_identifier_workflows_active_get() won't be retried, so I add this command to shield the retry problem after deleting active workflow.
https://github.com/RackHD/RackHD/pull/232/files#diff-d618aea7eb216b7b27d43c843a01227cR153

You may consider whether it can be adopted.

test/tests/api/v1_1/os_install_tests.py
yyscamper: @jlongever , but the OS-bootstrap will always depend on there is ipmi-obm-service, otherwise it will fail. If ipmi-obm-service is existing, the OS-boostrap will  temporary set set boot-order to pxe regardless the BIOS persistent boot order setting.

jlongever: I'm taking about the node discovery.  If you discover a node and then provision the node with an OS,  then clear mongo,  the next power cycle/reset of the node will boot off the disk.  How does RackHD rediscover it?  It needs to PXE boot right?  How does RackHD do that without knowing anything about the node, there are no more OBM settings because mongo was cleared.  So at the end of each OS installer test run I run a workflow to clear the MBR so I can clear mongo and rerun from scratch again.  Fwiw, there is a proposal/spec requesting to do this exact thing.

test/tests/api/v1_1/os_install_tests.py
jlongever: Well it works setting version to 14.04 because I tested it.  Is that broken?

jlongever: @yyscamper 14.04 is hard coded as the default, the repo option can be overwritten with RACKHD_UBUNTU_REPO_PATH env variable.

test/tests/api/v1_1/os_install_tests.py
yyscamper: Got your point.

test/tests/api/v1_1/os_install_tests.py
yyscamper: @jlongever : If your test passes, then I am OK with this.

jlongever: I'll just change it to trusty to avoid preseed rendering issues.

test/tests/api/v1_1/workflows_tests.py
jlongever: Good feedback.  I'll pull in your change and follow that logic.

test/tests/api/v1_1/os_install_tests.py
amymullins: OS's covered by this story are : CentOS, ESXi, SUSE, and Ubuntu.

yyscamper:
:+1:

tannoa2:
+1

zyoung51:
+1

PR #237, opened 2016-05-23
packer vagrant box using node 4 and based on ubuntu 16.04

anhou:
1. Ubuntu 16.04 is released recently. update packer scripts for ubuntu 16.04. refer to https://github.com/chef/bento/blob/master/ubuntu-16.04-amd64.json
2. upgrade node 0.10 to the latest node 4 for both ubuntu  14.04 and 16.04. 
3. Adjust control network interface from 'eth1' to 'enp0s8'  for Ubuntu 16.04 (default name for virtualbox second NIC).  vagrant don't support systemd biosdevname setting(only support 'eth1'/...), workaround it using config file.

Updated

@RackHD/corecommitters @heckj @benbp @iceiilin

### DIFF
--- a/packer/ansible/roles/node/tasks/main.yml
+++ b/packer/ansible/roles/node/tasks/main.yml
@@ -1,15 +1,10 @@
 ---
-- name: Download n Installer
-  get_url: url=http://git.io/n-install dest=/tmp/n-install mode=0755
-  register: download_n
-  sudo: yes
 
-- name: Install n
-  shell: "yes | N_PREFIX={{ ansible_env.HOME }}/n /tmp/n-install"
-  sudo: yes
-  when: download_n|changed
+- name: Setup nodejs source for installing node v4 
+  shell: curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -
 
-- name: Install node
-  shell: "{{ ansible_env.HOME }}/n/bin/n {{ node_version | default('0.10.37') }}"
+- name: Install node v4
+  apt: pkg={{ item }} state=installed
+  with_items:
+    - nodejs
   sudo: yes
-  when: download_n|changed
--- a/packer/README.md
+++ b/packer/README.md
@@ -17,7 +17,13 @@ enable:
 
 ## To build locally (using code from source)
 
-    packer build -only=virtualbox-iso template.json
+### Build vagrant box based on Ubuntu 14.04
+
+    packer build -only=virtualbox-iso template-ubuntu-14.04.json
+
+### Build vagrant box based on Ubuntu 16.04
+
+    packer build -only=virtualbox-iso template-ubuntu-16.04.json
 
 ## To build locally (using pre-built debian packages)
 
--- a/packer/ansible/rackhd_local.yml
+++ b/packer/ansible/rackhd_local.yml
@@ -8,7 +8,7 @@
     - git
     - rabbitmq
     - mongodb
-    - node-legacy
+    - node
     - isc-dhcp-server
     - monorail
     - snmptool
@@ -21,5 +21,3 @@
   vars:
 # branch of https://github.com/rackhd/rackhd to be used
     - branch: "master"
-# version of Node.js to use
-    - node_version: "0.10.41"
--- a/packer/ansible/roles/isc-dhcp-server/files/systemd-control.cfg
+++ b/packer/ansible/roles/isc-dhcp-server/files/systemd-control.cfg
@@ -1,0 +1,5 @@
+# CONTROL lan
+auto enp0s8
+iface enp0s8 inet static
+address 172.31.128.1
+netmask 255.255.252.0
--- a/packer/ansible/roles/isc-dhcp-server/files/systemd-isc-dhcp-server
+++ b/packer/ansible/roles/isc-dhcp-server/files/systemd-isc-dhcp-server
@@ -1,0 +1,21 @@
+# Defaults for isc-dhcp-server initscript
+# sourced by /etc/init.d/isc-dhcp-server
+# installed at /etc/default/isc-dhcp-server by the maintainer scripts
+
+#
+# This is a POSIX shell fragment
+#
+
+# Path to dhcpd's config file (default: /etc/dhcp/dhcpd.conf).
+#DHCPD_CONF=/etc/dhcp/dhcpd.conf
+
+# Path to dhcpd's PID file (default: /var/run/dhcpd.pid).
+#DHCPD_PID=/var/run/dhcpd.pid
+
+# Additional options to start dhcpd with.
+#	Don't use options -cf or -pf here; use DHCPD_CONF/ DHCPD_PID instead
+#OPTIONS=""
+
+# On what interfaces should the DHCP server (dhcpd) serve DHCP requests?
+#	Separate multiple interfaces with spaces, e.g. "eth0 eth1".
+INTERFACES="enp0s8"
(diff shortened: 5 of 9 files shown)


jlongever:
test this please

packer/ansible/roles/node/tasks/main.yml
```diff
@@ -1,15 +1,21 @@
...
-  register: download_n
-  sudo: yes
 
-- name: Install n
-  shell: "yes | N_PREFIX={{ ansible_env.HOME }}/n /tmp/n-install"
-  sudo: yes
-  when: download_n|changed
+# source ~/.bashrc is for interactive shell
+- name: Install nvm and Appending source string to ~/.bashrc
+  shell: wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.31.0/install.sh | bash
```
benbp: I think we're going to be much better off installing `n` rather than `nvm`, the reason being that `nvm` is a real pain to use when you need to install the same node version for the root and vagrant user, whereas with `n` you just do `sudo n <version>` and it automatically gets installed for all users.

benbp: Additionally you don't have to do any .bashrc updates like you have to do with nvm. It's generally just a much easier tool to use.

anhou: @benbp actually, originally 'node' role is not used, and 'n' tool is also not used , it seems that it was used before then it's disabled or not used finally,do you know the history?

for nvm installation globally, nvm provide some parameters to install it to globally, and someone also have fork 'nvm' to install node and npm globally https://github.com/xtuple/nvm

'nvm' want to reduce the impacts to others and limit the node and npm version to local user, so if want to let others use node, we could install it globally.  .bashrc update is only for interactive shell to run command link 'vagrant ssh -c 'nf start'' 

for why not use' n install' , the problem is that 'n install' only install node, didn't install npm, npm still keep the same with before in system, maybe newer or older, not sure about that,  the newer version will have some problem for 'npm install' . so I choose 'nvm install' to keep consistent with travisCI for 'nvm' ,also for both node and npm, it will help use isolate the problems that caused by version problem, either Jenkins, or TravisCI, or users report. using 'n install is also fine, just need to additional  steps or docs to indicate npm version, and change travisCI to keep consistent with steps here. at least for version for both node and npm.

How do you think?

benbp: @anhou ah I didn't realize the issues with npm updates and n. Yes, that's a bigger problem, so your decision to use nvm makes sense to me, and we can work out the issues with global installs.

benbp:
👍

packer/ansible/roles/node/tasks/main.yml
benbp: Yeah the problems I've seen before, and that seem apparent here, is that the ansible script only installs nvm for the vagrant user? But when you run on-http/on-dhcp-proxy/on-tftp with root as is required, they will not run the nvm installed version of node.

anhou:
@benbp Glad that you also think 'nvm' is our direction. yes, now only 'vagrant' user could use installed 'node' and 'npm'. and have 'sudo' privilege to run installed npm commands such as 'sudo nf start'.  but root could not use(there's no root user in Ubuntu by default) and other users also could not use. 

For installing nvm globally. what I could find the best method  is from https://github.com/xtuple/nvm.  and I double check if it works again after you raise this quetions. I found it also has the problem that when run 'nf start', it could not find 'nf' commands' after 'sudo npm install -g forman', because the environment variables is not updated with related 'npm' and 'node' 's version location PATH. so have to do more steps to refine  https://github.com/xtuple/nvm to let it work including switch npm node modules PATH environment variables when switch node version. and persist this PATH after switching. 

I'm thinking does we need to install node and npm globally for development environment such as vagrant box here,  or OnRack's ova release?  I think user's local installation is enough. if some users want to install nvm globally, they could refine 'nvm' tool by themselves. or do it in their production environment, even they could install node and npm manually or by compiling source codes, they just keep the same version with RackHD's,  we just provide our simplest way to install node and npm to let RackHD work.

Any thoughts?

anhou:
@benbp anyway, I've updated node 4 for ubuntu 14.04 vagrant box.

benbp:
@anhou so these are the difficulties I was referring to and the reason I switched to `n` a while back. Most other node developers don't have this trouble because they don't have to run their node apps as root/with sudo. Regardless, a local installation for the user is not enough because we have to run http/tftp/dhcp with sudo. So you need at bare minimum a root installation of the intended node version.

Thinking about this more, and given potential solutions for using nvm, I'm wondering whether it's simpler to just use `n` and have a manual instruction for installing `npm`. Older versions of `npm` seem to work just fine for installs regardless of the node version, they just download/structure dependencies in a different way across versions. It seems like that involves fewer steps/manual instructions than getting nvm to install node for all the necessary users.

Also, I'm not sure what you mean when you say there is no root user in Ubuntu by default?

anhou:
@benbp for 'no root user', I mean the root user is not enabled by Ubuntu, more accurately, we need to run 'sudo passd root' to add root user's password firstly, then can run commands under root account.

in my steps in this PR, the sudo privilege could work for node/npm and npm installed modules, because sudo privilege are kept when 'sudo'

n is also fine, I just think installed node veriosn by n, should keep node/npm version same with 'nvm install' in TravisCI, or using n in TravisCI? npm older versions works well indeed, I just think if not use nvm, we have to document which npm version works well when users install npm in deployment in different OS or environment. or we specified npm version? if these two things are all addressed, I think using n or nvm is fine. In my practice, nvm method could work in our vagrant box. and I also have a try for using 'n' to install latest node4 version, and install npm version which travisCI used, a little complex for me, and still not succeed . 
so I still think for vagrant box deployment, we just keep node/npm consistent with travisci, and make sure it works for 'vagrant' user. and Jenkins based on vagrant box also works well with the same node/npm version. it's just a box for development. so I prefer to defer multiuser support to users, users use either n or nvm, or they refine nvm tool to support multi-users. they decide in their deployment.

davidzuhn:
There's no need to give the root user a password and the ability to log in.   Provided that the user the RackHD software is running as is enabled to run via sudo, that will be completely sufficient to run as root from a lower privilege.   That's what we do on the OnRack appliance.

benbp:
@anhou I didn't think the PATH not resetting with sudo works out of the box? For me, after going through the nvm install instructions, it does not work, e.g.

```
vagrant@vagrant-ubuntu-trusty-64:~$ nvm use 4.4
Now using node v4.4.5 (npm v2.15.5)
vagrant@vagrant-ubuntu-trusty-64:~$ node -v
v4.4.5
vagrant@vagrant-ubuntu-trusty-64:~$ sudo node -v
v0.10.25
```

@davidzuhn agreed. Our demo system has global sudo for the vagrant user as well.

anhou:
@benbp   the handling to reserve the PATH in this PR is
```
+- name : Keep Environment for vagrant
 +  shell : /bin/bash -lic 'echo "Defaults exempt_group=vagrant" >> /etc/sudoers.d/vagrant'
```

anhou:
@benbp I've updated this PR with https://github.com/nodesource/distributions method. it could be used by mulitple user/sudo  and also could keep consistent with travisci's node/npm version. previously I misunderstand that https://github.com/nodesource/distributions only install nodejs, not install bundled npm like ubuntu distributions.but it could. so this is a better way.

benbp:
👍

yyscamper:
:+1:

PR #245, opened 2016-05-27
Add default OBM service options to os-installer

jlongever:
@RackHD/corecommitters @keedya @uppalk1 @tannoa2 @zyoung51 

Updated to reflect changes that required explicit OBM options input.


keedya:
:+1:

PR #257, opened 2016-06-06
OS installer timeout option and reformat default AMQP output

jlongever:
- Set explicit timeout option per OS install test
- Set explicit installDisk
- Cleanup log output in default AMQP handler (format json)


anhou:
:+1:

yyscamper:
:+1:

PR #267, opened 2016-06-10
Add and setup PostgreSQL backend services

jlongever:
@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2 @keedya

### DIFF
--- a/packer/ansible/rackhd_local.yml
+++ b/packer/ansible/rackhd_local.yml
@@ -18,6 +18,7 @@
     - images
     - swagger
     - integrationtest
+    - postgresql
   vars:
 # branch of https://github.com/rackhd/rackhd to be used
     - branch: "master"
--- a/packer/ansible/rackhd_package.yml
+++ b/packer/ansible/rackhd_package.yml
@@ -16,3 +16,4 @@
     - ipmitool
     - images
     - integrationtest
+    - postgresql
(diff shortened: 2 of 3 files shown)


heckj:
did you want to always add postgres to the image, or choose between postgres and mongo?

jlongever:
Always - at least temporarily since the current workflow related stores are tied to mongo so the initial postgres PR will exclude them.  That support will come later after some additional refactoring around the workflow related store.  

@benbp @zyoung51 correct me if I'm wrong here.

jlongever:
@heckj fyi,  @zyoung51 added backend DB override options to the config:
https://github.com/RackHD/on-build-config/pull/9/files#diff-c6eccdca1e61b2dd0a6da07d0b0b8d67R65

So there could actually be mixed databases supported here,  at least in the short term.

heckj:
ah, cool - missed that :+1: 

Could we also get documentation added to the docs describing the databaseOvereide config structure and how to use it please?

jlongever:
@zyoung51 is still working on getting the PR up (I'm a little ahead of him ;)) but the docs will definitely get updated once its in.

heckj:
ah! :+1: code/PR race is on, eh?

jlongever:
In @zyoung51's defense his PR is a little more involved.

yyscamper:
:+1:

PR #271, opened 2016-06-14
fix the bug: when a node with enclosure has no manufacturer, t case fails.

changev:
when a none enclosure node has no manufacturer info the tag based on its id will be tagged to not only the none manufacturer nodes but also the enclosure node, in this case the enclosure node didn't be used to tag but to test tags.
@RackHD/corecommitters


benbp:
👍

yyscamper:
:+1:

PR #272, opened 2016-06-15
Functional test fixes

zyoung51:
- Modify tags and skus to request the 'dmi' catalog instead of attempting
  to parse the catalog array.  This resolves an issue where the 'dmi' was
  not the first element in the array, and the tests would fail.
- Modify the 1.1 Delete:/nodes/:mac/dhcp/whitelist to only delete compute
  type identifiers to align with the POST:/nodes/:mac/dhcp/whitelist test.

@RackHD/corecommitters @uppalk1 @keedya @tannoa2 @jlongever


keedya:
:+1:

PR #274, opened 2016-06-15
added the windows files

tannoa2:


anhou:
:+1:

example/samples/install_windows_payload_full.json
```diff
@@ -0,0 +1,28 @@
+{
+  "options": {
+    "defaults": {
+      "hostname": "localhost",
+      "domain": "rackhd",
+      "password": "onrack1",
```
anhou: password could 'RackHDRocks!' be consistent with other OS installation payloads. it's also the default . it will help when someone install os in node, and leave there, and another one want to login have a try to check something, at least we found  it's very useful in Shanghai side in practice. not a big deal, just for convenience :)

tannoa2:
@anhou  all set

tannoa2:
test this please

keedya:
:+1:

PR #279, opened 2016-06-22
Enable auth and create first user in packer build

jlongever:
- Add script to create first user for 2.0 and redfish API authentication.  Uses the localhost exception method to generate the first user: https://docs.mongodb.com/manual/core/security-users/#localhost-exception 

@RackHD/corecommitters @zyoung51 @tannoa2 @uppalk1 @keedya


yyscamper:
:+1:

PR #319, opened 2016-07-14
Add NFS NIC device last, so that eth1 in /etc/default/isc-dhcp-server…

benbp:
… points to the right interface

Hat tip to @DavidjohnBlodgett for finding this issue.

Basically, if you uncomment the private network settings for using NFS mounts, in its current place in the Vagrantfile, it will add a network interface to be enumerated _before_ the closednet interface that the RackHD/DHCP services run on. This conflicts with the entry in `/etc/default/isc-dhcp-server` that points to `eth1`, such that the isc-dhcp-server upstart service will fail on startup. If we move the extra private network interface to be after all other network setup commands in the Vagrantfile, then it gets enumerated as `eth2` and the closednet interface stays on `eth1`

@RackHD/corecommitters @zhangy42 @lacarb

### DIFF
--- a/example/Vagrantfile
+++ b/example/Vagrantfile
@@ -15,18 +15,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             v.customize ["modifyvm", :id, "--nicpromisc2", "allow-all", "--ioapic", "on"]
         end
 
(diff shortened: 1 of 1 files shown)


DavidjohnBlodgett:
👍  tested.

benbp:
testing something, ignore

👍

yyscamper:
:+1:

PR #328, opened 2016-07-19
update example/ to use InfraSim

heckj:
- virtual SKU definitions added
- example CentOS install workflow for SKU
- updated README with InfraSim how to details

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -17,21 +17,13 @@ and [Ansible](http://docs.ansible.com/ansible/intro_installation.html)
 installed onto your system in order to use this script.
 
 **NOTE**: Do not use Vagrant 1.8.0, as the private network settings don't appear
-to be working correctly. Bug filed upstream with Vagrant at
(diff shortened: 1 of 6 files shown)


heckj:
@RackHD/corecommitters @fub2 <-- not exactly perfect in the README (it's more confusing than ever now with it's growing detail), but functionally I've vetted it works. There's some definite quirks in using the InfraSim with Vagrant, some of which comes from Vagrant's dual-nic setup and the bridging constructs, and some of which is just degraded performance due to lack of VTx support in Virtualbox. It is, however, the only path I see to allowing an end-user to get to be able run an integration test that does something meaningful and vet the whole system as operational to at least some level.

Opened related issues in InfraSim (https://github.com/InfraSIM/InfraSIM/issues/2, https://github.com/InfraSIM/InfraSIM/issues/3), while closing the earliest https://github.com/InfraSIM/InfraSIM/issues/1 as now resolved since I could install CentOS with what's posted.

anhou:
:+1:

Issue #338, opened 2016-07-27
Make control network NIC name/IP configurable in /opt/monorail/config.json


panpan0000:
so far, we always use eth1 with 172.31.128.1
but in Ubuntu 16.04 and some newer OS , the naming convention is totally the different(ensxx).
and sometimes, IP may not always be  172.31.128.1

I tried to modify the /opt/monorail/config.json

but seems it doesn't work if simply change above config file and restart services...

panpan0000:
@anhou

leoyjchang:
Story created to track this work at https://www.pivotaltracker.com/story/show/127327783

yyscamper:
@panpan0000 : What's the config key? The `apiServerAddress`? Now only the `address` in `httpEndpoints`  works.

panpan0000:
@yyscamper , my requirement is that : 
RackHD can run on control NIC with any NIC name besides eth1, run on any IP besides  172.31.128.1.
and the control port NIC name and IP are configurable in  /opt/monorail/config.json

PR #339, opened 2016-07-27
Add a simple quick install RackHD script, which as a quick try for users (only for Ubuntu trusty for now)

panpan0000:
## **Google Group Discussion**: 
https://groups.google.com/forum/#!topic/rackhd/7W425hRyXdU

## **Motivation**: 
 RackHD installation is not quite easy and fast. a ONE-KEY-Install is a good ask.
 people can directly install by   `wget -qO- https://raw.githubusercontent.com/panpan0000/RackHD/one-key-quick-install/example/install_rackhd.sh  | sudo bash`
## **Solution**:
 just combine the steps(install from package) in readthedocs, and make it a single script.
http://rackhd.readthedocs.io/en/latest/rackhd2.0/ubuntu_package_installation.html?highlight=install

NOTE: it's only tested in Ubuntu 14.04. and definitely,  it can't apply to other Linux distribution or even Ubuntu 16.04.later on , we will do OS switch-case in this script to make it workable in other OS.

## **Problem**:
the eth1 and IP 172.31.128.1 seems can't be easily modified. so I force check it at script beginning.
if I'm wrong, then **my script can read out the user's NIC name and IP , and modify the monorail/config.json before services starting.** 
This confusion is relevant to Issue https://github.com/RackHD/RackHD/issues/338

### DIFF
--- a/example/install_rackhd.sh
+++ b/example/install_rackhd.sh
@@ -31,8 +31,23 @@ check_NIC(){
     NIC_IP=$( ifconfig $NIC_Name | grep "inet addr" | awk -F: '{print $2}' | awk '{print $1}' )
 
     if [[ ! "$NIC_IP" == '172.31.128.1' ]]; then
-        echo "[Error] default RackHD configure files takes eth1 with IP 172.31.128.1 as control NIC port.RackHD may not function well if you left all config as default !"
-        exit -4
+	echo -n "Do you want this script to force set your $NIC_Name IP to 172.31.128.1 ? (type \"yes\" to force set, others to abort) > "
+	read willing_to_force_ip
+	if  [[ "$willing_to_force_ip" == 'yes' ]]; then
+		sudo echo "\
+		auto $NIC_Name
+		iface eth1 inet static
+		address 172.31.128.1
+		netmask 255.255.252.0" >> /etc/network/interfaces
+		echo "[INFO] will restart your $NIC_Name..."
+		ifdown $NIC_Name
+		ifup $NIC_Name
+		New_IP= $( ifconfig $NIC_Name | grep "inet addr" | awk -F: '{print $2}' | awk '{print $1}' )
+		echo "Your $NIC_Name new IP address is $New_IP"
+	else
+	        echo "[Error] default RackHD configure files takes eth1 with IP 172.31.128.1 as control NIC port.RackHD may not function well if you left all config as default !"
+	        exit -4
+	fi
     fi
 }
 
@@ -49,7 +64,8 @@ check_install_nodejs(){
         NODEJS_VER=$(nodejs --version | sed s/v//)
         if [[ "$NODEJS_VER" < "4.0" ]] ; then
             # Remove Old NodeJS
-            #sudo apt-get remove nodejs nodejs-legacy
+            echo "[INFO] removing old version [$NODEJS_VER] of NodeJS..."
+            sudo apt-get -y remove nodeijs nodejs-legacy
             do_install=true;
         else
             do_install=false;
--- a/example/install_rackhd.sh
+++ b/example/install_rackhd.sh
@@ -1,0 +1,236 @@
+#!/bin/bash
+#
+########################################################################################################
+#
+#  this script is based on RackHD document "Ubuntu Package Based Installation"
+#
+#  http://rackhd.readthedocs.io/en/latest/rackhd2.0/ubuntu_package_installation.html?highlight=install
+#
+#
+#########################################################################################################
+
+
+#############################################
+#  Check you have 2 NIC
+#  eth0 for the public network - providing access to RackHD APIs, and providing routed (layer3) access to out of band network for machines under management
+# eth1 for dhcp/pxe to boot/configure the machines
+#
+#
+# Parameter #1 ($1) is the NIC name of the control port for RacKHD
+#
+#############################################
+
+check_NIC(){
+    NIC_Name=$1
+    echo $NIC_Name
+
+    if [[ "$NIC_Name" -ne "eth1" ]]; then
+        echo "[WARNING] RackHD takes the control NIC port as eth1 by default , with IP 172.31.128.0/22. RackHD may not function well if you left all config as default ! "
+        sleep 2
+    fi
+    NIC_IP=$( ifconfig $NIC_Name | grep "inet addr" | awk -F: '{print $2}' | awk '{print $1}' )
+
+    if [[ ! "$NIC_IP" == '172.31.128.1' ]]; then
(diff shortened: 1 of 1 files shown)


panpan0000:
@heckj @iceiilin @yyscamper @anhou

heckj:
Do we want to maintain this in addition to the pattern we have of installing RackHD with ansible? (the two ansible playbooks - one to install from packages, and the other to install from source) at https://github.com/RackHD/RackHD/blob/master/packer/ansible/rackhd_local.yml and https://github.com/RackHD/RackHD/blob/master/packer/ansible/rackhd_package.yml respectively?

example/install_rackhd.sh
```diff
@@ -0,0 +1,236 @@
+#!/bin/bash
+#
+########################################################################################################
+#
+#  this script is based on RackHD document "Ubuntu Package Based Installation"
+#
+#  http://rackhd.readthedocs.io/en/latest/rackhd2.0/ubuntu_package_installation.html?highlight=install
```
amymullins: @panpan0000 we are moving away from separate 1.1 / 2.0 user's guide, the link will no longer valid.  Check out: https://github.com/RackHD/docs/pull/312#issuecomment-235677275 and the latest comment "Just added a commit that deletes the RackHD 2.0 user's guide and adds 2.0 API commands to the existing User's Guide when there's differences"

panpan0000:
@heckj 
you are right, I just realize this PR achieves the very close purpose as ansible scripts rackhd_package.yml. I might should have gone the path of enhancing it .. How I could forget it ...

but some worries:
(1)  ansible is not quite friendly to develop/debug..(maybe I'm quite a newbie of it)
(2)  you missed remove legacy nodeJS ( < 4.x ) step in ansible roles.. I feel a bit paint to achieve it via ansible... but easy with my scripts :-P
(3) I have an user-interaction-step , to gain the nic name which user want to assign as RackHD control port. because "eth1" naming convention doesn't always apply other Linux distribution.
(4)  I have a step in my script , with user agreement,  can force set eth1(or the nic which user input as RackHD control port) IP to 172.31.128.1.  Do you think it necessary ? (otherwise, RackHD doesn't work)
(5) I've not digger if ansible can achieve (3)(4) this kind of user interaction via stdin.
(6) at the end of install scripts, I've an idea, that dump some basic API to user, to tell them how to validate if RackHD API works.


anyway,  **we need to add in readthedoc , quick-start page and install page, to tell people to use this "one-line-installation ansible method"**   this is VERY important, to  new trier :-) 

I will close my PR, if you think the ideas is worthy and we find above ideas can easily achieve by ansible

zyoung51:
@heckj @panpan0000 This type of script is a +1 for me provided it can be maintained.  Going through the required steps of installing the proper version of ansible, and even the necessity of ansible, should not be a barrier to entry on the installation of RackHD.

example/install_rackhd.sh
```diff
@@ -0,0 +1,252 @@
+#!/bin/bash
+#
```
zyoung51: Missing set -e and error handling throughout the script.  Can you add the checks to validate that the commands completed as you expected them to?  ie: check that the apt-get install worked, the wgets worked, etc.

heckj:
@panpan0000 you make good points. I'm interested in making sure we have a clear, documented, and maintained path to do this sort of operation - not that we specifically use ansible over a shell script. I just sent an email more broadly to the community asking for feedback. If we want to go to a shell script based mechanism, I'm good with that, but would like to be explicit about it - and make sure that we integrate plans to maintain it, and deprecate any other paths. 

I think as a project we need to keep a clear mechanism for doing things like "quick install and setup" - ansible may be significant overkill for that. I'm looking forward to feedback from our community on that topic, so please leave this PR open and let's consider where we want to go as a community with this.

ytjohn:
@panpan0000 I think it would be better to keep expanding ansible playbook. If we want a script - make a script that installs ansible, downloads the playbook, and runs that. 

I have started the task of improving the ansible playbook to break it away from hard-coded variables. Take a look at #347 for the first cycle of that. The advantage here is that you can run the playbook, or parts of it (tags) over and over again. I'll also be replacing hard coded configs (ie: `wget https://raw.githubusercontent.com/RackHD/RackHD/master/packer/ansible/roles/monorail/files/config.json  -O /opt/monorail/config.json`) with templates.

benbp:
I agree with @zyoung51 that it is kind of heavy-handed to require ansible for setup (I can't think of any open source project off the top of my head that doesn't have a shell script installer). That said, our installation isn't as simple as dropping a single binary and a couple dependencies into $PATH. We have heavy dependencies and a complex setup, which I think could turn into a hellishly complex shell script over time as people with different environments find bugs with the installer. So I'm not sure what the best approach is, just some of my thoughts on the matter above.

ytjohn:
A good example for comparison in this space is [Foreman](https://theforeman.org/), which does its install using puppet underneath, but provides a tool called [foreman-installer](https://theforeman.org/manuals/1.12/index.html#3.2.2InstallerOptions) which  can run in interactive mode or you can set all the options on the command line. foreman-installer will then create a variables file, install puppet, and the use that to install the application.

I think it would be good to build a robust, flexible installer that uses an established config management toolset (puppet, salt, chef, ansible), and then a simple to use "one command" installer tool, which could easily be a batch script or perhaps a higher level language (foreman-installer is done in ruby and uses modules to handle all the potential options). Ansible makes good sense as an underlying manager, as the rackhd product requires ansible for most of the post-install workflows.

davidzuhn:
If you go down the road of a script, please make it very clear what's being done in it, so that those who do not trust 'curl http://some.random.url/ | sudo bash' can readily see what actions are being done.   Many, many people do not trust the Internet with sudo.

example/install_rackhd.sh
panpan0000: fixed. thx.

example/install_rackhd.sh
panpan0000: fixed. thanks

heckj:
For a future view of what a "simple shell script" could become: https://git.openstack.org/cgit/openstack-dev/devstack :-)

👍 for merging once you add a `set -e` near the top of the script to force immediate failure on shell script internal error results on any command

yyscamper:
If I have to choose one from bash script and ansible playbook, I am in favor of bash script bases on the learning cost of ansible and the complex of ansible playbook. The bash script is very simple and straightforward.

yyscamper:
:+1:

panpan0000:
update per comments.   looking forward #338 can be addressed.

panpan0000:
test this please

changev:
test this please

PengTian0:
test this please

PengTian0:
test this please

heckj:
👍

PR #344, opened 2016-07-28
adding sudo for package install

heckj:
Thanks Jason (and sorry for the pain of missing this)

PR #346, opened 2016-07-28
Install libkrb5-dev package as sudo

jlongever:
@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @zyoung51


heckj:
how about a +1 on #344 instead - already made the PR

jlongever:
already done here https://github.com/RackHD/RackHD/pull/344

PR #347, opened 2016-07-28
Adding Vagrant in tests

ytjohn:
This begins the process of improving the ansible playbook and providing a vagrant instance that installs from packages using that playbook.

 - Create Vagrantfile under test that will use ansible playbook upon `vagrant up`
 - Add default variables to packer/ansible playbook
 - Variable enhancements to node, monorail, and rackhd-packages roles
 - Create monorail/config.json from ansible variables.
 - Create inventory/group_vars for vagrant environment

Once the initial changes are implemented, we can work on improving more roles and adding tests around the vagrant instance.

PR #370, opened 2016-08-08
Add FIT tests to rackhd test

gpaulos:
This PR is the FIT (Functional Integration Test) test suite that is used by OnRack (Corsair) to test RackHD/OnRack via Jenkins.

FIT tests are located in the RackHD/test/fit_tests directory. They will not interfere with existing test jobs running from RackHD/test directory.

The goal is to integrate all RackHD/OnRack testing under a single universal and unified framework. This is the first step towards achieveing that goal.

Currently FIT contains 185 tests total, of which 164 are 'smoke' tests.

The framework has been tested and operated under both real hardware stacks and Vagrant simulated stacks. It is adaptable to many diverse environments via configuration files in RackHD/test/fit_tests/config/.  Refer to README files for operation and configuration.

Result from a Jenkins FIT test run is available here (EMC internal):
https://roebling.hwimo.lab.emc.com/jenkins/job/Running-Tests/job/RackHD-FIT-Smoke-Test/65/

The test failures are related to the following bugs:
ODR-801 - SKU pack failures
ODR-803 - management server fails on redfish
ODR-812 - api/2.0/skus fail post/put
ODR-796 - chassis/enclosure issues
ODR-795 - chassis/enclosure issues
ODR-793 - chassis/enclosure issues

Short-term goal is to get a clean run of FIT smoke test by fixing product bugs and any other test bugs that may crop up.

Medium-term goal is to run FIT along with legacy test suite as the release test gate.

Long-term goal is to unify all testing under a single suite.

- George Paulos


test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+- 'deploy' contains deployment and installation scripts
+- 'templates' contains script templates for making new tests
+- 'util' contains non-test utilities
+
+## Configuration
+
+Local runtime parameters are set from the 'config/global_config.json' file.
+Stack definitions are set from the 'config/stack_config.json' file.
+An alternate configuration directory can be selected using the -config argument.
+More details in config/README.mg.
```
heckj: I think this should be `config/README.md` instead of `config/README.mg`

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+      -stack STACK      stack label (test bed), overrides -ora
+      -ora ORA          OnRack/RackHD appliance IP address or hostname, default:
+                        localhost
+      -version VERSION  OnRack version, example:onrack-release-0.3.0, default:
+                        onrack-devel
+      -xunit            generates xUnit XML report files
+      -list             generates test list only
+      -sku SKU          node SKU, example:Phoenix, default=all
+      -obmmac OBMMAC    node OBM MAC address, example:00:1e:67:b1:d5:64,
+                        default=all
```
heckj: I can't quite tell what this option is doing for me. Why would I invoke it, and why does it default to `all`?

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+      -nodeid NODEID    node identifier string of a discovered node, example:
```
heckj: I can't quite tell what this option is doing for me. Why would I invoke it? How would I use it?

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+                        56ddcf9a8eff16614e79ec74
+      -http             force using http protocol
```
heckj: this may be better represented as `forces the tests to utilize the http protocol` rather than 'force using http protocol'

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+                        [-list] [-sku SKU] [-obmmac OBMMAC | -nodeid NODEID]
+                        [-http | -https] [-v V]
+
+    Command Help
+
+    optional arguments:
+      -h, --help        show this help message and exit
+      -test TEST        test to execute, default: tests/
+      -config CONFIG    config file location, default: fit_tests/config
+      -group GROUP      test group to execute: 'smoke', 'regression', 'extended',
```
heckj: how are groups defined here, and does these overlay at all with the dependency groups in the existing tests?

test/fit_tests/config/stack11.json
```diff
@@ -0,0 +1,23 @@
+{
+  "11": {
```
heckj: doesn't this duplicate information in `stack_config.json`?

heckj:
Thanks @gpaulos - can you provide some additional detail on how a remote developer could invoke these tests to work with a vagrant-based local development setup - RackHD in a vagrant instance, and a single target machine in a vagrant instance, per the demonstration/example setup at https://github.com/RackHD/RackHD/tree/master/example? Is that something you could add to the README perhaps?

This request overlays with https://github.com/RackHD/RackHD/pull/347, where @ytjohn is trying to set up an environment to deploy and test RackHD within a constrained environment to vet functionality.

If someone wanted to leverage the `-stack` option or `-ora` option locally, what do they need to set up, how does it need to interact, and where is that documented? @michaelsteven has a build/test environment remote to our EMC internal setup where I think this would make sense - I think he's using Bamboo locally - so where we can get detail for him on how to set up and run these tests?

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+      -http             force using http protocol
+      -https            force using https protocol
+      -v V              Verbosity level of console output, default=0, Built Ins:
+                        0: No debug, 2: User script output, 4: rest calls and
+                        status info, 6: other common calls (ipmi, ssh), 9: all the
+                        rest
+
+This example will run the RackHD installer onto stack 1 via the wrapper script:
+
+    python run_tests.py -stack 1 -test deploy/run_rackhd_installer.py
```
heckj: In an example where the tests & virtualenv for them are running on my laptop with Vagrant, I have RackHD responding at localhost:9090 and authenticated w/ HTTPS at localhost:9093 - how can I specify those endpoints in either configuration or options to the fit tests to run against a local instance?

i.e. this local testing scenario:

```
+----------------------------------------------------------+
|                                                          | 
|  .venv/FIT TESTS -------> [virtualbox: RackHD]           |
|                   \          |                           |
|                    \         |                           |
|                     +---> [virtualbox: vQuanta D51]      |
|                                                          | 
+----------------------------------------------------------+
```

ytjohn: It looks like you would have to modify that in test/fit_tests/config/global_config.json.  

A local_config.json override file (not checked into git) might be a good future enhancement.

ytjohn:
After reviewing this, I would say it shouldn't conflict with #347. #347 is entirely about using ansible to install rackhd into a vagrant instance. I was then going to wrap tests to execute inside the vagrant instance (instead of external to it). 

I would rather see these tests integrated, and then make any necessary changes in my PR. I have some local work to run the smoke tests. I could expand that to run these fit tests.

test/fit_tests/README.md
gpaulos: This is to specify a specific node to test where you don't yet know the node ID, it may not have been discovered yet.

test/fit_tests/README.md
gpaulos: Groups are just tags that are set in the test script with an '@attr' decorator.

Example:
@attr(all=True, regression=True)
class os_bootstrap_base(fit_common.unittest.TestCase):

Nosetests does a search on the attr tag and runs any that match.

test/fit_tests/config/stack11.json
gpaulos: This is a override file. Any keys in an override file replace entries in the stack_config file. This is described in the config/README.md.

test/fit_tests/README.md
gpaulos: An alternate config file directory can be specified with the -config parameter. See README.

test/fit_tests/README.md
gpaulos: Fixed in latest commit.

test/fit_tests/README.md
gpaulos: Fixed in latest commit.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+    if VERBOSITY >= 4:
+        print "restful: Action = ", rest_action, ", URL = ", url_command
+
+    # prepare payload for XML output
+    payload_print = []
+    try:
+        json.dumps(rest_payload)
+    except ValueError:
+        payload_print = []
+    else:
```
nortonluo: If the play load is binary-put or binary post, the json dump will get exceptions. Suggest to check if it is  binary-put or binary post, it will skip the json.dump

gpaulos: That is why it has a try/except/else check. This works.

test/fit_tests/README.md
gpaulos: In the instance of http://localhost:9090, I suggest that you specify 9090 as the http port in the configuration file, then use 172.31.128.1 as the ora address and for http. This will bypass the default. Same goes for https.

heckj: I'm not entirely clear on how to set up the configuration file to work - here's what I'm trying:

`config/vagrant_local.json`:
```
{
  "vagrant": {
    "ora": "172.31.128.1",
    "ports": {
      "http": 9090,
      "https": 9093
    }
  }
}
```

terminal from `rackhd/example`:
  - `vagrant up dev && vagrant ssh dev -c "sudo nf start"`
another window in `rackhd/example`
  - `vagrant up quanta_d51`

Not sure how to coordinate this with invoking the tests:
`racked/test/fit_tests`:
```
python run_tests.py -config vagrant_local.json
**** Using config file path: vagrant_local.json
**** Global Config file: /Users/heckj/src/rackhd/test/fit_tests/config/vagrant_local.json/global_config.json missing or corrupted! Exiting....
```

I tried several variations - how should I be using the `-config` option to make this work, and is the config file constructed correctly? There wasn't any detail/documentation on how to do that, or what was accepted - just the examples, so I guessed at it.

test/fit_tests/common/fit_common.py
nortonluo: I get the error if use this try exception check. Seems this is not a kind of value error thus the exception could not handle it.
======================================================================
ERROR: test_api_11_update_quanta_d51_2u_skupack (test_rackhd11_api_skupack.test_rackhd11_api_skupack)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/luon1/newfit/RackHD/test/fit_tests/tests/rackhd11/test_rackhd11_api_skupack.py", line 244, in test_api_11_update_quanta_d51_2u_skupack
    self.rackhd_api_11_update_skupack("Quanta D51 2U")
  File "/home/luon1/newfit/RackHD/test/fit_tests/tests/rackhd11/test_rackhd11_api_skupack.py", line 97, in rackhd_api_11_update_skupack
    api_response = fit_common.rackhdapi('/api/1.1/skus/'+ item['id']+'/pack', action='binary-put',payload=f)
  File "/home/luon1/newfit/RackHD/test/fit_tests/common/fit_common.py", line 254, in rackhdapi
    rest_action=action, rest_payload=payload, rest_timeout=timeout, rest_headers=rest_headers)
  File "/home/luon1/newfit/RackHD/test/fit_tests/common/fit_common.py", line 282, in restful
    json.dumps(rest_payload)
  File "/usr/lib/python2.7/json/__init__.py", line 243, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: <open file u'quanta-d51-2u_d51.tar.gz', mode 'rb' at 0x7fdcc77e0a50> is not JSON serializable

gpaulos: Need more info Norton. It works for me. Is it failing the test or is it just giving traceback errors? What is the command line you are running? What environment?

gpaulos: Try this Norton:

Edit the fit_common file and take out the valueError string:

except:  
     payload_print = []  

If that works, I'll check it in.

test/fit_tests/README.md
gpaulos: There is a README in the config dir but I think it needs to be clearer.
The -config parameter specified a directory not a file.
The directory must contain at least a global_config.json file. Stack config is optional.
Your example is combining the two. 
I will add clearer verbiage in the README.

In the mean time just edit the existing global config with your port numbers
Then run with the -ora parameter:
python run_tests.py -ora 172.31.128.1 -test tests -group smoke

heckj:
Thanks for the updates in the README, but I'm still not clear on how to get this all working correctly in an entirely local environment. Can you please add an explicit, step-by-step example that can be used with the vagrant example that is included in this repository to vet this testing mechanism by a local developer?

Here's what I'm doing now:

## basic setup
```
cd ~/src
git clone https://github.com/rackhd/rackhd
# to get this specific PR
git remote add gpaulos https://github.com/gpaulos/rackhd
git fetch --all --prune
git checkout add-fit_tests-to-rackhd-test
# force update to any code from upstream that you've pushed in this PR
git reset --hard gpaulos/add-fit_tests-to-rackhd-test 
cd rackhd/test/fit_test
virtualenv .venv
source .venv/bin/activate
pip install -r requirements.txt
```

Modified `config/global_config.json` to the local ports per your suggestion:
```
diff --git a/test/fit_tests/config/global_config.json b/test/fit_tests/config/global_config.json
index 8717f98..64e5dc4 100755
--- a/test/fit_tests/config/global_config.json
+++ b/test/fit_tests/config/global_config.json
@@ -73,7 +73,7 @@
   },
   "ports": {
     "_comment": "These are the northbound rest API port assignments",
-    "http": 8080,
-    "https": 8443
+    "http": 9090,
+    "https": 9093
   }
 }
```

## window 1 - RackHD
- new terminal window
```
cd ~/src/rackhd/example
vagrant up dev && vagrant ssh dev -c "sudo nf start"
```

RackHD is up and running at the end of this process in a single Vagrant VM


## window 2 - tests

- new terminal window
```
cd ~/src/rackhd/test/fit_tests
source .venv/bin/activate
python run_tests.py -ora 172.31.128.1 -test tests -group smoke
```

The tests start running, but only with errors as their output:
 - `EEEEEE`

Do I need to have another window open and invoke the a virtual quanta_d51 into discovery as I start the tests? What coordination with any existing systems are needed in order to see this operate?

test/fit_tests/common/fit_common.py
nortonluo: Yes, that works now in my system.

test/fit_tests/README.md
yyscamper: So what's meaning if set the obmmac to `all`? 
Meanwhile, obm only uses BMC mac, but there are still other MACs (in node.indentifiers), whether those MACs will still be scanned?

test/fit_tests/README.md
yyscamper: It there any command line to list all groups?

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+      -version VERSION  OnRack version, example:onrack-release-0.3.0, default:
+                        onrack-devel
+      -xunit            generates xUnit XML report files
+      -list             generates test list only
+      -sku SKU          node SKU, example:Phoenix, default=all
+      -obmmac OBMMAC    node OBM MAC address, example:00:1e:67:b1:d5:64,
+                        default=all
+      -nodeid NODEID    node identifier string of a discovered node, example:
+                        56ddcf9a8eff16614e79ec74
+      -http             fforces the tests to utilize the http API protocol
```
yyscamper: typo: "fforces" --> "forces"

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+      -https            forces the tests to utilize the https API protocol
```
yyscamper: Need username/password for https API authentication, Is there any command to specify them?

test/fit_tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+    Command Help
+
+    optional arguments:
+      -h, --help        show this help message and exit
+      -test TEST        test to execute, default: tests/
+      -config CONFIG    config file location, default: fit_tests/config
+      -group GROUP      test group to execute: 'smoke', 'regression', 'extended',
+                        default: 'all'
+      -stack STACK      stack label (test bed), overrides -ora
```
yyscamper: I wonder whether the word `stack` is a general term that developer usually knows what it means. If not, I'd suggest adding some explanation for it.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+    "sku": os.getenv("SKU", "all"), # node SKU name
+    "obmmac": os.getenv("OBMMAC", "all"), # node OBM MAC address
+    "nodeid": os.getenv("NODEID", "None"), # node ID
+    "hyper": "None", # hypervisor address
+    "version": os.getenv("VERSION", "onrack-devel"), # code version
+    "xunit": os.getenv("XUNIT", False), # XUNIT output
+    "list": os.getenv("LIST", False), # list tests
+    "group": os.getenv("GROUP", "all"), # test group
+    "http": os.getenv("HTTP", "False"), # force http api protocol
+    "https": os.getenv("HTTPS", "False"), # force https api protocol
```
yyscamper: I don't think using the environment to store the argument is a good design, especially in consideration of using some common words, such as "VERSION", "CONFIG", "LIST", "HTTP", .etc.

I'd suggest creating a file to store the default value, something like the .vimrc for vim.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+    if VERBOSITY >= 9:
+        print "remote_shell: STDOUT =\n"
+        logfile_redirect = sys.stdout
+
+    if expect_receive == "" or expect_send == "":
+        (command_output, exitstatus) = \
+            pexpect.run("ssh -q -o StrictHostKeyChecking=no -t " + user + "@"
+                        + address + " sudo bash -c \\\"" + shell_cmd + "\\\"",
+                        withexitstatus=1,
+                        events={"assword": password + "\n"},
```
yyscamper: intentionally use `assword`? not `password`?

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+    subprocess.call(["touch ~/.ssh/known_hosts;ssh-keygen -R "
+                     + address  + " -f ~/.ssh/known_hosts >/dev/null 2>&1"], shell=True)
+
+    logfile_redirect = None
+    shell_cmd.replace("'", "\\\'")
+    if VERBOSITY >= 4:
+        print "remote_shell: Host =", address
+        print "remote_shell: Command =", shell_cmd
+
+    if VERBOSITY >= 9:
```
yyscamper: So many verbosity levels, could you please define them with enum?

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+    '''
+
+    logfile_redirect = file('/dev/null', 'w')
+    just_fname = os.path.basename(src_file_name)
+    # if localhost just copy to home dir
+    if ARGS_LIST['ora'] == 'localhost':
+        shutil.copyfile(src_file_name , '~/' + src_file_name)
+        return just_fname
+
+    scp_target = 'onrack@{0}:'.format(ARGS_LIST["ora"])
```
yyscamper: The  user name is better to be fetch from configuration.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+if ARGS_LIST['http'] == "True":
+    API_PROTOCOL = 'http'
+    API_PORT = GLOBAL_CONFIG['ports']['http']
+if ARGS_LIST['https'] == "True":
+    API_PROTOCOL = 'https'
+    API_PORT = GLOBAL_CONFIG['ports']['http']
+if ARGS_LIST["ora"] == "localhost":
+    API_PROTOCOL = 'http'
+    API_PORT = '8080'
+
```
yyscamper: Better to add some assertion logic here to check whether some required arguments are missing and whether there are illegal values so that following code will not fall into the trouble for argument validation.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+                        timeout=timeout, logfile=logfile_redirect)
+        return {'stdout':command_output, 'exitcode':exitstatus}
+
+    # this clears the ssh key from ~/.ssh/known_hosts
+    subprocess.call(["touch ~/.ssh/known_hosts;ssh-keygen -R "
+                     + address  + " -f ~/.ssh/known_hosts >/dev/null 2>&1"], shell=True)
+
+    logfile_redirect = None
+    shell_cmd.replace("'", "\\\'")
+    if VERBOSITY >= 4:
```
yyscamper: I see a lot judgement of `if VERBOSITY >= 4` and `if VERBOSITY >= 9`, It's not self-explained, change them to be more descriptive will be better.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+        print "remote_shell: Command =", shell_cmd
+
+    if VERBOSITY >= 9:
+        print "remote_shell: STDOUT =\n"
+        logfile_redirect = sys.stdout
+
+    if expect_receive == "" or expect_send == "":
+        (command_output, exitstatus) = \
+            pexpect.run("ssh -q -o StrictHostKeyChecking=no -t " + user + "@"
+                        + address + " sudo bash -c \\\"" + shell_cmd + "\\\"",
```
yyscamper: A little suggestion, use `format` will be more clear:
```python
"ssh -q -o StrictHostKeyChecking=no -t {0}@{1} sudo bash -c \\\"{2}\\\"".format(user, address, shell_cmd)
```

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+    idlist = [] # list of unique dcmi node IDs
+    # If we have already done this, use that list
+    if len(BMC_LIST) == 0:
+        ipscan = remote_shell('arp')['stdout'].split()
+
+        for ipaddr in ipscan:
+            if ipaddr[0:3] == "172" and remote_shell('ping -c 1 -w 5 ' + ipaddr)['exitcode'] == 0:
+                # iterate through all known IPMI users
+                for item in GLOBAL_CONFIG['credentials']['bmc']:
+                    return_code = remote_shell('ipmitool -I lanplus -H ' + ipaddr + ' -U ' + item['username'] \
```
yyscamper: I think wrap a util function to send IPMI command in remote shell will bring much benefit, as I see there are a lot other codes have similar code to send raw ipmi command.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+    remote_shell('rm -f /var/lib/dhcp/dhcpd.leases')
+    remote_shell('rm -f /var/log/onrack-conductor-event.log')
+    remote_shell('/opt/onrack/bin/monorail start')
+    if remote_shell('service onrack-conductor start')['exitcode'] > 0:
+        return 1
+    return 0
+
+def appliance_reset():
+    return_code = subprocess.call("ipmitool -I lanplus -H " + ARGS_LIST["bmc"] \
+                                  + " -U root -P 1234567 chassis power reset", shell=True)
```
yyscamper: Why hard code the user/password here? The RackHD appliance may setup in different hardwares.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+        # Find SKU ID
+        skumap = rackhdapi('/api/2.0/skus')
+        if skumap['status'] != 200:
+            print '**** Unable to retrieve SKU list via API.\n'
+            sys.exit(255)
+        for skuentry in skumap['json']:
+            if str(ARGS_LIST['sku']) in json.dumps(skuentry):
+                skuid = skuentry['id']
+        # Collect node IDs
+        catalog = rackhdapi('/api/2.0/nodes')
```
yyscamper: Little suggestion, rename `catalog` to `nodes` will not let developer confuse this is catalog data from catalog API.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+        if skumap['status'] != 200:
+            print '**** Unable to retrieve node list via API.\n'
+            sys.exit(255)
+        # Select node by SKU
+        for nodeentry in catalog['json']:
+            if ARGS_LIST["sku"] == 'all':
+                # Select only managed compute nodes
+                if nodeentry['type'] == 'compute':
```
yyscamper: Since only compute nodes are selected if ARG `skus` equals to `all`, whether that means if it is not `all`, it's an exception if that node is not compute?

test/fit_tests/common/pdu_lib.py
```diff
@@ -0,0 +1,161 @@
...
+    '14': 'eventOff',
+    '15': 'eventOn',
+    '16': 'eventReboot',
+    '17': 'eventShutdown'
+}
+
+def check_pdu_type(pduaddr=STACK_CONFIG[ARGS_LIST['stack']]['pdu']):
+    # execute check_pdu_type() to return PDU type if available, else return 'Unknown'
+    if remote_shell('ping -c 1 ' + pduaddr)['exitcode'] == 0:
+        if STACK_CONFIG[ARGS_LIST['stack']]['pdu'] == "192.168.1.254":
```
yyscamper: Hard code those IP address will make it hard to setup the FIT test in another environment.

test/fit_tests/deploy/rackhd_package_install.py
```diff
@@ -0,0 +1,210 @@
...
+                    "mongo": "mongodb://localhost/pxe",
+                    "sharedKey": "qxfO2D3tIJsZACu7UA6Fbw0avowo8r79ALzn+WeuC8M=",
+                    "statsd": "127.0.0.1:8125",
+                    "subnetmask": "255.255.252.0",
+                    "syslogBindAddress": "172.31.128.1",
+                    "syslogBindPort": 514,
+                    "tftpBindAddress": "172.31.128.1",
+                    "tftpBindPort": 69,
+                    "tftpRoot": "./static/tftp",
+                    "minLogLevel": 2
```
yyscamper: Duplicated `minLogLevel` with line 139. For testing, I suggest setting the minLogLevel to zero, so more data will be printed.

test/fit_tests/common/test_api_utils.py
```diff
@@ -0,0 +1,521 @@
...
+import fit_common
+
+import unicodedata
+import json
+
+##########################################
+#  Node List Utils
+##########################################
+
+def rackhd_get_node_list():
```
yyscamper: @jlongever had designed a feature that auto-generates the RackHD API library from RackHD swagger file, so we don't need to create a lot util functions for each API call, and also that library can be auto-synced even if there is some API change. see https://github.com/jlongever/on-http-api2.0-java

Are you aware of this feature? I think this feature this is pretty good.

test/fit_tests/config/global_config.json
```diff
@@ -0,0 +1,83 @@
...
+        "password": "admn"
+      }
+    ]
+  },
+  "snmp":{
+    "community": "onrack"
+  },
+  "repos": {
+    "_comment": "This is the list of repositories for each category",
+    "proxy": "http://web.hwimo.lab.emc.com:3128",
```
yyscamper: These EMC internal links are not suitable to be in the public repo.

test/fit_tests/deploy/onrack_stack_init.py
```diff
@@ -0,0 +1,306 @@
...
+    def test02_power_on_nodes(self):
+        # ServerTech PDU case
+        if pdu_lib.check_pdu_type() != "Unknown":
+            print '**** PDU found, powering on PDU outlets'
+            self.assertTrue(pdu_lib.pdu_control_compute_nodes("on"), 'Failed to power on all outlets')
+            # Wait about 30 seconds for the outlets to all come on and nodes to DHCP
+            fit_common.countdown(30)
+        # no PDU case
+        else:
+            print '**** No supported PDU found, restarting nodes using IMPI.'
```
yyscamper: typo: "IMPI" --> "IPMI"

test/fit_tests/deploy/onrack_stack_init.py
```diff
@@ -0,0 +1,306 @@
...
+                    }
+                },
+                "tasks": [
+                    {
+                        "label": "obm-ipmi-task",
+                        "taskName": "Task.Obm.Ipmi.CreateSettings"
+                    }
+            ]
+            }
+            api_data = fit_common.rackhdapi("/api/1.1/workflows", action="put", payload=payload)
```
yyscamper: I see these test code sometimes use API 1.1, sometimes use API 2.0.  Need to use the same API version.

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,528 @@
...
+        events={'(?i)assword: ':ARGS_LIST['pwd'] + '\n'},
+        logfile=logfile_redirect)
+    if VERBOSITY >= 4:
+        print "scp_file_to_ora: Exit Code = {0}".format(ecode)
+
+    assert ecode == 0, \
+        'failed "{0}" because {1}. Output={2}'.format(cmd, ecode, command_output)
+    return just_fname
+
+def rackhdapi(url_cmd, action='get', payload=[], timeout=None, rest_headers={}):
```
yyscamper: Suggest creating more util functions for each action:
```
rackhdapi_get
rackhdapi_post
rackhdapi_delete
rackhdapi_put
...
```

test/fit_tests/deploy/onrack_stack_init.py
```diff
@@ -0,0 +1,306 @@
...
+            if nodes_data['status'] == 200 and len(nodes_data['json']) > 0:
+                # if there are nodes present, determine if discovery has completed on them
+                discovery_complete = True
+
+                for node in nodes_data['json']:
+                    if node['type'] == 'compute':
+                        self.assertIn('id', node, 'node does not contain id')
+                        node_id = node['id']
+                        # determine if there are any active worlflows. If so, discovery not completed
+                        awf_data = fit_common.rackhdapi('/api/1.1/nodes/'
```
yyscamper: Don't mix up API 1.1 & API 2.0. Suggest to only use only one version.

test/fit_tests/common/test_api_utils.py
hohene: Was not aware of that, will take a look and bug you off line! thanks!

heckj: @hohene the on-http git repository includes bits to generate two additional debian packageswith a 1.1 and 2.0 python library from swagger definition files. Packages are pushed to bintray as well. Probably not the ideal place to store libraries for RackHD, but since they update with this repository, we've been using it to date for the current tests in https://github.com/rackhd/rackhd tests

test/fit_tests/common/fit_common.py
hohene: There are wrapper ipmi functions in test_api_utils, that could be moved into fit_common.  test_api_utils includes fit_common...so...can't include that here.

test/fit_tests/common/fit_common.py
hohene: We wanted more info logging levels for user output than what the python logger provided to use when running these tests individually and when capturing logs for runs.  That is where verbosity levels came at the time.  Will look at improving this and at least get enum values for verbosity -  there was a layout to the levels - will get that.

test/fit_tests/common/fit_common.py
gpaulos: Good point, this should be moved to global_config.json

test/fit_tests/common/pdu_lib.py
gpaulos: Needed for MN lab config only. PDU IP address is defined in stack config which is intended to be user-defined. I know this is clumsy, but until PDU support is better defined, this will have to stay.

test/fit_tests/deploy/onrack_stack_init.py
gpaulos: Agreed that we should migrate  to 2.0 but while this was being coded 2.0 APIs where changing rapidly so used 1.1 as a stable base. Migrating all base code to 2.0 is planned.

test/fit_tests/common/fit_common.py
hohene: Here's what we had for verbosity levels to use when scripting - shown in the help:
```
  -v V              Verbosity level of console output, default=0, Built Ins:
                    0: No debug, 2: User script output, 4: rest calls and
                    status info, 6: other common calls (ipmi, ssh), 9: all the
                    rest
```
Basically, 1-3 is reserved for user defined verbosity in scripts, then 4 and above would be logging levels reporting by the framework areas, so users could set verbosity to just see what they wanted from their test script.  That was the intention.

gpaulos:
Summary of recent changes to address review concerns:

- added '-port' parameter to allow setting http/https port at runtime 
- added support for Ubuntu 16.04
- Updated base code to API 2.0 
- Updated README files for clarity and completeness
- some bugfixes

heckj:
@gpaulos please also provide, either within the extended readme, or external documentation elsewhere, a full walk-through of how to use and invoke this with a Vagrant based instance so that I can see these tests running for myself on my own laptop's virtual environment. 👎 until that exists and is functional.

gpaulos:
Added Vagrant instructions to README per Joe Heck request.

test/fit_tests/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+    vagrant up quanta_d51
+
+Log into the Vagrant box:
+
+    vagrant ssh dev
+
+Clone FIT tests:
+
+    git clone https://github.com/onrack2/fit_tests
```
heckj: we wouldn't want to keep this in place once this PR is merged, correct? The test code will already be at `~/src/test` on the vagrant instance, perhaps use that instead?

test/fit_tests/README.md
```diff
@@ -0,0 +1,161 @@
...
+Load virtual environment:
+
+    cd fit_tests/tests/fit_tests
+    virtualenv .venv
+    source .venv/bin/activate
+    pip install -r requirements.txt
+
+Load SKU packs:
+
+    python run_tests.py -test util/load_sku_packs.py
```
heckj: the vagrant instance didn't include the python-dev package to allow this to work straight off the bat. For the meantime, I'd recommend adding: `sudo apt-get install python-dev` to this list

test/fit_tests/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+Run Smoke test suite:
+
+    python run_tests.py -test tests/ -group smoke
+
+Use run_tests.py options if needed such as '-port' '-ora' '-https' to select IP address, port and protocol.
```
heckj: Is the expectation that the quanta_d51 would have already gone through it's discovery and been processed entirely when the tests are invoked?

When I invoke it first, I get some failures, as happens when I invoke it after a discovery:

```
(.venv)vagrant@rackhd-demo:~/fit_tests/test/fit_tests$ python run_tests.py -test tests/ -group smoke
....F[
    "Nodeid: 57bcb91dc5b2bab107f6de91 Bad return code 404 for url: /redfish/v1/Chassis/57bcb91dc5b2bab107f6de91/Power",
    "Nodeid: QTFCJ05160195 Bad return code 404 for url: /redfish/v1/Chassis/QTFCJ05160195/Power"
]
F[
    "Nodeid: 57bcb91dc5b2bab107f6de91 Bad return code 404 for url: /redfish/v1/Chassis/57bcb91dc5b2bab107f6de91/Thermal",
    "Nodeid: QTFCJ05160195 Bad return code 404 for url: /redfish/v1/Chassis/QTFCJ05160195/Thermal"
]
F.................F...FFF................[
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8c, driveHealth failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8f, sel failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8e, selInformation failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8d, sdr failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de90, chassis failed to return any data"
]
F[
    "Error: Poller 57bcb91dc5b2bab107f6de8c ID: driveHealth - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de8f ID: sel - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de8e ID: selInformation - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de8d ID: sdr - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de90 ID: chassis - Pollers not running"
]
F[
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8c Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8f Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8e Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8d Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de90 Failed to get current poller data, status 404"
]
F[
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8c Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8f Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8e Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8d Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de90 Failed to get current poller data, status 404"
]
F[
    "Node: 57bcb7f9f1b359b307c0e320 Poller: driveHealth 57bcb91dc5b2bab107f6de8c reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: sel 57bcb91dc5b2bab107f6de8f reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: selInformation 57bcb91dc5b2bab107f6de8e reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: sdr 57bcb91dc5b2bab107f6de8d reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: chassis 57bcb91dc5b2bab107f6de90 reported 3 failureCount"
]
F...[
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8c, driveHealth failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8f, sel failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8e, selInformation failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de8d, sdr failed to return any data",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller ID 57bcb91dc5b2bab107f6de90, chassis failed to return any data"
]
F[
    "Error: Poller 57bcb91dc5b2bab107f6de8c ID: driveHealth - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de8f ID: sel - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de8e ID: selInformation - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de8d ID: sdr - Pollers not running",
    "Error: Poller 57bcb91dc5b2bab107f6de90 ID: chassis - Pollers not running"
]
F[
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8c Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8f Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8e Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8d Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de90 Failed to get current poller data, status 404"
]
F[
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8c Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8f Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8e Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de8d Failed to get current poller data, status 404",
    "Error: Node 57bcb7f9f1b359b307c0e320 Poller_ID 57bcb91dc5b2bab107f6de90 Failed to get current poller data, status 404"
]
F[
    "Node: 57bcb7f9f1b359b307c0e320 Poller: driveHealth 57bcb91dc5b2bab107f6de8c reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: sel 57bcb91dc5b2bab107f6de8f reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: selInformation 57bcb91dc5b2bab107f6de8e reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: sdr 57bcb91dc5b2bab107f6de8d reported 2 failureCount",
    "Node: 57bcb7f9f1b359b307c0e320 Poller: chassis 57bcb91dc5b2bab107f6de90 reported 3 failureCount"
]
... (more failures after this) ...
```

End result:
```
Ran 166 tests in 289.672s

FAILED (SKIP=16, failures=33)
```

Are there any intended coordination points for specifically powering on additional machines and triggering a discovery process, or being aware of machines that are expected to "be there", what they should be, and how that should be configured for a test environment?

I did this all per the instructions for a Vagrant environment - I would expect all the tests could pass for anything we're proposing.

test/fit_tests/deploy/rackhd_stack_init.py
```diff
@@ -0,0 +1,343 @@
...
+        for subdir, dirs, files in os.walk('on-skupack/tarballs'):
+            for skupacks in files:
+                print "\n**** Loading SKU Pack for " + skupacks
+                fit_common.rackhdapi("/api/1.1/skus/pack", action="binary-post",
+                                     payload=file(fit_common.TEST_PATH + "on-skupack/tarballs/" + skupacks).read())
+            break
+        print "\n"
+        # check SKU directory against source files
+        errorcount = 0
+        skulist = fit_common.json.dumps(fit_common.rackhdapi("/api/2.0/skus")['json'])
```
yyscamper: This test use both API 2.0 (line 53) and API 1.1 (line 47), I think we'd use the same API version here. 2.0 is recommend since 1.1 will be deprecated.

gpaulos: Yes, we need to migrate to API 2.0. This is the last API 1.1 in the common code. I would have migrated it but can't get it to work. Gives 'validation errors'. This should work with API 2.0 according to documentation. Any ideas?

test/fit_tests/README.md
gpaulos: It appears that the Vagrant environment that I have been using does not contain current code. FIT tests are continually updated to match the current API spec. Back code will fail. Also the code is changing on a daily basis. The FIT test passed fully for only one day last week. Then a series of merges broke a bunch of tests. PR-317 will be reverted due to serious flaw in the ipmi poller. It's a moving target.

gpaulos:
All reviewers: Note that FIT tests passed fully for only one day last Tuesday. Code changes and some regressions have resulted in over 50 test failures today. I have been busy updating the test repo the best I can to maintain the test base since this PR was opened. At some point, the code will stabilize. Until then, expect sporadic test failures from FIT.

EMC staffers can monitor our Jenkins RackHD FIT Smoke test here:
https://roebling.hwimo.lab.emc.com/jenkins/job/Running-Tests/job/RackHD-FIT-Smoke-Test/

gpaulos:
I have updated FIT with the API changes that have been merged as of 8/29. It looks clean as of the moment. I will monitor tests for other changes that are coming in that will impact the test base.

gpaulos:
After the latest FIT updates, here is the Vagrant smoke test results:

python run_tests.py -stack vagrant  -test tests/ -group smoke
SSSSSSSSSSSSSSSS...........................F...............................................................[
    "Error: Node 57c6e98407d360cc07923de8 Poller_ID 57c6ea81c617b2c107155e1d Failed to get current poller data, status 404"
]
...
more test output
...
----------------------------------------------------------------------
Ran 166 tests in 93.827s

FAILED (SKIP=16, failures=5)

The last few failures are due to missing poller data from the D51 simulator. This appears to be a simulator deficiency. The poller test is a critical component of the smoke test which discovered an important bug last week. I think that the poller tests should remain in the smoke test and the simulator be fixed.

heckj:
@gpaulos I'd tend to agree there - do you have any bugs filed with https://github.com/infrasim/infrasim/issues related to this deficiency?

gpaulos:
After some fiddling and rebooting, got the pollers working on the virtual D51. It has some timing differences and latency as compared to real hardware. It seems a bit fragile. Working to compose a stack init routine that will accommodate the virtual nodes.

Down to 2 failures:
----------------------------------------------------------------------
Ran 166 tests in 100.919s

FAILED (SKIP=16, failures=2)

These 2 failures represent the following issues:

/api/1.1/profiles/library - error message
/api/2.0/obms - invalid node URLs

These APIs work in our environment. Don't know what the difference is.

test/fit_tests/common/fit_common.py
nortonluo: That is a trick in case of different system prompt for "Password/password"

test/fit_tests/config/global_config.json
nortonluo: That is problem , the global_config file contains many internal links. @gpaulos can we create two global_config.json for public github and internal github, Only the one in internal Github have that internal links?

PengTian0:
test this please

test/fit_tests/common/fit_common.py
```diff
@@ -0,0 +1,631 @@
...
+        rest_headers.update({"authorization": "JWT " + AUTH_TOKEN, "X-Auth-Token": REDFISH_TOKEN})
+    # Perform rest request
+    try:
+        if rest_action == "get":
+            result_data = requests.get(url_command,
+                                       timeout=rest_timeout,
+                                       verify=sslverify,
+                                       headers=rest_headers)
+        if rest_action == "delete":
+            result_data = requests.delete(url_command,
```
nortonluo: For some API such as rack node relation ship management, it requires send json payload in http delete method. Can you add  data=json.dumps(rest_payload), in the request.delete call?

gpaulos:
Yes, I will add the payload to the delete action. Is there a possibility of a payload for the 'get' action?

nortonluo:
Yes, although there is no such 'get' with payload situation so far, add a payload here in case there is such API call in feature(if there is no other side effect)

test/fit_tests/deploy/rackhd_stack_init.py
yyscamper: I guess you are encountering this problem: https://github.com/RackHD/RackHD/issues/388

heckj:
per external email thread - merging as is, with expected following PRs to bring it into a gating mechanism for RackHD

Issue #345, opened 2016-07-28
missing ipxe binary artifact in on-imagebuilder


heckj:
from @codenrhoden 

> Any idea where I can find the file `monorail.intel.ipxe`, as referenced here: > https://github.com/RackHD/on-dhcp-proxy/blob/master/lib/message-handler.js#L257 ? 
> I tried grabbing it from bintray, as the `files` Docker image grabs the rest from a URL like this: "https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail.ipxe" but “monorail.intel.ipxe” does not exist there

heckj:
log into an ubuntu box (or spin up a vagrant instance):
 - `git clone https://github.com/rackhd/on-imagebuilder`
 - `cd on-imagebuilder`
 - run the iPXE build generation process
 - `sudo ansible-playbook -i hosts common/ipxe_builder.yml -e "config_file=vars/ipxe.yml"`

This builds iPXE in `/tmp/on-imagebuilder/ipxe_build_env/ipxe/src`

- `cd /tmp/on-imagebuilder/ipxe_build_env/ipxe/src`
- `sudo make bin/intel.pxe EMBED=default.ipxe`

grab the file at `/tmp/on-imagebuilder/ipxe_build_env/ipxe/src/bin/intel.pxe` and rename it to "monorail.intel.ipxe" and put it into the same TFTP static files location with the other bootloaders on your system (if you're doing this from source, that's `~/src/on-tftp/static/tftp`, if from debian packages, it's more likely `/var/renasar/on-tftp/static/tftp`.

codenrhoden:
The ipxe image this generates works fine for my NUC.  Thanks!  Once the fix is merged into on-imagebuilder, does the file get automically pushed up to bintray?

heckj:
@codenrhoden yep, it's dropped into the right place so it'll automatically populate. I've just put up another PR that will update our vagrant demo instance to make sure it gets added to local VMs as well when someone uses Ansible to install: https://github.com/RackHD/RackHD/pull/349

PR #380, opened 2016-08-12
Add intel iPXE file to "files" docker container

codenrhoden:
Since the `monorail.intel.ipxe` file is now built and uploaded to bintray by default, it should be added to the "files" container.

Reference #345


heckj:
/me smacks forehead. Mea Culpa!

:+1:

yyscamper:
:+1:

Issue #285, opened 2016-06-25
Many unused config values


benbp:
We have a bunch of unused/outdated/un-documented values in our sample config.json.

`CIDRNet` - appears unused
`broadcastaddr` - this could do with a rename (used for on-dhcp-proxy)
`apiServerAddress` - not compatible with our southbound proxy config data structure. Have to define the values in two places right now
`gatewayaddr` - appears unused
`minLogLevel` - this is defined twice in the example file (once with a value of 3, again with a value of 2). This makes it harder to ask people to change their log level when debugging!
`subnetmask` - appears unused

anhou:
during adjust DHCP IP range, I remove DHCP related unused parameters. PR is https://github.com/RackHD/RackHD/pull/386.  so after this PR.   CIDRNet ,subnetmask ,broadcastaddr will be removed. there is no 'gatewayaddr ' at present. so only below items are left:

`apiServerAddress` - not compatible with our southbound proxy config data structure. Have to define the values in two places right now
`minLogLevel` - this is defined twice in the example file (once with a value of 3, again with a value of 2). This makes it harder to ask people to change their log level when debugging!

PR #386, opened 2016-08-19
Adjust DHCP IP range and remove unused config parameters

anhou:
1. ODR-824. Adjust dhcp IP range to about <4096 from < 1024, in original dhcp range, it supports <1020 ips if exclude .0, and .255. so need to extend range to 2 times of original range to support 1024 compute nodes. and BMC also need the same amount of IPs, so need to extend to 4 times of original range

2. Remove unused config parameters such as subnetmask, broadcastaddr and CIDRNet which are related to DHCP, resolve parts of issue https://github.com/RackHD/RackHD/issues/285
related PR https://github.com/RackHD/on-dhcp-proxy/pull/55

3. a following RackHD/docs PR will be submitted

@RackHD/corecommitters @johren


yyscamper:
:+1:

brianparry:
:+1:

PR #400, opened 2016-09-01
remove duplicated minLogLevel & change minLogLevel to 0

yyscamper:
Change includes:
(1) Remove the duplicated minLogLevel
(2) Change minLogLevel to 0, so that the debug message can also be printed. This will help to identify issue in Jenkin CI (The Jenkins uses this config in testing) and customer reported bugs.

@heckj @RackHD/corecommitters @jlongever   @panpan0000 @PengTian0 @changev

### DIFF
--- a/packer/ansible/roles/monorail/files/config.json
+++ b/packer/ansible/roles/monorail/files/config.json
@@ -41,7 +41,6 @@
         "remotePath": "/amd64-usr"
     }],
     "httpStaticRoot": "/opt/monorail/static/http",
-    "minLogLevel": 3,
     "authUsername": "admin",
     "authPasswordHash": "KcBN9YobNV0wdux8h0fKNqi4uoKCgGl/j8c6YGlG7iA0PB3P9ojbmANGhDlcSBE0iOTIsYsGbtSsbqP4wvsVcw==",
     "authPasswordSalt": "zlxkgxjvcFwm0M8sWaGojh25qNYO8tuNWUMN4xKPH93PidwkCAvaX2JItLA3p7BSCWIzkw4GwWuezoMvKf3UXg==",


panpan0000:
good catch to remove the duplicated loglevel definiton.

if the loglevel 0 will not bring system slow down ,and the burst of debug message does helpful to developers ( including external users who use Vagrant releases ), then fine to me .

yyscamper:
Jenkins: test this please

benbp:
👍

heckj:
I'd prefer to leave the minLogLevel in the default configuration to 3 (info level), simply because I think it would be a better experience for people trying out RackHD with Vagrant to not get all the additional detail, which is prominently displayed based on how we run the example setup from source.

I totally agree we should reset this as needed for instances like Jenkins where we could use the additional debugging. I *think* it would be easier to make the one-time change needed for jenkins, or developers can make the change manually in their own local systems if they want to see increased debugging.

Alternatively, we can make the public Vagrant instance of RackHD with an alternate file that has the decreased debugging logging.

@panpan0000 how are you using vagrant in your CI setup? Are you pulling directly from rackhd/rackhd vagrant image and using that directly? Are you building your own locally?

PengTian0:
test this please

panpan0000:
@heckj  , you are right. 
it should use a customized config.json , there does have one, but never take effect.
I made an PR as below:
https://github.com/RackHD/on-build-config/pull/19

But Felix's PR still valid to remove duplicated twice Log-level definition :-) 

@yyscamper , can you rework your PR?

yyscamper:
@panpan0000 , @heckj : I will keep the previous information log level

yyscamper:
@panpan0000 @heckj : done!

PengTian0:
test this please

anhou:
:+1:

cgx027:
:+1:

PR #411, opened 2016-09-07
updates for Day 1 of EMC -> Dell merger

heckj:
Note: Copyrights and related details *do not* need to be proactively changed. We can change them as we would any other copyright update - when we edit the files as we go forward.

Also I didn't include the email address change from rackhd@emc.com to rackhd@dell.com because the email alias hasn't yet been migrated. There's a ticket with IT where I'm tracking that transition, and I'll update relevant docs and links afterwards.


DavidjohnBlodgett:
👍

Issue #407, opened 2016-09-06
local-user exception for login to create initial user no longer functions - packer builds failing


heckj:
Packer builds with the latest code are failing. The issue stems from the packer authentication setup script "firstuser.sh" not succeeding. 

Relevant output from packer (with `PACKER_LOG=1` prefixed)
```
2016/09/06 12:11:04 ui:     virtualbox-iso: * Started on-http daemon
    virtualbox-iso: * Started on-http daemon
2016/09/06 12:11:04 ui:     virtualbox-iso: * Started on-taskgraph daemon
    virtualbox-iso: * Started on-taskgraph daemon
2016/09/06 12:11:04 ui:     virtualbox-iso: * Started on-dhcp-proxy daemon
    virtualbox-iso: * Started on-dhcp-proxy daemon
2016/09/06 12:11:04 ui:     virtualbox-iso: * Started on-syslog daemon
    virtualbox-iso: * Started on-syslog daemon
2016/09/06 12:11:04 ui:     virtualbox-iso: * Started on-tftp daemon
    virtualbox-iso: * Started on-tftp daemon
2016/09/06 12:11:04 ui:     virtualbox-iso: waiting for rackhd services..
    virtualbox-iso: waiting for rackhd services..
2016/09/06 12:11:25 ui:     virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
2016/09/06 12:11:35 ui:     virtualbox-iso: Username/Password Authentication Failed.
2016/09/06 12:11:45 ui:     virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
2016/09/06 12:11:55 ui:     virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
2016/09/06 12:12:05 ui:     virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
2016/09/06 12:12:15 ui:     virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
2016/09/06 12:12:25 ui:     virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
    virtualbox-iso: Username/Password Authentication Failed.
[...]

PR #415, opened 2016-09-13
fix packer build issue

keedya:
resolves  https://github.com/RackHD/RackHD/issues/407

@heckj @RackHD/corecommitters @jlongever @amymullins @zyoung51 @uppalk1 

here is a link to jenkins job with this fix:
http://rackhdci.lss.emc.com/job/vagrant-build/77/

### DIFF
--- a/packer/scripts/firstuser.sh
+++ b/packer/scripts/firstuser.sh
@@ -44,7 +44,7 @@ stopServices() {
 waitForServices() {
   local attempt=0
   local maxto=20
-  local url=${HTTP_URL}/api/2.0/nodes
+  local url=${HTTP_URL}/swagger-ui
   echo "waiting for rackhd services.. "
   while [ ${attempt} != ${maxto} ]; do
     echo -ne "waited `expr $attempt \* 10` seconds\r"


heckj:
nice Andre, thank you!

👍 

Worth nothing - change the PR description to say "resolves #407" to get this to resolve that issue when this is merged...

jlongever:
Is the HTTP endpoint still necessary since authentication is enabled on the HTTPS endpoint?  Are both TLS/SSL and non-TLS/SSL endpoints needed?

Could the HTTPS be removed from config.json and firstuser.sh script? https://github.com/RackHD/RackHD/pull/415/files#diff-352352fdea3bd8f0c8601a2f1b1fbec0R6
https://github.com/RackHD/RackHD/blob/master/packer/ansible/roles/monorail/files/config.json#L12

heckj:
I would be totally game for that - probably means we need to update the tutorial a bit, but we should head this direction to be sane and safe

jlongever:
Agree-  having two authenticated endpoints seems redundant to me;  unless we want to demonstrate both encrypted and non-encrypted methods?

amymullins:
@RackHD/corecommitters can you help with a review / merge? This PR will fix the broken packer build issue. 

@brianparry , can you help confirm a couple of @jlongever 's questions as this was enabled as part of the 2.0 API feature:
_Is the HTTP endpoint still necessary since authentication is enabled on the HTTPS endpoint? Are both TLS/SSL and non-TLS/SSL endpoints needed?_

brianparry:
@jlongever We've been planning to get rid of the HTTP endpoint as part of the 1.1 API deprecation effort.  IMO, we should merge this PR as is and handle the endpoint stuff separately.

PR #416, opened 2016-09-14
Log node discovery task failure

jlongever:
- Log the entire workflow task data on a node discovery task failure.
- ACK message if graph and route ID's match.

@RackHD/corecommitters @changev @panpan0000


anhou:
:+1:

Issue #270, opened 2016-06-14
The skip condition of Testing Graph.Discovery  is not  rigorous enough


changev:
Now when found data **not null** in 
```
/api/1.1/nodes
```
the Testing Graph.Discovery will be **skipped**. 

But Graph.Discovery **not only** writes the nodes' data in mongo but also does other jobs like collecting catalogs, so finding data in **/api/1.1/nodes** doesn't mean  Graph.Discovery is over.
Another side ,if **not skipped**, the Graph.Discovery will monitor amqp info and **make sure** Graph.Discovery be finished.
So the judging conditions of the two flows are not symmetrical.

Then in some cases, after rackhd service is running  and all the nodes are power on for test ,but  there are **too many** nodes so the power on spends more time hence before Test Graph.Discovery some nodes has been **imperfect** discovered(/api/1.1/nodes not null but catalogs null), as a result  Test Graph.Discovery will be skipped and next test cases will fail.

So i recommand enhancement of the skip condition in Testing Graph.Discovery.

benbp:
I'm a little confused as to the problem description here. Currently the criteria used for whether a node should be discovered is whether it has catalogs or not (see https://github.com/RackHD/on-core/blob/master/lib/models/node.js#L74-L83) and whether there is already an active workflow for it (https://github.com/RackHD/on-http/blob/88b7f5981cd37f67717ff77b97cd6dd3690a205b/lib/services/profiles-api-service.js#L107-L116).

Are you referring solely to RackHD here, or also to OnRack, or perhaps to the RackHD integration test suite, or something else? I don't believe there are any places within the RackHD code that make API calls to `/api/1.1/nodes` in order to determine behavior, but there probably are within OnRack and our integration tests. Which Graph.Discovery test case are you specifically referring to?

changev:
Thanks for your attention. 
It's my fault that I didn't tell this is Graph.Discovery in **proboscis function test**(see https://github.com/RackHD/RackHD/blob/master/test/tests/api/v1_1/nodes_tests.py#L113-115).
and the ```check_compute_count``` just use ```Nodes().nodes_get()``` which get data from ```/api/1.1/nodes```.

benbp:
Ah, ok that makes sense @changev thanks. Yep it seems like we could be doing that better with a catalog check.

changev:
@benbp thanks.

It's a good way to add catalog checking in "already discovered" criteria, but depends on our testing experience maybe it's imperfect too.
[...]

PR #428, opened 2016-09-22
fix node discover test case issue: RackHD/RackHD issue #270

changev:
enhance https://github.com/RackHD/RackHD/issues/270

Import a external value **NODES_NUMBER** ,which means how many nodes are used for test.

Enhanced node discovery test case skip condition, only if all nodes are in mongodb this case skip.

@iceiilin @panpan0000 @RackHD/corecommitters    

Jenkins: depends on https://github.com/RackHD/on-build-config/pull/23

### DIFF
--- a/test/tests/api/v1_1/nodes_tests.py
+++ b/test/tests/api/v1_1/nodes_tests.py
@@ -106,7 +106,8 @@ class NodesTests(object):
     @test(groups=['nodes.discovery.test'])
     def test_nodes_discovery(self):
         """ Testing Graph.Discovery completion """
-        if self.check_compute_count():
+        count = defaults.get('RACKHD_NODE_COUNT', '')
+        if (count.isdigit() and self.check_compute_count() == int(count)) or self.check_compute_count():
             LOG.warning('Nodes already discovered!')
             return
         self.__discovery_duration = datetime.now()
--- a/test/tests/api/v2_0/nodes_tests.py
+++ b/test/tests/api/v2_0/nodes_tests.py
@@ -119,7 +119,8 @@ class NodesTests(object):
     @test(groups=['nodes.api2.discovery.test'])
     def test_nodes_discovery(self):
         """ API 2.0 Testing Graph.Discovery completion """
-        if self.check_compute_count():
+        count = defaults.get('RACKHD_NODE_COUNT', '')
+        if (count.isdigit() and self.check_compute_count() == int(count)) or self.check_compute_count():
             LOG.warning('Nodes already discovered!')
             return
         self.__discovery_duration = datetime.now()


changev:
test this please

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

changev:
test this please

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

changev:
test this please

iceiilin:
It is robust to cover the corner case when nodes are being discovered and not ready for further catalog test. :+1: 
@jlongever

changev:
test this please

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

jlongever:
It seems like a lot of changes to add a fixed node count option.  Could you do something similar to this: https://github.com/jlongever/RackHD/commit/acda755022920ba4ed90313a3a5cd176b6e70357 ?

Also keep in mind that with a fixed node count all of the nodes will have to be deleted once you get to the next 2.0 test group, otherwise you'll always timeout.  Because this change creates a hard expectation to wait for `n` number of discovery graphs to complete for each API node discovery test.  Maybe add a dependent test to clear out all node documents prior to running the `node.discovery.test` if `RACKHD_NODES_NUMBER` is defined?

rackhdmirror:
Build finished.

test/tests/api/v1_1/nodes_tests.py
```diff
@@ -92,21 +94,28 @@ def __post_workflow(self, id, graph_name, data):
...
-        count = 0
+        discovered_count = 0
         for n in nodes:
             type = n.get('type')
             if type == 'compute':
-                count += 1
-        return count
+                id = n.get('id')
+                status = self.__get_workflow_status(id)
+                if status != 'pending' and status != 'running':
```
jlongever: This is probably not safe.   You're assuming the status will be correct which we have seen in the past its not always the case.  The status in AMQP is probably more reliable.  Secondly,  it assumes the status is for discovery graph,  other tests run workflows against nodes.  so by the time you get to 2.0 test groups,  you could be looking at workflow status for other workflows executed in prior tests.   It would probably be safer to just go through each node and check if the catalog is fulfilled.

changev: Thanks, I got it. 

So in V1_1 and V2_0 tests, instead of "find in db and no related active workflow", we should check catalogs and make sure a node is actually discovered.

"to_be_discovered" is still useful, right?

jlongever: IMHO, I think the logic is a bit unnecessary.  Something like https://github.com/jlongever/RackHD/commit/acda755022920ba4ed90313a3a5cd176b6e70357 but include node catalog checks if the test is using a fixed count?

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

changev:
test this please

rackhdmirror:
Build finished.

test/tests/api/v1_1/nodes_tests.py
```diff
@@ -106,9 +106,11 @@ def get_undiscovered_compute_count(self):
                 workflows = self.__get_data()
                 LOG.info(workflows)
                 for workflow in workflows:
-                    if workflow.get['injectableName'] == 'Graph.SKU.Discovery' \
-                     and workflow.get['status'] == 'succeeded':
-                     discovered_count += 1
+                    definition = workflow.get('definition', {})
+                    injectableName = definition.get('injectableName')
+                    if injectableName == 'Graph.SKU.Discovery':
+                        workflow.get['_status'] == 'succeeded':
```
jlongever: As I mentioned before,  this isn't going to be reliable.

changev: a succeeded Graph.SKU.Discovery workflow doesn't means the node has been discovered successfully?

changev: yeah i see "You're assuming the status will be correct which we have seen in the past its not always the case, The status in AMQP is probably more reliable" But why it's not reliable

jlongever:
@changev Sorry,  I think this PR is a -1 for me.  It seems reasonable to add an optional fixed node count as an enhancement,  but I'm not quite sure what "bug" these changes are fixing.  The random node discovery failures are more likely environment based (e.g. nodes failing to PXE boot,  or missed `graph.finished` AMQP events, etc) than they are test logic based.

Fwiw,  the node discovery tests haven't changed much,  and for the most part when we see random discovery failures,  its due to a real regression or test env change.

rackhdmirror:
Build finished.

changev:
Ok ,thanks joel.  I realize that you haven't known the bug clearly yet.
The "bug" has shown clearly in related issue. When add more vnodes for testing, especially add physic nodes(spend more time for get ip and discover), this issue **is almost 100% occurs.** 
**Before all nodes up, some nodes has been write to mongodb and all nodes are seen as discovered ,the discovery test skip , then tests failed**. This is indeed a wrong logic.

1 I think the key now is why "a succeeded Graph.SKU.Discovery workflow" not reliable.
2 If it's true not reliable, is there any other solution you suggested can make sure a node has been discovered? Catalogs doesn't has a "discovered field" and different vnodes has different kinds of catalogs so we can't count them.

jlongever:
@changev As I mentioned,  adding an optional fixed node count helps with boot time variances when using different systems (virtual & physical).  But that is only an enhancement since the current env doesn't mix systems. 

1) interrogate the rackhd logs, and the environment when we see a random node failure to root cause (See my comment in https://github.com/RackHD/RackHD/issues/270)
2) the proper way to determine if the discovery graph completed,  is to listen for `graph.finished` events on AMQP.  The test already does that.  We have seen that polling the /workflows is less reliable because the `status` is not always updated in mongo when it is checked.

zyoung51:
@changev Based on your comment above this is a -1 for me too.  The solution to the test problem should not be to ignore "unreliable" things in the code under test and the root cause should be understood before the fix is submitted.  What root caused issue are you trying to fix by using a hard coded node count?

rackhdmirror:
Build finished.

changev:
@jlongever 
```
count = defaults.get('RACKHD_NODE_COUNT', '')
if count.isdigit() and self.check_compute_count() == int(count):`
```

self.check_compute_count() == int(count) is enough to solve the time issue.

jlongever:
@changev it would be safer to check `isdigit()` as well.

changev:
@jlongever 
already checked, thanks

jlongever:
@changev thanks,  missed the commit update.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

changev:
test this please

cgx027:
looks good to me. :+1:

yyscamper:
@jlongever @zyoung51 : bases on the latest change, do you still have any concern for this PR?

PR #431, opened 2016-09-28
Updates to FIT test common library

gpaulos:
This is an update to the FIT common library that includes several required fixes:
- The primary change is to the appy_obm_settings(). This is a new routine to apply OBM settings in parallel rather than the old sequential routing. The legacy routine has been renamed to apply_obm_settings_seq(). This has code been tested for about two weeks in the MN lab.
- check_active_workflows() needed revision to be compatible with the new 'status' field in the workflows API JSON. OnRack 2.1 was branched before that change so the old '_status' field needs to be supported as well.
- restful() 'binary-put' and 'binary-post' content type was changed to "application/x-www-form-urlencoded" for compatibility with API 2.0. Now the /api/2.0/skus/pack upload will work.
- Line 217 still had a reference to the appliance 'onrack' username which was changed to the global parameter ARGS_LIST['usr']

### DIFF
--- a/test/fit_tests/common/fit_common.py
+++ b/test/fit_tests/common/fit_common.py
@@ -615,7 +615,7 @@ def cancel_active_workflows(nodeid):
        exitstatus = False
     return exitstatus
 
-def apply_obm_settings():
+def apply_obm_settings(retry=30):
     # New routine to install OBM credentials via workflows in parallel
     count = 0
     for creds in GLOBAL_CONFIG['credentials']['bmc']:
@@ -672,7 +672,7 @@ def apply_obm_settings():
 
     # run each OBM credential workflow on each node in parallel until success
     nodestatus = {} # dictionary with node IDs and status of each node
-    for dummy in range(0, 30):
+    for dummy in range(0, retry):
         nodelist = node_select()
         for node in nodelist:
             if node not in nodestatus:
--- a/test/fit_tests/common/fit_common.py
+++ b/test/fit_tests/common/fit_common.py
@@ -214,7 +214,7 @@ def scp_file_to_ora(src_file_name):
         remote_shell('cp ' + src_file_name + ' ~/' + src_file_name)
         return src_file_name
 
-    scp_target = 'onrack@{0}:'.format(ARGS_LIST["ora"])
+    scp_target = ARGS_LIST['usr'] + '@{0}:'.format(ARGS_LIST["ora"])
     cmd = 'scp -o StrictHostKeyChecking=no {0} {1}'.format(src_file_name, scp_target)
     if VERBOSITY >= 4:
         print "scp_file_to_ora: '{0}'".format(cmd)
@@ -363,7 +363,7 @@ def restful(url_command, rest_action='get', rest_payload=[], rest_timeout=None, 
                                        verify=sslverify,
                                        )
         if rest_action == "binary-put":
-            rest_headers.update({"Content-Type": "application/octet-stream"})
+            rest_headers.update({"Content-Type": "application/x-www-form-urlencoded"})
             result_data = requests.put(url_command,
                                        data=rest_payload,
                                        headers=rest_headers,
@@ -386,7 +386,7 @@ def restful(url_command, rest_action='get', rest_payload=[], rest_timeout=None, 
                                         verify=sslverify
                                         )
         if rest_action == "binary-post":
-            rest_headers.update({"Content-Type": "application/octet-stream"})
+            rest_headers.update({"Content-Type": "application/x-www-form-urlencoded"})
             result_data = requests.post(url_command,
                                         data=rest_payload,
                                         headers=rest_headers,
@@ -587,18 +587,23 @@ def get_node_sku(nodeid):
                 if VERBOSITY >= 2:
                     errmsg = "Error: SKU API failed {}, return code {} ".format(sku, skudata['status'])
                     print errmsg
+                    return "unknown"
         else:
-            if VERBOSITY >= 2:
-                errmsg = "Error: nodeid {} did not return a valid sku in get_rackhd_nodetype{}".format(nodeid,sku)
-                print errmsg
+            return "unknown"
     return nodetype
 
 def check_active_workflows(nodeid):
     # Return True if active workflows are found on node
     workflows = rackhdapi('/api/2.0/nodes/' + nodeid + '/workflows')['json']
     for item in workflows:
-        if 'running' in item['_status'] or 'pending' in item['_status']:
-            return True
+        if '_status' in item:
+            if item['_status'] in ['running', 'pending']:
+                return True
+        if 'status' in item:
+            if item['status'] in ['running', 'pending']:
+                return True
+        else:
+            return False
     return False
 
 def cancel_active_workflows(nodeid):
@@ -610,8 +615,8 @@ def cancel_active_workflows(nodeid):
        exitstatus = False
     return exitstatus
 
-def apply_obm_settings_new():
-    # Experimental routine to install OBM credentials via workflows
+def apply_obm_settings():
+    # New routine to install OBM credentials via workflows in parallel
     count = 0
     for creds in GLOBAL_CONFIG['credentials']['bmc']:
         # greate graph for setting OBM credentials
@@ -666,29 +671,28 @@ def apply_obm_settings_new():
         count += 1
 
     # run each OBM credential workflow on each node in parallel until success
-    nodelist = node_select()
     nodestatus = {} # dictionary with node IDs and status of each node
-    for node in nodelist:
-        nodestatus[node]= {"status": "pending", "instanceId": "", "sku": get_node_sku(node), "retry": 0}
-    for dummy in range(0, 60):
+    for dummy in range(0, 30):
+        nodelist = node_select()
+        for node in nodelist:
+            if node not in nodestatus:
+                nodestatus[node] = {"status": "pending", "instanceId": "", "sku": get_node_sku(node), "retry": 0}
         for num in range(0, count):
             for node in nodelist:
-                skuid = rackhdapi('/api/2.0/nodes/' + node)['json'].get("sku")
-                skudata = rackhdapi(skuid)['text']
-                if "rmm.data.MAC" in skudata:
-                    workflow = {"name": 'Graph.Obm.Ipmi.CreateSettings.RMM' + str(num)}
-                else:
-                    workflow = {"name": 'Graph.Obm.Ipmi.CreateSettings' + str(num)}
                 # try workflow
                 if nodestatus[node]['status'] == "pending":
-                    for dummy in range(0, 60):
-                        # retry if other workflows active
+                    skuid = rackhdapi('/api/2.0/nodes/' + node)['json'].get("sku")
+                    if skuid:
+                        if nodestatus[node]['sku'] == "unknown":
+                            nodestatus[node].update({"sku": get_node_sku(node)})
+                        skudata = rackhdapi(skuid)['text']
+                        if "rmm.data.MAC" in skudata:
+                            workflow = {"name": 'Graph.Obm.Ipmi.CreateSettings.RMM' + str(num)}
+                        else:
+                            workflow = {"name": 'Graph.Obm.Ipmi.CreateSettings' + str(num)}
                         result = rackhdapi("/api/2.0/nodes/"  + node + "/workflows", action="post", payload=workflow)
                         if result['status'] == 201:
-                            nodestatus[node].update({"status": "running", "instanceId": result['json']["instanceId"], "retry": 0})
-                            break
-                        else:
-                           time.sleep(5)
+                            nodestatus[node].update({"status": "running", "instanceId": result['json']["instanceId"]})
             for node in nodelist:
                 # check OBM workflow status
                 if nodestatus[node]['status'] == "running":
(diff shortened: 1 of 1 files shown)


test/fit_tests/common/fit_common.py
```diff
@@ -363,7 +363,7 @@ def restful(url_command, rest_action='get', rest_payload=[], rest_timeout=None,
                                        verify=sslverify,
                                        )
         if rest_action == "binary-put":
```
heckj: rather than track the content-headers in a common library here, why not use the swagger generated python libraries that map directly to the swagger definition file? It would remove this sort of issue from the future, as well as give you immediately expanded scope for actions to be used in test cases.

test/fit_tests/common/fit_common.py
```diff
@@ -587,18 +587,23 @@ def get_node_sku(nodeid):
                 if VERBOSITY >= 2:
                     errmsg = "Error: SKU API failed {}, return code {} ".format(sku, skudata['status'])
                     print errmsg
+                    return "unknown"
         else:
-            if VERBOSITY >= 2:
-                errmsg = "Error: nodeid {} did not return a valid sku in get_rackhd_nodetype{}".format(nodeid,sku)
-                print errmsg
+            return "unknown"
```
heckj: is the string "unknown" here a better response than either an empty string or a null response? I suspect a semantically better response would be to return null rather than a value, which could be checked more easily to understand if the result wasn't a value as opposed to looking for a specific string `unknown` that represented an unknown or unset value for the SKU name

test/fit_tests/common/fit_common.py
```diff
@@ -666,29 +671,28 @@ def apply_obm_settings_new():
         count += 1
 
     # run each OBM credential workflow on each node in parallel until success
-    nodelist = node_select()
     nodestatus = {} # dictionary with node IDs and status of each node
-    for node in nodelist:
-        nodestatus[node]= {"status": "pending", "instanceId": "", "sku": get_node_sku(node), "retry": 0}
-    for dummy in range(0, 60):
+    for dummy in range(0, 30):
```
heckj: where does `30` (or the previous `60`) come from in terms of length? Would this be better either being passed in a argument to this function or deterministic based on the number of nodes discovered in the system?

test/fit_tests/common/fit_common.py
```diff
@@ -703,18 +707,27 @@ def apply_obm_settings_new():
...
+                                        if obmref:
+                                            rackhdapi(obmref, action="delete")
+
+        if VERBOSITY >= 4:
             print "**** Node(s) OBM status:\n", json.dumps(nodestatus, sort_keys=True, indent=4,)
         if "pending" not in str(nodestatus) and "running" not in str(nodestatus):
             # All OBM settings successful
             return True
-        time.sleep(10)
+        time.sleep(30)
```
heckj: Am I reading the code correctly that using this functional will mandate a minimum time to invoke this method as being `30` seconds (from this `time.sleep` function) times the 30 count above, making it take 15 minutes to ever invoke this method, regardless of how long it takes the various workflows to complete?

Can we improve this to optimistically escape the range count if all the workflows come back with a "workflow complete, success" marker?

Is there any intent that a failure of one of the workflows is exposed from within this function beyond the logging? (i.e. is the intent here that this is a setup/utility method that will work 99% of the time, or is this the core of a use case/test that would be better served with more explicit error information when it fails?)

rackhdmirror:

test/fit_tests/common/fit_common.py
gpaulos: These timeouts were experimentally determined using real hardware under a lot of different conditions. They could be parameterized and defaulted to these values.

test/fit_tests/common/fit_common.py
gpaulos: This routine was designed to more efficiently set OBMs by running in parallel. The routine will complete when all of the workflows end successfully. If one or more node fails, they it will run for the full 15 minutes. The routine puts out a message into logs or to the console if running verbosity of 4 or greater that shows the node list, node ID, workflow ID and completion status on every retry.

Example:
**** Node(s) OBM status:
{
    "57ec10b586574e9007966338": {
        "instanceId": "c71ad907-99f2-4064-af93-bccb0f517241",
        "retry": 6,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec10ee86574e900796633f": {
        "instanceId": "14b7f7fb-d3e0-4aed-9b42-dc9b73b287d6",
        "retry": 6,
        "sku": "Phoenix",
        "status": "running"
    },
    "57ec110586574e9007966346": {
        "instanceId": "b2dbe6b2-18bf-451c-bf9e-699c0006bcbe",
        "retry": 6,
        "sku": "Kylin",
        "status": "running"
    }
}

test/fit_tests/common/fit_common.py
gpaulos: An unset sku is presented in the logs as an "unknown" sku string. This could be a null value, but it is easier and clearer as a text string, at least in my opinion.

test/fit_tests/common/fit_common.py
gpaulos: That is a good future enhancement. Let's write up a set of test enhancements in a story queue. We are going to be migrating off Jira. Should these be on Pivotal?

test/fit_tests/common/fit_common.py
gpaulos: This routine was designed to more efficiently set OBMs by running in parallel. The routine will complete when all of the workflows end successfully. If one or more node fails, they it will run for the full 15 minutes. The routine puts out a message into logs or to the console if running verbosity of 4 or greater that shows the node list, node ID, workflow ID and completion status on every retry.

Example:
**** Node(s) OBM status:
{
    "57ec10b586574e9007966338": {
        "instanceId": "c71ad907-99f2-4064-af93-bccb0f517241",
        "retry": 6,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec10ee86574e900796633f": {
        "instanceId": "14b7f7fb-d3e0-4aed-9b42-dc9b73b287d6",
        "retry": 6,
        "sku": "Phoenix",
        "status": "running"
    },
    "57ec110586574e9007966346": {
        "instanceId": "b2dbe6b2-18bf-451c-bf9e-699c0006bcbe",
        "retry": 6,
        "sku": "Kylin",
        "status": "running"
    }
}

gpaulos: This routine was designed to more efficiently set OBMs by running in parallel. The routine will complete when all of the workflows end successfully. If one or more node fails, they it will run for the full 15 minutes. The routine puts out a message into logs or to the console if running verbosity of 4 or greater that shows the node list, node ID, workflow ID and completion status on every retry.

Example:
**** Node(s) OBM status:
{
    "57ec10b586574e9007966338": {
        "instanceId": "c71ad907-99f2-4064-af93-bccb0f517241",
        "retry": 6,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec10ee86574e900796633f": {
        "instanceId": "14b7f7fb-d3e0-4aed-9b42-dc9b73b287d6",
        "retry": 6,
        "sku": "Phoenix",
        "status": "running"
    },
    "57ec110586574e9007966346": {
        "instanceId": "b2dbe6b2-18bf-451c-bf9e-699c0006bcbe",
        "retry": 6,
        "sku": "Kylin",
        "status": "running"
    }
}

Here is the completion after 95 seconds:

**** Node(s) OBM status:
{
    "57ec10b586574e9007966338": {
        "instanceId": "c71ad907-99f2-4064-af93-bccb0f517241",
        "retry": 6,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec10ee86574e900796633f": {
        "instanceId": "14b7f7fb-d3e0-4aed-9b42-dc9b73b287d6",
        "retry": 7,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec110586574e9007966346": {
        "instanceId": "b2dbe6b2-18bf-451c-bf9e-699c0006bcbe",
        "retry": 7,
        "sku": "Kylin",
        "status": "succeeded"
    }
}
.
----------------------------------------------------------------------
Ran 1 test in 95.030s

gpaulos: This routine was designed to more efficiently set OBMs by running in parallel. The routine will complete when all of the workflows end successfully. If one or more node fails, they it will run for the full 15 minutes. The routine puts out a message into logs or to the console if running verbosity of 4 or greater that shows the node list, node ID, workflow ID and completion status on every retry.

Example:
**** Node(s) OBM status:
{
    "57ec10b586574e9007966338": {
        "instanceId": "c71ad907-99f2-4064-af93-bccb0f517241",
        "retry": 6,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec10ee86574e900796633f": {
        "instanceId": "14b7f7fb-d3e0-4aed-9b42-dc9b73b287d6",
        "retry": 6,
        "sku": "Phoenix",
        "status": "running"
    },
    "57ec110586574e9007966346": {
        "instanceId": "b2dbe6b2-18bf-451c-bf9e-699c0006bcbe",
        "retry": 6,
        "sku": "Kylin",
        "status": "running"
    }
}

Here is the completion after 95 seconds:

**** Node(s) OBM status:
{
    "57ec10b586574e9007966338": {
        "instanceId": "c71ad907-99f2-4064-af93-bccb0f517241",
        "retry": 6,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec10ee86574e900796633f": {
        "instanceId": "14b7f7fb-d3e0-4aed-9b42-dc9b73b287d6",
        "retry": 7,
        "sku": "Phoenix",
        "status": "succeeded"
    },
    "57ec110586574e9007966346": {
        "instanceId": "b2dbe6b2-18bf-451c-bf9e-699c0006bcbe",
        "retry": 7,
        "sku": "Kylin",
        "status": "succeeded"
    }
}
.
----------------------------------------------------------------------
Ran 1 test in 95.030s

The timeout used to be shorter, but the new SEL poller routines really dragged down OBM processing so it really does sometimes take over 10 minutes.

diontje:
+1 looks good to me.

test/fit_tests/common/fit_common.py
hohene: At a certain point, we have to verify that the API calls we are making are actually valid with the intended format of the API call.  Using the restful interfaces allows us to test that content and return codes of the calls are correctly implemented.  This catches unintended changes to the APIs.  Using the swagger generated python library bypasses one layer of that by using the generated code.  To me, we should have at least one suite of actual API tests that verify in the expected user format off the box, and then most other tests could use the swagger libraries.  So I think we need to keep this in place for a base API specific test suite.

test/fit_tests/common/fit_common.py
hohene: I am fine with the 30 second update, as the max time it takes will be 15 minutes, as we will exit early if all complete successfully.
I'm confused on the comment regarding SEL poller routines and how it relates to the OBM workflows taking longer to run, as the workflow to create IPMI settings doesn't start pollers, but pollers rely on these to be set correctly.  We should look at using the SKU pack options to set the OBMs instead of workflows, as that would speed this up ( I have tested it out a bit and seems to help.)

hohene:
+1 Using the changes above, I've tested out the install and stack initialization using the updates in the PR, ran successfully.

gpaulos:

anhou:
:+1:

test/fit_tests/common/fit_common.py
```diff
@@ -703,18 +707,27 @@ def apply_obm_settings_new():
                             nodestatus[node]['status'] = "succeeded"
                         if state in ["failed", "cancelled", "timeout"]:
                             nodestatus[node]['status'] = "pending"
-        if VERBOSITY > 4:
+                            # if the workflow left an invalid OBM, delete it
+                            result = rackhdapi("/api/2.0/nodes/" + node)
+                            if result['status'] == 200:
+                                if result['json']['obms']:
+                                    for ref in result['json']['obms']:
+                                        obmref = ref.get('ref')
```
brianparry: @gpaulos To me it looks like you are pulling the `obms` property out of a node.  I don't understand the `ref.get('ref') part.  What are you looking at here.  None of the obms should have a `ref` property.

test/fit_tests/common/fit_common.py
```diff
@@ -774,35 +787,71 @@ def apply_obm_settings():
...
-                failedlist.append(node)
+
+            if wfstatus != "failed":
+                # wait for OBM workflow to complete
+                counter = 0
+                for counter in range(0, 60):
+                    time.sleep(10)
+                    state_data = rackhdapi("/api/2.0/workflows/" + result['json']["instanceId"])
+                    if state_data['status'] == 200:
+                        if "_status" in state_data['json']:
```
brianparry: @gpaulos Recent versions of rackhd should return workflows with a `status` not `_status` property.  Going forward, we should probably be associating these tests with particular versions of the rackhd API or in some way make the tests version aware.

test/fit_tests/common/fit_common.py
```diff
@@ -363,7 +363,7 @@ def restful(url_command, rest_action='get', rest_payload=[], rest_timeout=None,
                                        verify=sslverify,
                                        )
         if rest_action == "binary-put":
-            rest_headers.update({"Content-Type": "application/octet-stream"})
+            rest_headers.update({"Content-Type": "application/x-www-form-urlencoded"})
```
gpaulos: Content-type has been a problem in the binary put and I don't know if this is a good universal solution for it. This seems to work for all of the cases that I have tested.

test/fit_tests/common/fit_common.py
gpaulos: Yes, these functions need to be version-aware. The problem here is that API 2.0 has had both status keys at different times.

gpaulos:
@brianparry @anhou I think this is ready for merge. If so, please do. If, not contact me.

PR #449, opened 2016-10-13
Add function test for update firmware with sku pack

nortonluo:
I add the tests for firmware update with sku packs. The skupacks with firmware are packed in an external link defined in global_config.json file.  To start a test, you need:
1. run command to update skupack"python run_tests.py -test ./tests/rackhd{API version}/test_rackhd{API version}_api_skupack.py:test_rackhd_api_skupack.test_api_{sku} -ora {rackhd ip}"
1. run command to test firmware update: "python run_tests.py -test ./tests/rackhd{API version}/test_rackhdkupack.py:test_rackhd{API version}/_api_skupack.test_api_{API version}_{sku and test name }-ora 192.168.128.10"

### DIFF
--- a/test/fit_tests/config/global_config.json
+++ b/test/fit_tests/config/global_config.json
@@ -75,7 +75,12 @@
       },
     "skupack": [
       "https://github.com/RackHD/on-skupack"
-      ]
+      ],
+    "skupacks":{"Quanta T41":"http://web.hwimo.lab.emc.com/autotest/skupack/quanta-t41/quanta-t41_t41.tar.gz",
+               "Quanta D51 2U":"http://web.hwimo.lab.emc.com/autotest/skupack/quanta-d51/quanta-d51-2u_d51.tar.gz",
+               "Rinjin KP":"http://web.hwimo.lab.emc.com/autotest/skupack/Rinjin-KP/Rinjin-KP.tgz",
+               "Rinjin TP":"http://web.hwimo.lab.emc.com/autotest/skupack/Rinjin-TP/Rinjin-TP.tgz",
+               "Hydra":"http://web.hwimo.lab.emc.com/autotest/skupack/Hydra/Hydra.tgz"}
   },
   "ports": {
     "_comment": "These are the northbound rest API port assignments",
--- a/test/fit_tests/config/global_config.json
+++ b/test/fit_tests/config/global_config.json
@@ -75,9 +75,7 @@
       },
     "skupack": [
       "https://github.com/RackHD/on-skupack"
-      ],
-    "skupacks":{"Quanta T41":"http://web.hwimo.lab.emc.com/autotest/skupack/quanta-t41/quanta-t41_t41.tar.gz",
-               "Quanta D51 2U":"http://web.hwimo.lab.emc.com/autotest/skupack/quanta-d51/quanta-d51-2u_d51.tar.gz"}
+      ]
   },
   "ports": {
     "_comment": "These are the northbound rest API port assignments",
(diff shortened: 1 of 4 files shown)


rackhdmirror:

amymullins:
@nortonluo , is there a need to introduce new tests for the 1.1 API at this point?  from the 10/5 Core Committer meeting minutes we discussed not introducing any new functionality for this API (once 2.0 released, 1.1 is deprecated and no additional feature enhancements)

Also, is there a way to mock out the actual "write" part of the FW update process?  Not sure if this test is intended to run as a PR quality gate or at some other regular frequency.  Definitely don't want to exhaust the lifetime of the programmable devices if this is intended to be a PR quality gate.

nortonluo:
@amymullins this is the test script we used for 2.1.0 release test, it is not a new test. This test is part of FIT test. I think the release 2.1.0 already passed these test before it get released.

I fully agree you concern about the lifetime. So I just set the attr parameter in the script to "regression=False, smoke=False". In this way, the smoke test and regression test will not automatically trigger this test case.

leoyjchang:
@nortonluo great work to add the FW upgrade test script in, which exactly matches what @iceiilin 's conclusion after postmortem 19 FW upgrade ODRs.  I can share you the details. In short, the top #1 action we identified is to add "FW upgrade" regression test in different HW platform/sku pack/fw packages. Not required in this PR, let's see how we can get "FW upgrade" regression test solution ready.

test/fit_tests/tests/rackhd11/test_rackhd11_api_skupack.py
```diff
@@ -18,7 +18,7 @@
 
 # Select test group here using @attr
 from nose.plugins.attrib import attr
-@attr(all=True, regression=True, smoke=True)
+@attr(all=True, regression=False, smoke=False)
```
hohene: @nortonluo I thin you need to change the all=True to False.  Also, how about another tag for firmware - we could use that to find all firmware tests?  firmware=True.

test/fit_tests/tests/rackhd20/test_rackhd20_api_skupack.py
```diff
@@ -18,7 +18,7 @@
 
 # Select test group here using @attr
 from nose.plugins.attrib import attr
-@attr(all=True, regression=True, smoke=True)
+@attr(all=True, regression=False, smoke=False)
```
hohene: same comment as above.

amymullins:
Thanks for the feedback @nortonluo and for adding in the flags for FW update.  

As for 1.1 api support, even though these tests were performed for the 2.1 release, this PR is introducing  new 1.1 based files and tests that will be executed and require maintenance.  I'm wondering if upcoming RackHD releases will be 2.0 only so we may not require any 1.1 testing.  Not a showstopper IMO and can add this to the agenda for next weeks CC meeting for further clarification.

amymullins:
Also please consider updates to platform naming conventions: https://hardware.corp.emc.com/display/CI/_Glossary

test/fit_tests/config/global_config.json
```diff
@@ -75,7 +75,12 @@
       },
     "skupack": [
       "https://github.com/RackHD/on-skupack"
-      ]
+      ],
+    "skupacks":{"Quanta T41":"http://web.hwimo.lab.emc.com/autotest/skupack/quanta-t41/quanta-t41_t41.tar.gz",
```
hohene: Let's move this into a separate config file - fw_config.json.  You are referencing an internal server for these tests, so it relies on running in a specific test environment.

johren:
Checking in on this PR.  It looks like the last activity was on 11/14.  Please update with further comment if you intend to move forward with this PR.  Otherwise, I will close on 2/18.

nortonluo:
The Pull request is outdated. It was based on old FIT framework. Will open another based on new FIT.

PR #466, opened 2016-10-25
 SUSE OS installation was failing when the node has too many Ethernet …

tannoa2:
…ports.

We added a wait statement to give a chance for the ports to initialize


tannoa2:
@RackHD/corecommitters @uppalk1 @geoff-reid @keedya @amymullins

PR #470, opened 2016-10-27
Add a function to determine if static IP/vlanid are set correctly

keedya:
This commit is adding simple testing to the network setup in the OS installation 
Jenkins network config: 
- eth2.10 : 192.168.100.1
- eth2 : 192.168.101.1


keedya:
@geoff-reid @tannoa2 @uppalk1

PR #475, opened 2016-11-01
enable rackhd on ubuntu 16.04

PengTian0:
1.  Remove the "--force". Because ubuntu 16.04 doesn't support the option.
2.  Enable rackhd start on startup
3.  Start on-xxx service only when the service is not running

@yyscamper @anhou @panpan0000

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -37,14 +37,14 @@ copy_RackHD_static_bins(){
 # Stop on-xxx 
 # Use Rackhd service to manager the status of on-xxx services
 #############################################
-stop_component_service(){
+stop_component_services(){
     for i in ${SERVICES[@]}; do
         service ${i} stop
     done
 }
 
 copy_RackHD_static_bins
-stop_component_service
+stop_component_services
 
 # Automatically added by dh_installinit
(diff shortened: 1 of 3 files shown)


debian/postinst
```diff
@@ -38,15 +34,22 @@ copy_RackHD_static_bins(){
...
+# Stop on-xxx 
+# Use Rackhd service to manager the status of on-xxx services
 #############################################
-start_service(){
-    db_get RackHD/WILLING_TO_SETUP_RACKHD
-    if [ "$RET" == "true" ];then
-        service rackhd restart
-    fi
-    db_stop
+stop_component_service(){
```
anhou: services？

debian/postinst
```diff
@@ -38,15 +34,22 @@ copy_RackHD_static_bins(){
...
+    done
 }
 
 copy_RackHD_static_bins
-start_service
+stop_component_services
+
+# Automatically added by dh_installinit
+if [ -x "/etc/init.d/rackhd" ]; then
+        update-rc.d rackhd defaults >/dev/null
```
yyscamper: Please refine the indent

PR #480, opened 2016-11-02
fix: rackhd services start before load config

PengTian0:
The rackhd services will start after installation.
But rackhd is installed before loading config file and the config file takes no effect on rackhd services.

This PR move rackhd installation after loading config file.

@heckj @panpan0000 @iceiilin @anhou @yyscamper


iceiilin:
:+1:

PR #484, opened 2016-11-04
 Auto Config primary IP address, when OVA is deployed 

panpan0000:
**Background**
Default RackHD OVA( same as Onrack) eth0(primary admin NIC) are configured as DHCP.
**How about  enviroment **without DHCP** on admin port (eth0) ?**
For many user scenario,   the OVA will be deployed into a network without DHCP against its eth0.
to enable this VM with static IP, before anyone can remote talk to this RackHD instance : then it will require manual operation: going to VMWare Console, and login into VM, modify the /etc/network/interface to re-configure that VM to static IP ....

A more automatic way is definitely needed.

---
**Summary of Implementation**
1. modify the OVF template file(XML) , to inject customized fields 
2. when deploying the OVA, pass IP/DNS/Gateway..etc parameter to ovftool or vCenter GUI
3. when the VM boot up , there's a script will be run during bootup (```config_IP_with_OVF_ENV.sh```), it will read the OVF ENV parameters via VMWare Tools, then config the /etc/network/interface

reference from 
-http://www.virtuallyghetto.com/2014/06/an-alternate-way-to-inject-ovf-properties-when-deploying-virtual-appliances-directly-onto-esxi.html
-http://www.v-front.de/2014/01/building-self-configuring-nested-esxi.html

---
**FAQ & Impacts**
- it supports both Ubuntu 14.04 & Ubuntu 16.04
- it doesn't impact vagrant box (```config_IP_with_OVF_ENV.sh``` will not take effect... ) 
- also it will not impact OVA in a DHCP environment or impact OVA without specific those parameters

---
**Usage** (NOTE : Only works for VCenter) 
   ** a. using ovftool as command line **
   1. using HWIMO-BUILD script, the OVA build is ready for this feature
   2. when deploying , below is an example of ovftool usage:
   ovftool --prop:adminIP=1.2.3.4 --prop:adminGateway=1.2.3.1 --prop:adminNetmask=255.255.255.0  --prop:adminDNS=10.1.1.1    --net:"ADMIN"="ADMIN vSwitch" --datastore="ds" --name="yourVMname"  rackhd_ovf_file_build.ovf  vi://user:pwd@vCenterIP

  ** b. using vCenter GUI **
image

---

**What's been Tested**

original  /etc/network/interface ( inside the OVA) before this PR:
image

1. Created OVA for 14.04 & 16.04

2. using ```ovftool``` to deploy the OVA with specified IP setting  , then when boot up , the  /etc/network/interface as below:
image

3.  using ```ovftool``` to deploy the OVA **without** specified IP setting  , when boot up , the VM's /etc/network/interface remains no change as original content.

---
**Who to Review**
@RackHD/corecommitters   @changev

### DIFF
--- a/packer/HWIMO-BUILD
+++ b/packer/HWIMO-BUILD
@@ -107,7 +107,6 @@ BASENAME=${VM_NAME}
 OVA="${BASENAME}.ova"
 ovftool $SIGN_ARGS -o ${VMDIR}/${VM_NAME}.vmx $OVA
 
-
 if [ $? != 0 ]; then
     echo "ovftool exec failed.. exit"
     exit 4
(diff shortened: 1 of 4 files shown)


changev:
GGGGGGreat feature!!!

:+1:

leoyjchang:
I like this feature and especially appreciated the comprehensive content (what/how/test result/) provided to help the review easier. This is a "role model" of PR submission. +1

PR #485, opened 2016-11-07
OS Install Regression test enhancements for Windows OS

uppalk1:
- Update max payload for windows server 2012 installation


uppalk1:
Depends RackHD/on-http#526

PR #491, opened 2016-11-14
Updated rackHD installer scripts

gpaulos:
This PR is to update the RackHD installer scripts as part of Phase 1 of RackHD Test Framework:
https://groups.google.com/forum/?hl=en#!topic/rackhd/cscQ2T20Uxs

Files updated are:
test/fit_tests/deploy/rackhd_source_install.py  - installs RackHD from source using rackhd_local.yml 
test/fit_tests/deploy/rackhd_package_install.py  - installs RackHD from packages using rackhd_package.yml 
test/fit_tests/config/global_config.json - contains repo and branch selections for source installer

The new installers are intended meet the following objectives:
- run either remote or local on management server (default)
- run from ESXi-based or Virtualbox-based management server VMs
- ability to run behind proxy
- run from Ubuntu 16 or Ubuntu 14 (see exception below)
- source installed from selectable repos and branches, default to RackHD/master
- install all RackHD package requirements, config files, network config
- start RackHD services without reboot
- check API to verify operational

The installers perform the following functions:
 - loads prerequisite packages git, ansible, etc.
 - downloads RackHD source or packages to management server 
 - installs using Ansible playbook
 - set up networking
 - load configuration files
 - startup and verify operations

Notes: 

Source installer fails on Ubuntu 14 due to a recent change in rackhd_local.yml than cause it to hang. If Ubuntu 14 is still supported then I will file this as a bug. Package installer works with both 14 and 16.

Source repos/branches installed via rackhd_source_install.py can be selected by changing the global_config.json file under the 'repos':'install' section. The configuration file format is likely to be changed when the Test Configuration Management story is completed. 

Updated source installer is currently running this Jenkins smoke test job in MN lab:
https://roebling.hwimo.lab.emc.com/jenkins/job/Running-Tests/job/RackHD-FIT-Smoke-Test/

### DIFF
--- a/test/fit_tests/deploy/rackhd_package_install.py
+++ b/test/fit_tests/deploy/rackhd_package_install.py
@@ -4,7 +4,20 @@ Copyright 2016, EMC, Inc.
 Author(s):
 George Paulos
 
-This script installs RackHD from BinTray packages onto blank OS.
+This script installs RackHD from BIntray packages onto blank Ubuntu 14 or 16 OS via Ansible installer.
+If test bed is behind proxy wall, make sure to enter proxy URL in config/global_config.json.
+This script performs the following functions:
+    - loads prerequisite packages git, ansible, etc.
+    - downloads RackHD source to management server
+    - installs using rackhd_package.yml playbook
+    - set up networking
+    - load configuration files
+    - startup and verify operations
+
+usage:
+    python run_tests.py -ova <ip or host> -test deploy/rackhd_package_install.py
+    or
+    python run_tests.py -stack <stack ID> -test deploy/rackhd_package_install.py
 '''
 
 import os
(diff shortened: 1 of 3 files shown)


test/fit_tests/deploy/rackhd_package_install.py
```diff
@@ -168,57 +228,30 @@ def test03_install_rackhd_config_files(self):
...
-            if shell_data['exitcode'] == 0:
-                break
+    def test06_startup(self):
+        print "Restart services."
+        self.assertEqual(fit_common.remote_shell("service isc-dhcp-server restart")['exitcode'], 0, "on-http failure.")
+        self.assertEqual(fit_common.remote_shell("service on-http restart")['exitcode'], 0, "on-http failure.")
+        self.assertEqual(fit_common.remote_shell("service on-dhcp-proxy restart")['exitcode'], 0, "on-http failure.")
+        self.assertEqual(fit_common.remote_shell("service on-syslog restart")['exitcode'], 0, "on-http failure.")
+        self.assertEqual(fit_common.remote_shell("service on-taskgraph restart")['exitcode'], 0, "on-http failure.")
+        self.assertEqual(fit_common.remote_shell("service on-tftp restart")['exitcode'], 0, "on-http failure.")
```
johren: cut-and-paste issue on the failure messages.  All of them say "on-http failure".

test/fit_tests/deploy/rackhd_package_install.py
```diff
@@ -168,57 +228,30 @@ def test03_install_rackhd_config_files(self):
...
-                fit_common.time.sleep(5)
-        self.assertEqual(shell_data['exitcode'], 0, "Shell test failed after appliance reboot")
-        fit_common.time.sleep(10)
+                break
         self.assertEqual(fit_common.rackhdapi("/api/2.0/config")['status'], 200, "Unable to contact RackHD.")
 
 if __name__ == '__main__':
-    fit_common.unittest.main()
\ No newline at end of file
+    fit_common.unittest.main()
```
johren: Not a good idea to import unittest from fit_common unless you are overriding it.  This should be imported within this module.

test/fit_tests/deploy/rackhd_source_install.py
```diff
@@ -39,21 +75,18 @@ def test01_install_rackhd_dependencies(self):
         self.assertEqual(fit_common.remote_shell(ENVVARS + "apt-get -y update")['exitcode'], 0, "update failure.")
         self.assertEqual(fit_common.remote_shell(ENVVARS + "apt-get -y dist-upgrade")['exitcode'], 0, "upgrade failure.")
         self.assertEqual(fit_common.remote_shell("git config --global http.sslverify false")['exitcode'], 0, "Git config failure.")
```
johren: Same comments regarding apt-get update and apt-get dist-upgrade as above.

test/fit_tests/deploy/rackhd_source_install.py
```diff
@@ -39,21 +75,18 @@ def test01_install_rackhd_dependencies(self):
...
-                                'wget --quiet -O - https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add -;'
-                                'echo "deb https://deb.nodesource.com/node_4.x trusty main" | tee /etc/apt/sources.list.d/nodesource.list;'
-                                'echo "deb-src https://deb.nodesource.com/node_4.x trusty main" | tee -a /etc/apt/sources.list.d/nodesource.list;')
-        fit_common.remote_shell(ENVVARS + 'apt-get -y update;apt-get -y install nodejs;')
-        fit_common.remote_shell(ENVVARS + 'apt-get -y install npm;npm config set https-proxy ' + fit_common.GLOBAL_CONFIG['repos']['proxy'])
-        self.assertEqual(fit_common.remote_shell('node -v')['exitcode'], 0, "Node install failed..")
-
-    def test03_clone_rackhd_source(self):
+        # install Ansible
+        self.assertEqual(fit_common.remote_shell(ENVVARS + "cd ~;apt-get -y install ansible")['exitcode'], 0, "Ansible Install failure.")
```
johren: Same comment regarding unnecessary cd ~

test/fit_tests/deploy/rackhd_source_install.py
```diff
@@ -39,21 +75,18 @@ def test01_install_rackhd_dependencies(self):
...
+        self.assertEqual(fit_common.remote_shell(ENVVARS + "apt-get -y update")['exitcode'], 0, "Ansible Update failure.")
```
johren: apt-get -y update is not necessary here.

test/fit_tests/deploy/rackhd_source_install.py
```diff
@@ -64,38 +97,92 @@ def test03_clone_rackhd_source(self):
...
         self.assertEqual(fit_common.remote_shell(ENVVARS +
                                                  "cd ~/rackhd/packer/ansible/;"
-                                                 "ansible-playbook -i 'local,' -c local rackhd_package.yml",
-                                                 timeout=600,
-                                                 )['exitcode'], 0, "Install failure.")
+                                                 "ansible-playbook -i 'local,' -c local rackhd_local.yml",
+                                                 timeout=1200,
+                                                 )['exitcode'], 0, "RackHD Install failure.")
+
+    def test04_install_network_config(self):
```
johren: Can this common code be pulled into fit_common?  It appears to be duplicated in both rackhd_source_install.py and rackhd_package_install.py.

test/fit_tests/deploy/rackhd_source_install.py
```diff
@@ -112,12 +199,20 @@ def test05_install_rackhd_config_files(self):
                         '}\n'
                          )
         dhcp_conf.close()
+        # copy file to ORA
+        fit_common.scp_file_to_ora('dhcpd.conf')
+        self.assertEqual(fit_common.remote_shell('cp dhcpd.conf /etc/dhcp/')['exitcode'], 0, "DHCP Config failure.")
+        os.remove('dhcpd.conf')
+
+    def test05_install_rackhd_config_files(self):
```
johren: Same here.  Can this be pulled into fit_common?

test/fit_tests/deploy/rackhd_source_install.py
```diff
@@ -184,56 +279,30 @@ def test05_install_rackhd_config_files(self):
...
-        print "**** Rebooting appliance..."
-        fit_common.remote_shell("shutdown -r now")
-        print "**** Waiting for login..."
-        fit_common.countdown(30)
-        shell_data = 0
-        for dummy in range(0, 30):
-            shell_data = fit_common.remote_shell("pwd")
-            if shell_data['exitcode'] == 0:
-                break
+    def test06_startup(self):
```
johren: Same here. Can this be pulled into fit_common?

test/fit_tests/deploy/rackhd_source_install.py
```diff
@@ -184,56 +279,30 @@ def test05_install_rackhd_config_files(self):
...
-                fit_common.time.sleep(5)
-        self.assertEqual(shell_data['exitcode'], 0, "Shell test failed after appliance reboot")
-        fit_common.time.sleep(10)
+                break
         self.assertEqual(fit_common.rackhdapi("/api/2.0/config")['status'], 200, "Unable to contact RackHD.")
 
 if __name__ == '__main__':
-    fit_common.unittest.main()
\ No newline at end of file
+    fit_common.unittest.main()
```
johren: Same comment here.  This should just be unittest.main().   Reader assumes this is overridden.

test/fit_tests/deploy/rackhd_package_install.py
```diff
@@ -14,72 +27,111 @@
...
+        sudoersproxy = open("sudoersproxy", 'w')
+        sudoersproxy.write('Defaults env_keep="HOME no_proxy http_proxy https_proxy"\n')
+        sudoersproxy.close()
+        fit_common.remote_shell('pwd')
+        fit_common.scp_file_to_ora("sudoersproxy")
+        self.assertEqual(fit_common.remote_shell('cp sudoersproxy /etc/sudoers.d/'
+                                                  )['exitcode'], 0, "sudoersproxy config failure.")
+        os.remove('sudoersproxy')
+        # install git
+        self.assertEqual(fit_common.remote_shell(ENVVARS + "apt-get -y install git")['exitcode'], 0, "Git install failure.")
```
gpaulos: Will do

test/fit_tests/deploy/rackhd_package_install.py
```diff
@@ -14,72 +27,111 @@
...
+        self.assertEqual(fit_common.remote_shell(ENVVARS + "apt-get -y update")['exitcode'], 0, "update failure.")
```
gpaulos: This is the sequence from the RackHD install documentation

test/fit_tests/deploy/rackhd_package_install.py
```diff
@@ -14,72 +27,111 @@
...
+        self.assertEqual(fit_common.remote_shell(ENVVARS + "apt-get -y dist-upgrade")['exitcode'], 0, "upgrade failure.")
```
gpaulos: This is the sequence from the RackHD install documentation. I'd love to get rid of it.

test/fit_tests/deploy/rackhd_source_install.py
gpaulos: Yep its duplicated, but only in the installers. Rather not put it in common.

test/fit_tests/deploy/rackhd_source_install.py
gpaulos: OK

test/fit_tests/deploy/rackhd_package_install.py
gpaulos: This is the same throughout the test code. If we do that here, then lets do it for all.

gpaulos: This is the same throughout the test code. If we do that here, then lets do it for all in a different PR.

johren: Doing it in this PR should not affect the other files.  I will do the work of getting it done in the rest of the files.  If you can do me a favor and do it in this file, we will not have to wait on each other to get our PRs in.

gpaulos:
Updated installer scripts to remove apt-get updates and change ENVVARS to PROXYVARS

diontje:
+1 Nice. 
Looking over the script, I do think that there are places that could use some waits. For example, you up your NIC and immediately ping it. But if your test runs are good, I would not bother.

gpaulos:
Torry, that's a sharp observation. Actually, the ping itself has a 5 second wait which seems to be adequate. It's been running reliably on our Jenkins for over a week.

hohene:
Pulled the branch and tested out the installation on two physical stacks, using config updates for MN lab.  Worked well on both.  +1,  after response to Jeanne's comment on the fit_common.unittest updates above.

gpaulos:

test/fit_tests/deploy/rackhd_source_install.py
gpaulos: The original intention of having the sole instance of unittest in test_common was to extract extended test data from the active unittest instance for extended reporting. This is how autotest retrieved data. We need to do some experiments to see if we can extract the same data from the test script instance of unittest.

johren: Okay, we'll deal with this in a separate PR.

PR #501, opened 2016-11-18
Include firewallDisable in options 

uppalk1:
.....to let the ping function test the new static IP


uppalk1:
depends on RackHD/on-http#537

uppalk1:
depends on RackHD/on-tasks#375

uppalk1:
@RackHD/corecommitters

PR #502, opened 2016-11-18
RAC-264 Add granular log-levels to logging infra for testing

stuart-stanley:
RAC-265 Retrofit FIT w/ new logging handles

This changeset is the first in a series of stories to add
stream-based watcher logic to the testing infrastructure. The focus
here was to:
* Get a set of handles available for different major areas of the code (actual tests vs test infrastructure and logic (run) flow vs raw data)
* Extend the very limited pallet of syslog based levels to allow levels within the syslog'ish constraints
* Handle the "I'm writing tests and don't want to fill up the world" use-case as well as build-time
* Handle mproc and greenlets
* Add some super-basic self-tests. As the stream-monitors come in later stories, these will be enhanced

In other words, this is very much a WIP with more to come.

NOTE: this does not change any of the main code-paths. There are other stories to start integrating this in.

@gpaulos, @hohene, @tfuehrer for code commentary/observation
@RackHD/corecommitters  for permission if one or more of the above ok it.


stuart-stanley:
Overrun with new stories while waiting for review. Code consumed by next steps and will be in that.

PR #504, opened 2016-11-22
Changing device name to pass MAC address 

uppalk1:
Device name doesn't stay consistent in Windows OS on Jenkins slave 5. So changing the name to MAC address.


uppalk1:
@geoff-reid @tannoa2 @keedya @brianparry

PR #505, opened 2016-11-23
Fix isc-dhcp-server fail to start in 16.04 OVA

panpan0000:
because /etc/hostname is rackhd-demo (by scripts/dep.sh)

but in /etc/hosts, with install-command-line host=RackHD, it will be like
```
127.0.0.1 localhost
127.1.0.1 RackHD
```

so ```rackhd-demo``` can't be resolved.


cgx027:
I would think ```rackhd-ova``` or ```rackhd-host``` may be a better name for both /etc/hostname and /etc/hosts. Using word ```demo``` as a name makes it looks like a more like a demo/dev env and stop our customers from using it in more formal env.

cgx027:
+0 and non-blocking comment anyway.

panpan0000:
HOLD ON MERGE!!!
I found some new problem..

panpan0000:
create new https://github.com/RackHD/RackHD/pull/508 to replace this one.

PR #506, opened 2016-11-24
WIP: Add tests for sel poller alerts

tannoa2:
-Generate and post the skupacks that create the alert defenition
-Validate that the selEntires pollers are created fo the compute
node after the skupack si posted
-validate the basic alert functionality of the poller by listining on AMQP
-validate the RackHD decodes the raw sel read correctly
("Event Type Code" and "Sensor Type Code")
-Validate the sel poller alert can handle the overflow option in the sel
-validate that the poller can handle when the sel is cleared
-validate that poller can digest the data of the whole sel log at once and entries are dropped


tannoa2:
test this please

tannoa2:
@RackHD/veyron

PR #507, opened 2016-11-24
Create rsyslog configure file example

pengz1:
Create rsyslog configure file example.
A new feature is added to OS installation to enable node to send OS installation syslog to RackHD server. This PR created a rsyslog configure that enables RackHD server rsyslog to received syslog from nodes and store logs in /var/log/rackhd/ with <node_ip>_<date> format.


pengz1:
@RackHD/corecommitters @cgx027 @iceiilin

leoyjchang:
Looks good to me as it is an example to user and is functional. +1

PR #510, opened 2016-11-28
change absolute path in rackhd-pm2-config.yml to relative path

changev:
```rackhd-pm2-config.yml``` use absolute path of "/home/vagrant/....." 
This will lead failure when user install rackhd according to http://rackhd.readthedocs.io/en/latest/rackhd/ubuntu_source_installation.html.

@RackHD/corecommitters


cgx027:
:+1:

PR #511, opened 2016-11-28
Always map the repo dir.

### LINKED ISSUE #499: Vagrantfile in example directory fails to map repos in specific cases
In the Vagrantfile located in the example directory there is special handling when the `WORKSPACE` and `REPO` variables are set.

In the general case the intent appears to be to map the repo specified in `REPO` through to the Vagrant VM. If `REPO` is `on-tasks` or `on-core` then care is taken to map those into the node_modules directory of the various projects that rely on them... and then they are never mapped into the top-level workspace.

Patch to fix this attached.
[workspace-repo-fix.txt](https://github.com/RackHD/RackHD/files/596264/workspace-repo-fix.txt)

jeamland:
Currently if the repo dir you specify via the environment is on-core on on-tasks it doesn't get mapped due to the mapping being in the else clause of the special-casing that maps them into the node_modules directories of the other projects.

This fixes that, allowing easier development of on-core and on-tasks.

Fixes #499


DavidjohnBlodgett:
👍

leoyjchang:
Thanks for the contribution, @jeamland. :+1. @RackHD/corecommitters

PR #512, opened 2016-11-29
added base version of docker-compose config file

zhangy42:
This is a configure file that only let docker-compose create 6 basic services (in docker containers) for rackhd. 

It is used in the situation when we need to deploy only basic services on customized OS (with daemons installed according to doc here https://github.com/zhangy42/RackHD/wiki/Deploying-RackHD-on-CentOS).

Original usage case is defined here https://www.pivotaltracker.com/story/show/131214777

@RackHD/corecommitters 
@rolandpoulter


docker/docker-compose-base.yml
```diff
@@ -0,0 +1,82 @@
+# Copyright 2016, EMC, Inc.
+
+version: '2'
+
+volumes:
+  dhcp-leases:
+    external: false
+
+services:
```
cgx027: Missing Rabbitmq?

docker/docker-compose-base.yml
```diff
@@ -0,0 +1,82 @@
...
+    depends_on:
+      - mongo
+    image: rackhd/on-dhcp-proxy:latest
+    network_mode: "host"
+    privileged: true
+    volumes:
+      - "dhcp-leases:/var/lib/dhcp"
+      - "./monorail:/opt/monorail"
+
+  http: # 9090, 9080
```
cgx027: I would think this should also depends on core service (https://github.com/RackHD/RackHD/blob/master/docker/docker-compose.yml#L47) and tasks(https://github.com/RackHD/RackHD/blob/master/docker/docker-compose.yml#L52) service, but I believe this file is tested to prove that core and tasks services are not needed.

docker/docker-compose-base.yml
cgx027: I would guess the requirement only expect RackHD-only services, but that does not seem explain why mongo service is presented here.

zhangy42:
@cgx027 You are correct, the purpose of this file is only keep the part that is modified by on-* services. For those are static, it assumes that the OS will provide.

The original reason for making this file, as mentioned in the ticket, is to make the docked rackhd to run on Centos, and I find that rabbit-mq and isc-dhcp service does not run stably with the OS, so I moved them to the prerequisite of this ticket (as is mentioned in the document here https://github.com/zhangy42/RackHD/wiki/Deploying-RackHD-on-CentOS).

cgx027:
@zhangy42 sounds good. maybe we can have some follow up task created in Pivotal Tracker on following up the centOS stability issue with rabbit-mq & isc-dhcp-service. 

Also, if this file is supposed to only run at centOS envirment, maybe we rename the file to ```docker-compose-centos.yml```. Just my 2 cents. :+1:  anyway.

docker/docker-compose-base.yml
rolandpoulter: RabbitMQ was not working in CentOS on Docker, so that is run outside of Docker.

docker/docker-compose-base.yml
rolandpoulter: Technically you would want to build `on-core` and `on-tasks` before building the other services. However I don't think it is important to include them in this docker-compose file.

rolandpoulter:
I agree, renaming to `docker-compose-centos.yaml` is more appropriate.
👍

docker/docker-compose-base.yml
cgx027: Thanks @rolandpoulter for adding this.

zhangy42:
How can my changes are related to 2.0 api?

zhangy42:
Jenkins: test this please

tldavies:
👍

pscharla:
@brianparry @yyscamper @anhou : Need CC review/approval for this PR. Thanks!!

pscharla:
@rolandpoulter : Now that you have merge capabilities, can you take care of this one?  Looks like it now positive reviews

PR #514, opened 2016-11-30
add longer timeout to test_link_up for OS installs

geoff-reid:
@keedya @tannoa2 @uppalk1 @brianparry


geoff-reid:
test this please

PR #515, opened 2016-11-30
Update POST /api/2.0/nodes related tests to provide now required 'type' field.

jimturnquist:
This fixes three test failures related to /api/2.0/nodes requiring a type field in its POST block.

@RackHD/corecommitters
@gpaulos @stuart-stanley


stuart-stanley:
+1

PR #508, opened 2016-11-24
Fix 16.04 OVA:  isc-dhcp-server can't start

panpan0000:
This is re-work of PR https://github.com/RackHD/RackHD/pull/505

------

**Background:**
   in OVA, the NIC1 is ens33, and in Vagrant , the NIC1 is enp0s8.
   isc-dhcp-server fails to start in 16.04 OVA (because no ens33  in /etc/default/isc-dhcp-server )
   moreover, vagrant 16.04 can't work ,because scripts/addnetif.sh & script/firstuser.sh doesn't consider enp0s8

------

**Root Cause:**
   same as above

------
***Implementation***
    Smartly retrieving the Control Port (Secondary NIC) Name:  leverage "ip addr" command, assume the secondary NIC are listed as index=3 in "ip addr" output.

**Some other fixes **
    1. when /etc/hostname updated, the /etc/hosts remains unchanged (its hostname came from DHCP server). it bring confusion of warning : "unable to resolve host rackhd",
    2. change ```rackhd-demo``` hostname to ```rackhd``` (per @cgx027's comment in PR https://github.com/RackHD/RackHD/pull/505)
    3.  remove ansible/roles/isc-dhcp-server/files/systemd-control.cfg. because it's duplicated with packer/scripts/addnetif.sh, and moreover, it's hardcoded as enp0s8

------

**FAQ**

**Q**: Will there be side-effect for "INTERFACES="ens32 ens192" ?
**A**: Nope. it will try to enable DHCP on both ens32 and ens192. in most case, they will NOT both exist.
but it will support both two NIC naming convention for e1000 & vmnet3. moreover, isc-dhcp-server will not fail even ens192 doesn't exist.

**Q**: why not change /etc/hosts together with hostname inside scripts/dep.sh ?
**A**: the /etc/hosts change requires reboot.  misalignment of run-time hostname and /etc/hosts will make lots of ```sudo``` operation very **slow** ( refer :http://askubuntu.com/questions/322514/terminal-command-with-sudo-takes-a-long-time). like firstuser.sh , the services start/stop will be very very slow. so move the /etc/hostname update into cleanup.sh

------

**Test Done**

After this PR fix, the new 16.04 OVA:
/etc/hostname will be rackhd
image
/etc/hosts will contain 127.0.1.1 rackhd
image

the isc-dhcp-server service runs normal.
image

-----

Thank you 
@yyscamper  @anhou  @cgx027  @changev 

---- vagrant box 16.04 Test result:

[...]

PR #516, opened 2016-12-01
Bug fix: move library script to VM's /tmp/

panpan0000:
this is a hot fix https://github.com/RackHD/RackHD/pull/508/

because any script in packer ( either in packer provision or ansible) runs inside the created VM,
so the shell library file should be copied to VM first, before invoking it.

@changev @cgx027 @anhou @yyscamper


changev:
:+1: fixed an impatient critical issue!

PR #518, opened 2016-12-02
adjust the order of packer provisioning: config eth1 before ansible

panpan0000:
What's up:

1. it's a bug  fix for 14.04 packer script. the ```get_nic_name_by_index.sh``` file copy should happen before ```addnetif.sh```
2. change the admin (eth1) configuration before ansible playbooks( install rackhd..etc)

Test Done:
successful build a 14.04 OVA , and verify RackHD service running well for that OVA

@changev  @yyscamper  @anhou


changev:
:+1:

PR #521, opened 2016-12-05
RAC-800: Make the Ansible can run without Packer ( option #1)

panpan0000:
It's option 1, option 2 is https://github.com/RackHD/RackHD/pull/522

**Background**
there's a common script(get_nic_name_by_index.sh) to be invoked by both ```packer/scripts/``` and ```packer/ansible/````
 simply putting the common-script in either place , will make the other can't access this common-scripts.
Example, packer will only copy content under "packer/ansbile" to VM's staging dir(/tmp/packer-provisioner-ansible-local), so the common script will never be found by ansible.

**Previous Fix Attempts**
- I used packer provision scripts(running before ansible) to copy the common-script to VM's filesystem. but it makes ansible has to be depend on the packer , and can't run alone
- I tried to leverage some packer parameter to workaround and upload the common script to VM's filesystem , like ```playbook_dir```. but still not working .

**This PR**
I have to give up the code reuse , and copy the code again here..

@cgx027  @yyscamper @anhou @changev

### DIFF
--- a/packer/ansible/roles/isc-dhcp-server/files/config_isc-dhcp-server.sh
+++ b/packer/ansible/roles/isc-dhcp-server/files/config_isc-dhcp-server.sh
@@ -1,9 +1,84 @@
 #!/bin/bash
 set -e
 
+############################################################
+##     Why Duplication code below ?
+##
+## Below two functions come from packer/scripts/common/get_nic_name_by_index.sh
+## becauce when packer execute ansible ,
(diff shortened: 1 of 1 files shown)


anhou:
:+1:

cgx027:
Option 1 looks better. :+1:

yyscamper:
After discussion, we agree on the option #1, so please close the option #2.

PR #526, opened 2016-12-08
RAC 3712 ansible source installer

gpaulos:
RAC-3712: This PR creates a new Ansible installer playbook that installs RackHD from the local source tree. This is needed to support installation from a selected PR or development branch from FIT deployment. The script rackhd_source_install.py creates the source tree from a repo list, the playbook rackhd_source.yml then uses that source instance to install RackHD.

@RackHD/corecommitters 
@hohene 
@johren 
@diontje

### DIFF
--- a/packer/ansible/rackhd_source.yml
+++ b/packer/ansible/rackhd_source.yml
@@ -1,0 +1,25 @@
+---
+- name: Development Server
+  hosts: local
+  sudo: no
+  roles:
+    - apt
+    - build
+    - git
+    - rabbitmq
+    - mongodb
+    - node
+    - isc-dhcp-server
+    - monorail
+    - snmptool
+    - ipmitool
+    - foreman
+    - pm2
+    - localsource
+    - images
+    - swagger
+    - integrationtest
+    - postgresql
+  vars:
+# branch of https://github.com/rackhd/rackhd to be used
+    - branch: "master"
--- a/packer/ansible/roles/localsource/tasks/main.yml
+++ b/packer/ansible/roles/localsource/tasks/main.yml
@@ -7,7 +7,7 @@
   sudo: yes
 
 - name: Install RackHD from local source
-  copy: src="../../{{ item }}" dest="{{ ansible_env.HOME }}/src/{{ item }}"
+  copy: src="../../{{ item }}" dest="{{ ansible_env.HOME }}/src/"
   with_items:
     - on-core
     - on-tasks
@@ -43,6 +43,10 @@
   shell: "npm run apidoc"
   args:
     chdir: "{{ ansible_env.HOME }}/src/on-http"
+
+- file:
+    path: "{{ ansible_env.HOME }}/src/on-http/install-task-doc.sh"
+    mode: 0777
 
 - name: Generate hosted task documentation
   shell: "npm run taskdoc"
--- a/packer/ansible/roles/localsource/tasks/main.yml
+++ b/packer/ansible/roles/localsource/tasks/main.yml
@@ -1,0 +1,45 @@
+---
+- name: install dev libraries for NPM installs
+  apt: pkg={{ item }} state=installed
+  with_items:
+    - libkrb5-dev
+    - unzip
+  sudo: yes
+
+- name: Install RackHD from local source
+  copy: ../../* ansible_env.HOME
+
+- name: Npm install Repos
+  npm: path="{{ ansible_env.HOME }}/src/{{ item }}"
+       production=yes
+  with_items:
+    - on-syslog
+    - on-tftp
+    - on-dhcp-proxy
+    - on-taskgraph
+    - on-http
+    - on-wss
+
+- name: Make common static directory
+  file: path="{{ ansible_env.HOME }}/src/on-http/static/http/common" state=directory
+
+- name: Npm install apidoc
+  shell: "npm install apidoc"
+  args:
+    chdir: "{{ ansible_env.HOME }}/src/on-http"
+
+- name: Render 1.1 API docs for local reference
+  shell: "npm run apidoc"
+  args:
+    chdir: "{{ ansible_env.HOME }}/src/on-http"
+
+- name: Generate hosted task documentation
+  shell: "npm run taskdoc"
+  args:
+    chdir: "{{ ansible_env.HOME }}/src/on-http"
+
+- name: set HTTP static directory for RackHD
+  set_fact: http_static_directory="{{ ansible_env.HOME }}/src/on-http/static/http"
+
+- name: set TFTP static directory for RackHD
+  set_fact: tftp_static_directory="{{ ansible_env.HOME }}/src/on-tftp/static/tftp"
--- a/packer/ansible/roles/localsource/tasks/main.yml
+++ b/packer/ansible/roles/localsource/tasks/main.yml
@@ -7,7 +7,18 @@
   sudo: yes
 
 - name: Install RackHD from local source
-  copy: ../../* ansible_env.HOME
+  copy: src="../../{{ item }}" dest="{{ ansible_env.HOME }}/src/{{ item }}"
+  with_items:
+    - on-core
+    - on-tasks
+    - on-http
+    - on-tftp
+    - on-dhcp-proxy
+    - on-taskgraph
+    - on-syslog
+    - on-wss
+    - on-tools
+    - on-imagebuilder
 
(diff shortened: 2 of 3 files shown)


yyscamper:
@panpan0000 @PengTian0 : Could you please take a look at this PR?

packer/ansible/roles/localsource/tasks/main.yml
```diff
@@ -0,0 +1,60 @@
...
+    - on-core
+    - on-tasks
+    - on-http
+    - on-tftp
+    - on-dhcp-proxy
+    - on-taskgraph
+    - on-syslog
+    - on-wss
+    - on-tools
+    - on-imagebuilder
```
yyscamper: on-tools, on-imagebuilder and on-wss is not required for RackHD installation.

packer/ansible/rackhd_source.yml
```diff
@@ -0,0 +1,25 @@
...
+  sudo: no
+  roles:
+    - apt
+    - build
+    - git
+    - rabbitmq
+    - mongodb
+    - node
+    - isc-dhcp-server
+    - monorail
```
yyscamper: What's this?

packer/ansible/roles/localsource/tasks/main.yml
gpaulos: This list of repos duplicates what is downloaded in the remoterepos playbook:
https://github.com/RackHD/RackHD/blob/master/packer/ansible/roles/remoterepos/tasks/main.yml

packer/ansible/rackhd_source.yml
gpaulos: This playbook was modeled on the rackhd_local.yml playbook:
https://github.com/RackHD/RackHD/blob/master/packer/ansible/rackhd_local.yml

yyscamper: I still don't have the history, if @heckj happens to see this comment, it will be very appreciate to give some info why include the `monorail` here and what is it?

Or whether @panpan0000 and @PengTian0 know some detail?

gpaulos: I don't know what all of this actually is. I do know that rackhd_local.yml works and is the model that I used for the source installer.

johren: Based on packer/ansible/roles/monorail/tasks/main.yml , it appears that the monorail role sets up the monorail.json config file.

packer/ansible/roles/localsource/tasks/main.yml
johren: +1 to Felix's comment

packer/ansible/roles/localsource/tasks/main.yml
```diff
@@ -0,0 +1,60 @@
...
+
+- name: Npm install Repos
+  npm: path="{{ ansible_env.HOME }}/src/{{ item }}"
+       production=yes
```
johren: Just a heads up that this may not get you what you expect.   For example, when you do an npm install in on-taskgraph, it is going to install the on-tasks and on-core modules from github rather than from local source unless you modify the package.json.

"dependencies": {
    "di": "git+https://github.com/RackHD/di.js.git",
    "dot-object": "^0.6.0",
    "lodash": "^3.10.1",
    "lru-cache": "^2.5.0",
    "node-uuid": "^1.4.2",
    "on-core": "git+https://github.com/RackHD/on-core.git",
    "on-tasks": "git+https://github.com/RackHD/on-tasks.git",
    "rx": "^2.3.22"
  },

panpan0000:
Hold on. My comments is majorly about the design of **input** to FIT as a box.
I can understand the purpose, FIT would like to act as PR Gate Test, so you wanted to build + test upon code already cloned .
But should FIT worry about  "clone + build" or "build" , or let Jenkins to do it ?
I don't want FIT to take too much bundle  as a test framework .
previously, CIT(Proboscis)  was run based on RackHD code all already cloned and build, and those clone and build are taken care by upstream Jenkins job.
I'm still thinking about how to integrate FIT into PR Gate.The easiest way is directly replace CIT in http://rackhdci.lss.emc.com/job/Templates/job/smoke-test/ with FIT.

packer/ansible/roles/localsource/tasks/main.yml
gpaulos: Well. That's a problem. This needs to be fixed. Let me ponder this.

gpaulos: Well. That's a problem. This needs to be fixed. Let me ponder this.

I see that the default playbooks download on-core and on-tasks twice.  Is there a reason for this or is this unintentional?

gpaulos:
@panpan0000 These scripts need to also be usable by the community who don't have Jenkins servers and want to run an install and/or test locally onto a vagrant or other environment. There is a world outside of Jenkins.

packer/ansible/roles/localsource/tasks/main.yml
johren: There may be a reason for this.  Can you point me at the specific playbooks you are referring to?  I'll take a look.

panpan0000:
@gpaulos , it sounds reasonable.

I'm still struggling:  Is it necessary FIT do everything "smartly" but "heavily",or using separated steps to do the deployment and testing ?

back to this PR:

 is it possible to **re-use current existing ansible** and extend its capability ? too much duplication between two scripts ...
I meant (https://github.com/RackHD/RackHD/blob/master/packer/ansible/roles/remoterepos/tasks/main.yml)
because the only difference is the **local** V.S. **remote**. maybe you update original scripts with some ansible variable or conditional execution , to tell between them ?

-----

original code's logic (ansible/roles/remoterepos) :

1#git clone RackHD/RackHD  to {{ ansible_env.HOME }}/src
2#git fetch 
3#git reset --hard {{branch}}  
4#npm install......etc

---------

how about extend it to with a variable ```local_build_folder``` or sth.. :
1# Do above clone/fetch/reset only  ```when: local_build_folder is undefined```   (refer to : http://docs.ansible.com/ansible/playbooks_conditionals.html). the ```local_build_folder``` is an ansible variable like ```{{branch}}``` in original script
2. if  ```when: local_build_folder is defined```, then the either copy  ```local_build_folder``` to ```{{ ansible_env.HOME }}/src``` like you did, or use an variable to redirect the ```npm install``` to your new folder.

--------


last but not least , there's 2 scenarios:  
1. FIT test against deployment from personal code   (PR Gate or personal test )
2. FIT test against local code

for local-code-test,  the target audience is developer/contributor.
in their own VM, necessary packages & network have been configured. 
Even with code, before FIT being called, they usually already runs npm install , and update package.json OR create symbolic link of on-xxxx/node_modules/xx pointing to local on-core/on-tasks.
it's redundant that FIT to do those things again. 

in another word, it's rare that RackHD dev use ansible playbooks in their day to day work, in their own VM.

if a fresh deployment to a new VM:
rather than use “ansbile_from_local_src”, I perfer extend ansible/roles/remoterepos , to accept a variable of personal repo and branch(Oh, branch is already a variable in that playbook)
so we can do sth like ```ansible-playbook  local rackhd_local.yml   	gpaulos    test-branch```
the only trouble:
although developer aims to change on-core, he still need to create a branch in github.com/gpaulos/RackHD, and change the on-core "pointer" to his on-core and new branch.....

packer/ansible/roles/localsource/tasks/main.yml
```diff
@@ -0,0 +1,60 @@
+---
+- name: install dev libraries for NPM installs
+  apt: pkg={{ item }} state=installed
+  with_items:
+    - libkrb5-dev
+    - unzip
+  sudo: yes
+
+- name: Install RackHD from local source
+  copy: src="../../{{ item }}" dest="{{ ansible_env.HOME }}/src/"
```
gpaulos: That is the base dir of RackHD as referenced from the directory of the playbook.

packer/ansible/roles/localsource/tasks/main.yml
gpaulos: rackhd_local.yml:
https://github.com/RackHD/RackHD/blob/master/packer/ansible/rackhd_local.yml

panpan0000:
Dear @gpaulos, what do you think idea of reuse current scripts with parameter ? like my above long comment?

gpaulos:
Temporarily removing this PR for revision.

PR #528, opened 2016-12-12
Fix outdated elasticsearch log file

halfspiral:
Elasticsearch seems to crash when using docker-compose to start the application. Steps to reproduce along with the error message:

```bash
$ cd docker
$ docker-compose -p rackhd up -d
   # ...now wait a bit
$ docker logs rackhd_elasticsearch_1
ERROR: no log4j2.properties found; tried [/usr/share/elasticsearch/config] and its subdirectories
```

According to [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/settings.html), logging.yml is no longer used in Elasticsearch 5.x as the whole logging system has been replaced with Log4j 2. This PR reflects that change.


heckj:
thanks for the update Jason! Nice to see code from you again!

👍

benbp:
👍

PR #531, opened 2016-12-13
Update os_ova_install for Multiple ova install

patelb10:
RAC-3784 This PR Update os_ova_install.py to create multiple OVAs on stack. Replace ovamac in stack.config file with ovaoui in global.config file


patelb10:
@gpaulos @hohene

gpaulos:
Does this require a specific MAC address convention which maps to stack number? If so please document that in the header of the os_ova_install.py file. 

Note that not all stack designators are numbers, for example 'vagrant'. It looks like OVA install will require numbered stacks. Please document that too.

test/fit_tests/README.md
```diff
@@ -67,6 +68,8 @@ All FIT tests can be run from the wrapper 'run_tests.py':
       -template TEMPLATE  path or URL link to OVA template or OnRack OVA, default
                           from global_config.json
       -xunit              generates xUnit XML report files
+      -numvms NUMVMS      number of virtual machine for deployment on specified
+                          stack
```
hohene: typo - number of virtual machines    (vs machine)  
is there a default?  ie. default is 1?

test/fit_tests/deploy/os_ova_install.py
```diff
@@ -4,7 +4,11 @@
 Author(s):
 George Paulos
 
-This script installs blank OVA template.
+This script installs blank OVA templates into selected stack.
+if numvms is greater than 1, MAC address installation on OVA is done by stack number convention
+
+usage:
+    python run_tests.py -stack <stack ID> -numvms <number> -test deploy/os_ova_install.py
```
hohene: adding usage here - do we HAVE to now specify -numvms when we use this script?
if not, can you also add example usage for the old method?
python run_tests.py -stack <stack id> -test deploy/os_ova_install.py

test/fit_tests/config/global_config.json
```diff
@@ -89,5 +89,6 @@
     "_comment": "These are the northbound rest API port assignments",
     "http": 8080,
     "https": 443
-  }
+  },
+  "ovaoui": "00:50:56"
```
hohene: Could you please add a comment in to explain what this variable is- "_comment": "xxx"
And is the value good for the world to use or just our lab?

test/fit_tests/run_tests.py
```diff
@@ -63,6 +63,8 @@
                         help="path or URL link to OVA template or OnRack OVA, default from global_config.json")
 ARG_PARSER.add_argument("-xunit", default="False", action="store_true",
                         help="generates xUnit XML report files")
+ARG_PARSER.add_argument("-numvms", default=1, type=int,
+                        help="number of virtual machine for deployment on specified stack")
```
hohene: add 's' to machine  'machines'

test/fit_tests/deploy/os_ova_install.py
```diff
@@ -51,64 +56,76 @@ def test01_install_ova_template(self):
...
+            # check number of vms for deployment
+            if numvms == 1:
+                ovamacs = fit_common.STACK_CONFIG[fit_common.ARGS_LIST['stack']]['ovamac']
             else:
-                fit_common.time.sleep(10)
-        self.assertEqual(rc, 0, "VM did not activate.")
-        # Sync time on ORA
-        localdate = fit_common.subprocess.check_output("date +%s", shell=True)
-        fit_common.remote_shell("/bin/date -s @" + localdate.replace("\n", "") + ";/sbin/hwclock --systohc")
+                ovamacs = fit_common.GLOBAL_CONFIG['ovaoui'] + ":00:" + fit_common.ARGS_LIST['stack'] + ":0" + str(vm)
```
hohene: if your vm value is greater than 9, you'll get an invalid mac address.   :0 + str(12) would be :012?
Not sure if there is a limit to the number of VMs you can install?  but you may want to choose one.
And do you need a leading zero if ARGS_LIST['stack']  value is in 0-9?  ie.  00:50:56:00:1:02

test/fit_tests/deploy/os_ova_install.py
```diff
@@ -4,7 +4,11 @@
 Author(s):
 George Paulos
 
-This script installs blank OVA template.
+This script installs blank OVA templates into selected stack.
+if numvms is greater than 1, MAC address installation on OVA is done by stack number convention
```
hohene: Sorry, I don't understand this comment - do you mean if numvms is greater than 1, you will create MAC addresses based on the stack number?  stacknumber + vm number. and if equal to 1, then uses the default mac specified in the stack config file?
thanks!

gpaulos: I think that a better plan is to have a hex counter rather than tying the MAC to a stack number. What if the stack label is not a number? A counter would be more deterministic and transparent. In our MAC scheme we put in a full byte of space between each stack address for additional OVAs just for this situation. Assume that the base MAC always ends with 00 then count up from there based upon the supplied MAC from stack config.

patelb10: @hohene Yes you are right

test/fit_tests/deploy/os_ova_install.py
patelb10: Sure I can do this

test/fit_tests/config/global_config.json
patelb10: This is specific to our Lab which has this kind of MAC address convention

test/fit_tests/run_tests.py
patelb10: Sure

test/fit_tests/deploy/os_ova_install.py
johren: We should describe here what the "stack number convention" is:
OUI:00:stack-number:VM-number

johren: @gpaulos :  She was following what you already had set up as a pattern in our DHCP and DNS configuration in the lab.  For example:  stack14-ora.admin is 00:50:56:00:14:00

test/fit_tests/deploy/os_ova_install.py
johren: For both the VM number and the stack number, you can use the string format to enforce fixed with of 2.  For example:

stacknum = 5
print "Stack number {0:02}".format(stacknum)

Will print "Stack number 05"

PR #535, opened 2016-12-20
RAC-805 - methods to use within scripts for checking the test bed

larry-dean:
* Created  file common/api-utils.py
* Added method api_node_select_argsList()
* Added method api_node_select()
* Ccrrected if chain in common/fit_common -> restful()
* Remove sys(255) call as suggested by PR

### DIFF
--- a/test/common/api_utils.py
+++ b/test/common/api_utils.py
@@ -1,0 +1,199 @@
+from on_http_api2_0 import ApiApi as Api
+from config.api2_0_config import config
+from json import loads
+import sys
+import os
+
+
+
+def api_node_select_argsList(node_type='compute', validate_obm=False):
+    """
+    This routine produces a list of node Ids that satisfy the requested parameters. 
+    Nodes may be filtered by Node Id,  SKU name, OBM MAC address which are defined
+    in the system environment.  The node type and OBM setting may be specified.
+    :param  node_type: Node type (i.e. compute) (default= 'compute')
+    :param  validate_obm: should the node have OBM settings (default=False)
+    return: A list with node IDs that match node type, SKU, and possible valid OBM settings.
+    """
+    return api_node_select(config.api_client,
+                           node_id=os.getenv("node_id", "None"),
+                           sku_name=os.getenv("SKU", "all"),
+                           obm_mac=os.getenv("obm_mac", "all"),
+                           node_type=node_type,
+                           validate_obm=validate_obm,
+                           verbosity=os.getenv("VERBOSITY", "0")
+                           )
+
+
+
+def api_node_select(client,
+                    node_id=None,
+                    sku_name='all',
+                    obm_mac='all',
+                    node_type='compute',
+                    validate_obm=False,
+                    verbosity=None
+                    ):
+    """
+    This routine produces a list of node Ids that satisfy the requested parameters. 
+    Nodes may be filtered by Node Id,     SKU name, OBM MAC address, or node type. 
+    The OBM setting may also be validated (validate_obm).
+    Note: Validation of obm setting is not provides if node id is specified.
+    :param  client: reference to config.api2_0_config
+    :param  node_id: ID of node object (default=None)
+    :param  sku_name: Name of sku (default='all')
+    :param  obm_mac: OBM mac address (default='all')
+    :param  node_type: Node type (i.e. compute) (default= 'compute')
+    :param  validate_obm: should the node have OBM settings (default=False)
+    return: A list with node IDs that match node type, SKU, and possible valid OBM settings.
+    """
+
+    node_id_list = []
+    sku_id = "None"
+
+    # set verbosity level
+    if not verbosity:
+        verbosity = os.getenv("VERBOSITY", "0")
+
+    # if user specified single node id, don't bother getting sku information
+    if not node_id:
+        #
+        # Get a list of the current SKU objects
+        #
+        Api().skus_get()
+        sku_list_rsp = client.last_response
+        sku_list = loads(sku_list_rsp.data)
+
+        # Determine if the list was obtained without error
+        if sku_list_rsp.status != 200:
+            # Can't get a sku so leave sku_id as None
+            print '**** Unable to retrieve SKU list via API. status ({})\n'.format(sku_list_rsp.status)
+
+        # Search the list of sku objects for the possible requested sku name. This is skipped if
+        # the requested sku name is 'all', which will then use the default value of sku_id ('None')
+        elif sku_name != 'all':
+            for sku_entry in sku_list:
+                if sku_name == sku_entry['name']:
+                    sku_id = sku_entry['id']
+                    break
+
+    #
+    # Collect a list of node objects that match node type and possible sku type
+    #
+    Api().nodes_get_all()
+    node_list_rsp = client.last_response
+    node_list = loads(node_list_rsp.data)
+
+    # Determine if the list was obtained without error
+    if node_list_rsp.status != 200:
+        # There is no chance of getting any nodes so return an empty list
+        print '**** Unable to retrieve node list via API. status ({})\n'.format(node_list_rsp.status)
+        return []
+
+    # Select node by node type and SKU
+    for node_entry in node_list:
+        if node_id and node_id == node_entry['id']:
+            # check if user specified a single node_id to run against
+            # user must know the node_id and any check for a valid node_id is skipped
+            node_id_list = [node_id]
+            break
+        elif obm_mac == 'all':
+            # We are searching for all nodes with the requested sku
+            if sku_name == 'all':
+                # Select only managed compute nodes
+                if node_type == 'all' or node_entry['type'] == node_type:
+                    if not validate_obm or validate_obm_settings(client, node_entry['id']):
+                        node_id_list.append(node_entry['id'])
+            else:
+                if node_entry.get('sku') and sku_id in node_entry['sku']:
+                    if not validate_obm or validate_obm_settings(client, node_entry['id']):
+                        node_id_list.append(node_entry['id'])
+        else:
+            # we are looking for a particular node with the requested OBM MAC
+            Api().nodes_get_obms_by_node_id(identifier=node_entry['id'])
+            obm_list_rsp = client.last_response
+            obm_list = loads(obm_list_rsp.data)
+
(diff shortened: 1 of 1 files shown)


larry-dean:
@johren @jimturnquist @derrickostertag @hohene @gpaulos

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
+from on_http_api2_0 import ApiApi as Api
+from config.api2_0_config import config
+from json import loads
+import sys
+import os
+
+
+def api_node_select_argsList(nodeType='compute', validate_obm=False):
+    """
+    This routine produces a list of node Id's that satisfy the requested parameters.
```
yyscamper: typo? `Id's` => `Ids`.
The same for line 37

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
...
+    :param  validate_obm: should the node have OBM settings (default=False)
+    return: A list with node IDs that match node type, SKU, and possible valid OBM settings.
+    """
+
+    nodeIdList = []
+    skuid = "None"
+
+    # set verbosity level
+    if not verbosity:
+        verbosity = os.getenv("VERBOSITY", "0")
```
yyscamper: Shouldn't  the default verbosity be the number `0` rather than string `"0"`?

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def api_node_select(client,
+                    nodeId=None,
+                    skuName='all',
+                    obmmac='all',
+                    nodeType='compute',
+                    validate_obm=False,
+                    verbosity=None
+                    ):
```
yyscamper: Considering the flexibility if we need further filter criteria in future and the easy to use because developer hard to remember all these argument name and order, I suggest putting all these criterias into a dict, below is the sample code:

```python
def api_node_select(client, criteria, verbosity):
  skuName = criteria.get('skuName', 'all')
  nodeType = criteria.get('nodeType', 'compute')
  ...
```

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
...
+        verbosity = os.getenv("VERBOSITY", "0")
+
+    # if user specified single node id, don't bother getting sku information
+    if not nodeId:
+        #
+        # Get a list of the current SKU objects
+        #
+        Api().skus_get()
+        skulistrsp = client.last_response
+        skulist = loads(skulistrsp.data)
```
yyscamper: This 3 lines of code confuse me. the instance `client` isn't passed into `Api().skus_get()`, but it *magically* stores the result  . So seems the `client` is a global variable, right? The code should avoid output surprise. Why not do as following?
```python
result = Api().skus.get()
status = result.status // or result.last_response.status
skulist = loads(result.data)
```

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Determine if the list was obtained without error
+        if skulistrsp.status != 200:
```
yyscamper: I even think this kind of error handling should be put inside the `Api().skus.get()`. Of course, this should belong to a defect for previous `Api().skus.get()` design, not need to blame your PR.

test/common/api_utils.py
yyscamper: one more problem is the naming for each argument. sometimes the words in argument is connected with underscore, such as `validate_obm`, sometimes they are connected by capitalizing the first letter, such as `skuName`, and sometime you connect them with all lower letters, such as `obmmac`.
Keep the variable naming in the same style is important for readability.  As I know,  the python prefer the underscore version.

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            if obmlistrsp.status != 200:
+                # There is no chance of getting the requested obm object using this node id, just 
+                # try another node.
+                print '**** Unable to retrieve obm list via API for node: {} status: ({}).\n'.format(nodeentry['id'],
+                                                                                                     obmlistrsp.status)
+                continue
+
+            # does this node use the requested OBM MAC
+            for obmentry in obmlist:
+                if obmentry['config'] and obmentry['config'].get('host') == obmmac:
```
yyscamper: Sometimes the `host` is not macAddress, it is the ipAddress of BMC

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
...
+    if obmlistrsp.status != 200:
+        # There is no chance of getting the requested obm object using this node id, just 
+        # return False to indicate no OBM settings
+        print '**** Unable to retrieve obm list via API for node:: {} status: ({}).\n'.format(identifier,
+                                                                                              obmlistrsp.status)
+        return False
+
+    # does this node use the requested OBM MAC
+    for obmentry in obmlist:
+        if obmentry['config'] and obmentry['config'].get('user'):
```
yyscamper: For some obm, like this one https://github.com/RackHD/on-http/blob/master/static/schemas/obms/panduit-obm-service.json,  there is no `user` field, but from your comment `determine if at least one has OBM settings`, this should return True as well. If so, you need a fix here.

test/common/api_utils.py
```diff
@@ -0,0 +1,170 @@
...
+                           skuName=os.getenv("SKU", "all"),
+                           obmmac=os.getenv("OBMMAC", "all"),
+                           nodeType=nodeType,
+                           validate_obm=validate_obm,
+                           verbosity=os.getenv("VERBOSITY", "0")
+                           )
+
+
+
+def api_node_select(client,
```
yyscamper: I tried to read the whole implementation of this function and gives me a feeling of the logic is not so clear. Maybe I don't fully understand your design and maybe I don't understand all challenge you meet, if you have enough reason, I'm OK for your code.

I just thinking about if I want to implement this feature, I will take following steps (just some pseudocode):
```python
node_list = api.get_all_nodes();
filter_by_nodeId(node_list); // if any node doesn't meet the nodeId requirement, it will be removed from node_list
filter_by_skuName(node_list);  // if any node doesn't meet the skuName requirement, it will be removed from node_list
filter_by_nodeType(node_list);  // if any node doesn't meet the nodeType requirement, it will be removed from node_list
.... //filter by other criteria
//Until this step, the node_list stores all valid node
node_ids = extract_node_ids(node_list); //return only node id rather than the whole node document.
```

test/common/api_utils.py
larry-dean: Api().sku.get() does not return a object containing status or data. It's another problem with Api().

test/common/api_utils.py
larry-dean: Will correct

test/common/api_utils.py
larry-dean: This area will redone after Stuarts logging changes go in.

test/common/api_utils.py
larry-dean: Will correct

test/common/api_utils.py
larry-dean: I'll add logic to differentiate between macs and ip addresses.

larry-dean:
Peer review changes pushed up

test/common/api_utils.py
larry-dean: The function is very similar to the select_node() function in fit_test. The function is designed to function as a form of dependency check and not as a general filter.

test/common/api_utils.py
```diff
@@ -0,0 +1,206 @@
...
+
+    dict_obj = source_dict
+
+    for search_key in search_string.split("."):
+        try:
+            dict_obj = next(val for key, val in dict_obj.iteritems() if key == search_key)
+        except StopIteration:
+            return default_if_not_found
+    return dict_obj
+
```
stuart-stanley: nit: lots of empty lines at the end of the file here...

test/common/api_utils.py
stuart-stanley: That makes sense (at least to me ;) ). If the call-count is small (which is appears to be... maybe in the order of tens), if we need the additional capabilities, we can trade the clarity of a named arguments for the flexibility of a search-set later down the line. (just my opinion there, of course)

test/common/api_utils.py
```diff
@@ -0,0 +1,206 @@
...
+                        node_id_list = [node_entry['id']]
+                        break
+
+            # determine if the MAC was found, if so exit node loop
+            if node_id_list:
+                break
+
+    # we now have a node ID list
+    #if verbosity >= 3:
+    if 4 == 3:
```
jimturnquist: What does the above mean?   if 4 == 3 I mean.

test/common/api_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    This routine produces a list of node Ids that satisfy the requested parameters. 
+    Nodes may be filtered by Node Id,  SKU name, OBM MAC address which are defined
+    in the system environment.  The node type and OBM setting may be specified.
+    :param  node_type: Node type (i.e. compute) (default= 'compute')
+    :param  validate_obm: should the node have OBM settings (default=False)
+    return: A list with node IDs that match node type, SKU, and possible valid OBM settings.
+    """
+    return api_node_select(config.api_client,
+                           node_id=os.getenv("node_id", "None"),
+                           sku_name=os.getenv("SKU", "all"),
```
jimturnquist: The os.environ behavior of RackHD tests is changing.  There won't be things like SKU passed in the environment.  We do have an arg_list that will be available through the config block.  The change as you have it is fine for today but will need to updated shortly.

larry-dean: This function was added for the sole reason of current argList comparability and I will shed no tears removing it if the environment changes.

test/common/api_utils.py
larry-dean: if 4 == 3: means skip the following code ;-). I'll remove it.

test/common/api_utils.py
larry-dean: I'll remove

larry-dean:
Changes push up.

johren:
It looks like @yycamper's comments have been addressed.  I will wait for @jimturnquist's approval from his comments before merging.

test/common/api_utils.py
jimturnquist: OK.  Leave as-is for now.

jimturnquist:
+1 from me

PR #536, opened 2016-12-20
RAC-3745 added base config files only

gpaulos:
This PR contains the new base configuration files for the default 'Vagrant' config directory and the 'MN lab' config.mn directory as described in RAC-3745. 

This commit is only the config files, they are not yet enabled in the code. That is in an upcoming PR. There may be additional config files in Phase 1 as needed but this is the base requirement. Committing this PR will not affect current code execution. 

MN lab specific entries have been removed from the default config dir because that is intended as the 'Vagrant' configuration base.

There are no CIT configuration files (*.ini and *.py) in config.mn. MN testing is FIT only because CIT configuration files only run from the default dir and will be deprecated when Phase 1 is complete anyways.

@hohene @jimturnquist @johren


hohene:
My initial review of this, it looks fine for staging. +1

gpaulos:
@yyscamper 

In principle, you are right. We should not put credentials etc. into the public repo.

However, if you compare the credentials files, they are pretty much identical because these are the defaults. The only actual differences between the two config dirs are the proxies and the stack config. I don't think anything here is a real security violation. 

I think having config.mn is a good example of an actual production environment.

If you are still not convinced, then I will remove config.mn, but this will slow us down here because we will need to put up our own config repo and overlays, etc.

ebedros:
+1 
@yyscamper The MN Config files uses the Default passwords and the MN links are another examples  provided for reference.  Maybe we can add a README file to explain the files if it helps.

gpaulos:
Removed config.mn directory to conform to @yyscamper request.

johren:
@yyscamper  Please verify that @gpaulos's latest changes address your concerns.   I will merge tonight unless I hear otherwise.

yyscamper:
I just see the repo has already stored some credential information in this file https://github.com/RackHD/RackHD/blob/master/test/config/global_config.json, It's a miss that we didn't catch this before. So it's not fair to blame this PR.

Though we are using the default password, and we don't care to let public know what the password we are using. But this surely expose security concern. How about in future we don't use default password?

Due to the historical reason and not want to slow down your current work, I am OK to change this to +0.

Though not to force you remove out the credential info in this PR, However, could you please make a plan to remove them in future PR?

johren:
Thanks, Felix.  I will work with the others regarding the security concerns.

PR #537, opened 2016-12-20
RAC-267 Multiple OVA installation per stack

patelb10:
@gpaulos @johren @hohene

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -38,6 +38,7 @@ ARGS_LIST = \
     "version": os.getenv("VERSION", "onrack-devel"), # code version
     "template": os.getenv("TEMPLATE", "None"), # path or URL link to OVA for deployment
     "xunit": os.getenv("XUNIT", False), # XUNIT output
+    "numvms" : os.getenv("NUMVMS", 1), # number of OVA for reployment
     "list": os.getenv("LIST", False), # list tests
     "group": os.getenv("GROUP", "all"), # test group
     "http": os.getenv("HTTP", "False"), # force http api protocol
@@ -94,6 +95,9 @@ ARGS_LIST.update(
 )
 
 if ARGS_LIST["stack"] != "None":
+    if ARGS_LIST["stack"] not in STACK_CONFIG:
+        print "**** Stack {0} not found in stack config file {1}.  Exiting....".format(ARGS_LIST["stack"], CONFIG_PATH + "stack_config.json")
+        sys.exit(255)
     if "ora" in STACK_CONFIG[ARGS_LIST["stack"]]:
         ARGS_LIST["ora"] = STACK_CONFIG[ARGS_LIST["stack"]]['ora']
     else:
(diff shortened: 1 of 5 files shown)


test/common/fit_common.py
```diff
@@ -38,6 +38,7 @@
     "version": os.getenv("VERSION", "onrack-devel"), # code version
     "template": os.getenv("TEMPLATE", "None"), # path or URL link to OVA for deployment
     "xunit": os.getenv("XUNIT", False), # XUNIT output
+    "numvms" : os.getenv("NUMVMS", 1), # number of OVA for reployment
```
yyscamper: typo: reployment => deployment.

test/deploy/os_ova_install.py
```diff
@@ -51,64 +63,78 @@ def test01_install_ova_template(self):
...
-            print "MAC address processing failed. Exiting..."
-            sys.exit(255)
-
-        # Poll the OVA via ping
-        for dummy in range(0, 30):
-            rc = subprocess.call("ping -c 1 -w 5 " + fit_common.ARGS_LIST['ora'], shell=True)
-            if rc == 0:
-                break
+        for vm in range(0, numvms):
+            print '**** Deploying OVA file on hypervisor ' + fit_common.ARGS_LIST['hyper']
```
yyscamper: Please create a sub function to handle the process about deploying OVA file for one VM (all codes inside this `for` loop), so the code will look more clean.

test/deploy/os_ova_install.py
```diff
@@ -18,19 +23,26 @@
...
-            fit_common.time.sleep(5)
 
         # Run probe to check for valid OVA file
         rc = fit_common.subprocess.call("ovftool " + ovafile, shell=True)
         self.assertEqual(rc, 0,'Invalid or missing OVA file: ' + ovafile)
 
+        # check for number of virtual machine and stack ID as number
+        self.assertTrue(numvms < 100, "Number of vms should not be greater than 99")
+        if numvms > 1:
+            self.assertTrue(fit_common.ARGS_LIST['stack'].isdigit(), "Stack ID must be a number if numvms is greater than 1")
```
patelb10: Here if numvms==1, then we use MAC address from stack_config file which has stack ID either number or word, And if numvm>1 then we defined MAC address from stack ID which must be number in order to create MAC address

yyscamper: @patelb10 : thanks for your reply. Could you please add your reply as comment into your code?

test/deploy/os_ova_install.py
```diff
@@ -49,15 +64,18 @@ def test01_install_ova_template(self):
                                 timeout=20, logfile=None)
             if exitstatus == 0:
                 break
-
         # Run OVA installer
+        for vm in range(0, numvms):
+            self.deploy_ova(vm,uname,passwd,numvms,ovafile)
+
+    def deploy_ova(self,vm,uname,passwd,numvms,ovafile):
```
johren: I don't think this should go inside the test.  That makes it harder to read.  It should be inside the class but outside of test01_install_ova_template.  Perhaps before the definition of test01_install_ova_template.

johren:
Needs one more approval.  @hohene or @gpaulos

johren:
@diontje

PR #539, opened 2016-12-22
Add decommission test to regression group

keedya:
Test includes:
- set bmc credentials using rackhd workflow
- decommission node includs:
-- remove selected bmc users
-- erase disk


keedya:
@RackHD/corecommitters @tannoa2 @geoff-reid @uppalk1

PR #548, opened 2016-12-28
use on-imagebuilder.deb instead of raw static files

changev:
Docker used to download static files from bintray directly.
But now static files have been wrapped into on-imagebuilder.deb, so this Docker file should be modified.

skip ci

@panpan0000 @PengTian0


changev:
This proposal is abandoned

PR #555, opened 2017-01-03
change wrong 'rackhd-mirror' to rackhd

changev:
rackhd-mirror is the test mirror of rackhd, it should not appear in prod code.

@panpan0000 @PengTian0


panpan0000:
:+1:

PR #558, opened 2017-01-05
Rac 3817 capture nose stuff

stuart-stanley:
See https://github.com/RackHD/RackHD/pull/569 instead, though the approvals here should still be valid.

*THIS IS BASED ON THE CURRENTLY UNMERGED RAC-3789* and includes that commit. Please review the 2nd commit, but don't yet!

Note that the following stories have been opened to round out the abilities added here:
* RAC-3869 to handle retro-grabbing of DEBUG level log output and such
* RAC-3870 to getting the progress line data into a log entry instead of raw output
* RAC-3875 to capture data _after_ an error has occured (like for tearDown and such)

Note: xunit, capture, and logcapture all do similar things, but everyone reports AFTER all the tests are done. Handy for style of use, really annoying for say Jenkins jobs!

Primary changes:
* Basically stole the xunit logic to capture stdout/stderr
* on addError or addFailure: grab the captured stdout/stderr and call any stream-source that implements "handle_capture"
* added handle_capture to stream_sources/log_type
** Manually grab list of logs to inject into for now (RAC-3869 deals with less hard-coded style)
** Inject stdout, stderr, and the traceback into those logs. Routine disables propegation to avoid duplication right now. De-hacking that would come out of RAC-3869)

TDD support changes: (note, since we don't HAVE the stream-monitor code yet, this is fairly sloppy cut-and-paste work)
* Removed near duplicate code for _TestStreamMonitorTester and _TestStreamMonitorPluginTester and added put into _BaseStreamMonitorPluginTester
* Adjusted existing tests to reference _BaseStreamMonitorPluginTester via "plugin_test_helper.resolve_helper_class()"
* Added new abilities to _BaseStreamMonitorPluginTester:
** Allow for sub-classes to set "_expect_nose_success" to False to disable checking for if the tests they ran had errors or not
** Check for and report if the sub-tests failed!
** Added _check_sequences to auto-build name-of-test used in test-sequence checking (got tired of constantly making changes to lines like "self._check_sequence_test('test_one_of_two_to_run (test_two_log_pass.TestTwoLoggerTest)\
')" for the next test that just replaced the three names in there!)
** Added a resolve_suitepath to the module to limit the filesystem knowledge to one place.
** Added a resolve_helper_class() to the module to get access to _BaseStreamMonitorPluginTester (if I leave the _ off, nosetests tries to "test" it, which causes all sorts of trouble. We only want classes _derived_ from it tes\
ted!)
* And then we have the super-crude changes in test_log_stream.py to add _TempLogfileObserver and _TempLogfileChecker. These:
** basically implement a "tail -f <logfile> | egrep <pattern>" (the Observer)
** and the biz logic to check for proper existance (or non-existance) of backtrace, captured stdout, and captured stderr data in the monitored file
* Also created _BacktraceBase with a derived class for each logfile to check for proper backtrace data showing up (or not)
* Also create _CaptureBase with logic to handle checking for proper stdout OR stderr capture data. This is not used directly, but is subclassed by the following who fill in key class-values:
** _StdoutNoCaptureBase (tests that no stdout capture occurs when there are no errors)
** _StdoutErrCaptureBase (tests that an error (i.e., code error that causes exception) captures stdout
** _StdoutFailCaptureBase (tests that a fail (i.e., unittest.assertTrue() based tests failures rather than errors) captures stdout
** _StderrNoCaptureBase, _StderrErrCaptureBase, and _StderrFailCaptureBase which do the same thing as the above, but for stderr
* Each of the above implements a plugin-test-class for each logfile, which deals with "how many times should I find this?" for each logfile

This means we have tests for the following to be seen in files that are supposed to have them, or not if they aren't.
* traceback generated by error
* traceback generated by fail (assertTrue test failure)
* stdout from setup and test-method generated by error
* stdout from other test-method in same test-suite as last doesn't show
* stderr from setup and test-method generated by error
* stderr from other test-method in same test-suite as last doesn't show
* if no-error occurs, no traceback, stdout, or stderr seen

Note: these tests or in the -plugin- directory, and are not run automatically anywhere at this point. I created RAC-3876 to cover getting this into its own jenkins at some point.

@RackHD/corecommitters @johren @hohene @gpaulos @larry-dean @jimturnquist @keedya @panpan0000 @changev @anhou

### DIFF
--- a/test/mkenv.sh
+++ b/test/mkenv.sh
@@ -50,7 +50,7 @@ source .venv/${env_name}/bin/activate
 export PIP_CONFIG_FILE=`pwd`/pip.conf
 
 # Update local-pip to latest
-pip install -U pip
+pip install --upgrade pip
 
 # Install all required packages
 pip install -r requirements.txt
--- a/test/stream-monitor/flogging/__init__.py
+++ b/test/stream-monitor/flogging/__init__.py
@@ -2,4 +2,5 @@ from infra_logging import *
 from logger_sets import get_loggers
 
 all = ['getLogger', 'getInfraRunLogger', 'getInfraDataLogger',
-       'getTestRunLogger', 'getTestDataLogger', 'get_loggers', 'logger_config_api']
+       'getTestRunLogger', 'getTestDataLogger', 'get_loggers',
+       'logger_get_logging_dir', 'logger_config_api']
--- a/test/stream-monitor/flogging/infra_logging.py
+++ b/test/stream-monitor/flogging/infra_logging.py
@@ -153,18 +153,23 @@ class _LoggerSetup(object):
             shutil.rmtree(trim_name)
 
         ts_ext = datetime.now().strftime('%Y-%m-%d_%X')
-        self.__lg_run_dir = os.path.join(lg_base_dir, 'run_{0}.d'.format(ts_ext))
+        run_name = 'run_{0}.d'.format(ts_ext)
+        self.__lg_run_dir = os.path.join(lg_base_dir, run_name)
         self.__prelog(logging.INFO, "this runs logging dir %s", self.__lg_run_dir)
         self.__makedirs_dash_p(os.path.join(self.__lg_run_dir))
 
         # now deal with and easy to use 'last'
-        last_name = os.path.join(lg_base_dir, 'run_last.d')
-        if os.path.islink(last_name):
-            os.unlink(last_name)
+        last_name = 'run_last.d'
+        last_path = os.path.join(lg_base_dir, last_name)
+        if os.path.islink(last_path):
+            os.unlink(last_path)
 
-        assert not os.path.lexists(last_name), \
+        assert not os.path.lexists(last_path), \
             "'{0}' still existed after unlink. Check for file/dir and remove manually".format(last_name)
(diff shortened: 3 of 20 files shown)


gpaulos:
Tested with a number of options. Appears to work as stated, but not quite clear what the expected results are supposed to look like. 

Tentative +1

stuart-stanley:
@gpaulos still a WIP, but as kind of a minimal view of the core change here was to make the stuff that nose dumps out at the end of the run all in a batch (errors and failures) to the console to also be included WHEN the test actually hits the error/failure. Both on the console, and in the on-disk .log files.

The stuff captured by xunit probably has not changed, I would think.

gpaulos:
I ran a job with 2 failures and you are correct, only the failed tests dumped to console and xunit.

stuart-stanley:
@hohene : the PR that turned off the new format for unmirgrated CIT tests also fixed the multi-rotation of log directories problem.  
It's probably what is causing the branch conflicts!  I'll try to fix those today once I am slightly awake.

hohene:
Great!!

johren:
Should this PR be closed?  It was replaced by another PR, right?

stuart-stanley:
re-done in https://github.com/RackHD/RackHD/issues/569

PR #559, opened 2017-01-05
Remove API docs tests from smoke/regression

gpaulos:
These tests are causing errors in the CIT/FIT production jobs because the production RackHD deployment does not install the docs pages. 

This just removes the tests from the smoke and regression jobs. They can still be run individually if desired.


hohene:
+1

PR #562, opened 2017-01-06
fix the obvious error message problem..

panpan0000:
the original PR https://github.com/RackHD/RackHD/pull/560 has a obvious typo.

I ask CC's favour to merge https://github.com/RackHD/RackHD/pull/560 in , but I should be in charged of to fix this obvious typo

moreover, @gpaulos , https://github.com/RackHD/RackHD/pull/560 is a workaround, please do fix the ultimate root cause.

### DIFF
--- a/test/deploy/rackhd_stack_init.py
+++ b/test/deploy/rackhd_stack_init.py
@@ -117,13 +117,13 @@ class rackhd_stack_init(unittest.TestCase):
             fit_common.countdown(30)
         # no PDU case
         else:
+            print '**** No supported PDU found, restarting nodes using IPMI.'
             # Power cycle all nodes via IPMI, display warning if no nodes found
             if fit_common.power_control_all_nodes("off") == 0:
                 print '**** No BMC IP addresses found in arp table, continuing without node restart.'
             else:
-                print '**** No supported PDU found, restarting nodes using IPMI.'
-            # power on all nodes under any circumstances
-            fit_common.power_control_all_nodes("on")
+                # power on all nodes under any circumstances
+                fit_common.power_control_all_nodes("on")
 
     # Optionally install control switch node if present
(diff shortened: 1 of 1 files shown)


panpan0000:
@gpaulos ,  Do you agree of this PR change ? if yes, can you +1 ?

rolandpoulter:
@panpan0000 I don't see the typo? Are you saying the conditional has it backwards and the `print` statements need to be flipped?

test/deploy/rackhd_stack_init.py
```diff
@@ -119,9 +119,9 @@ def test04_power_on_nodes(self):
         else:
             # Power cycle all nodes via IPMI, display warning if no nodes found
             if fit_common.power_control_all_nodes("off") == 0:
-                print '**** No supported PDU found, restarting nodes using IMPI.'
-            else:
                 print '**** No BMC IP addresses found in arp table, continuing without node restart.'
+            else:
+                print '**** No supported PDU found, restarting nodes using IMPI.'
```
anhou: the typo still exist

anhou:
@panpan0000   so this PR include two changes, one is the about condition flip as @rolandpoulter  metioned and the other is typo?

panpan0000:
@rolandpoulter @anhou 
Yes. should be 2 changes.
1.  IMPI --> IPMI
2. Yes. flip the two print  per @changev .  the ```fit_common.power_control_all_nodes()``` will power off node via BMC, and return 0 when no BMC IP detects

-----------
but personally , will below be better ?
```
        else:
            print '**** No supported PDU found, restarting nodes using IPMI.'
            # Power cycle all nodes via IPMI, display warning if no nodes found
            if fit_common.power_control_all_nodes("off") == 0:
                print '**** No BMC IP addresses found in arp table, continuing without node restart.'
            else:
                # power on all nodes under any circumstances
                fit_common.power_control_all_nodes("on")
```

gpaulos:
+1

PR #569, opened 2017-01-10
RAC-3817 capture nose stuff

stuart-stanley:
This is a re-submit of https://github.com/RackHD/RackHD/pull/558, which included the symlink changes in it (which came in under their own PR). @hohene and @gpaulos approved the nose-capture part of that PR already. @RackHD/corecommitters for merging please!

Note that the following stories have been opened to round out the abilities added here:
* RAC-3869 to handle retro-grabbing of DEBUG level log output and such
* RAC-3870 to getting the progress line data into a log entry instead of raw output
* RAC-3875 to capture data _after_ an error has occured (like for tearDown and such)

Note: xunit, capture, and logcapture all do similar things, but everyone reports AFTER all the tests are done. Handy for style of use, really annoying for say Jenkins jobs!

Primary changes:
* Basically stole the xunit logic to capture stdout/stderr
* on addError or addFailure: grab the captured stdout/stderr and call any stream-source that implements "handle_capture"
* added handle_capture to stream_sources/log_type
** Manually grab list of logs to inject into for now (RAC-3869 deals with less hard-coded style)
** Inject stdout, stderr, and the traceback into those logs. Routine disables propegation to avoid duplication right now. De-hacking that would come out of RAC-3869)

TDD support changes: (note, since we don't HAVE the stream-monitor code yet, this is fairly sloppy cut-and-paste work)
* Removed near duplicate code for _TestStreamMonitorTester and _TestStreamMonitorPluginTester and added put into _BaseStreamMonitorPluginTester
* Adjusted existing tests to reference _BaseStreamMonitorPluginTester via "plugin_test_helper.resolve_helper_class()"
* Added new abilities to _BaseStreamMonitorPluginTester:
** Allow for sub-classes to set "_expect_nose_success" to False to disable checking for if the tests they ran had errors or not
** Check for and report if the sub-tests failed!
** Added _check_sequences to auto-build name-of-test used in test-sequence checking (got tired of constantly making changes to lines like "self._check_sequence_test('test_one_of_two_to_run (test_two_log_pass.TestTwoLoggerTest)')" for the next test that just replaced the three names in there!)
** Added a resolve_suitepath to the module to limit the filesystem knowledge to one place.
** Added a resolve_helper_class() to the module to get access to _BaseStreamMonitorPluginTester (if I leave the _ off, nosetests tries to "test" it, which causes all sorts of trouble. We only want classes _derived_ from it tested!)
* And then we have the super-crude changes in test_log_stream.py to add _TempLogfileObserver and _TempLogfileChecker. These:
** basically implement a "tail -f <logfile> | egrep <pattern>" (the Observer)
** and the biz logic to check for proper existance (or non-existance) of backtrace, captured stdout, and captured stderr data in the monitored file
* Also created _BacktraceBase with a derived class for each logfile to check for proper backtrace data showing up (or not)
* Also create _CaptureBase with logic to handle checking for proper stdout OR stderr capture data. This is not used directly, but is subclassed by the following who fill in key class-values:
** _StdoutNoCaptureBase (tests that no stdout capture occurs when there are no errors)
** _StdoutErrCaptureBase (tests that an error (i.e., code error that causes exception) captures stdout
** _StdoutFailCaptureBase (tests that a fail (i.e., unittest.assertTrue() based tests failures rather than errors) captures stdout
** _StderrNoCaptureBase, _StderrErrCaptureBase, and _StderrFailCaptureBase which do the same thing as the above, but for stderr
* Each of the above implements a plugin-test-class for each logfile, which deals with "how many times should I find this?" for each logfile

This means we have tests for the following to be seen in files that are supposed to have them, or not if they aren't.

### DIFF
--- a/test/stream-monitor/sm_plugin/stream_monitor.py
+++ b/test/stream-monitor/sm_plugin/stream_monitor.py
@@ -3,11 +3,17 @@ import os
 from nose.plugins import Plugin
 from stream_sources import LoggingMarker
 import sys
+from nose.pyversion import format_exception
+from nose.plugins.xunit import Tee
+from nose import SkipTest
+from StringIO import StringIO
+from logging import ERROR, WARNING
 
 
 class StreamMonitorPlugin(Plugin):
     _singleton = None
     name = "stream-monitor"
+    encoding = 'UTF-8'
     def __init__(self, *args, **kwargs):
         assert StreamMonitorPlugin._singleton is None, \
             "infrastructure fault: more than one StreamMonitorPlugin exists"
@@ -18,6 +24,9 @@ class StreamMonitorPlugin(Plugin):
         # self.__print_to = sys.stderr
         # todo: use nose plugin debug options.
         self.__stream_plugins = {}
+        self.__capture_stack = []
+        self.__current_stdout = None
+        self.__current_stderr = None
         super(StreamMonitorPlugin, self).__init__(*args, **kwargs)
 
     @classmethod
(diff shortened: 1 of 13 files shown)


stuart-stanley:
test this please

PR #574, opened 2017-01-12
[skip ci]modifydocker-compose.yml to support release/ci-builds

changev:
Add ${TAG} in docker-compose, support:
```
$ cd RackHD/docker                            # TAG can be a release version, if not set default: latest
$ sudo TAG=${TAG} docker-compose pull         # Download prebuilt docker images.
$ sudo TAG=${TAG} docker-compose up           # Create containers and Run RackHD and ELK.
```

if just
``` 
sudo docker-compose pull  
```
docker-compose will get default value of ${TAG} in .env .
@panpan0000 @PengTian0

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -1,0 +1,1 @@
+TAG=latest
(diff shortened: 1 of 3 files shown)


PengTian0:
:+1:

PR #576, opened 2017-01-13
Corrections to Vagrantfile

gpaulos:
The FIT Vagrantfile broke when the fit_tests path was removed. This is a required correction to execute Vagrant FIT deploy/test.

These are the corrections:
- the path was changed from fit_tests/test to test
- the 'echo' statement was added at the end of the test commands because Vagrant stops script processing when it receives an non-zero exit code. This ensures that script processing continues even with a minor error.
- Port mapping was added to the template section to match the dev section.


gpaulos:
Have an oops, will resubmit.

PR #578, opened 2017-01-16
added wss to the pm2 config

DavidjohnBlodgett:
@heckj @RackHD/corecommitters 

Discovered that the pm2 config was missing the websocket server, so posting this quick fix.

-DJ

### DIFF
--- a/packer/ansible/roles/pm2/files/rackhd-pm2-config.yml
+++ b/packer/ansible/roles/pm2/files/rackhd-pm2-config.yml
@@ -14,3 +14,6 @@ apps:
    - script: index.js
      name: on-tftp
      cwd: ./src/on-tftp
+   - script: index.js
+     name: on-wss
+     cwd: ./src/on-wss


rolandpoulter:
@yyscamper I think the CI should use a different config specifically for testing essential RackHD components. Though it may be a good idea to include `on-wss` in the tests despite it being non-essential.

yyscamper:
@DavidjohnBlodgett @rolandpoulter : How about changing this PR to create a new pm2 config file? this config file will run on-wss, user can choose to run this or that, and meanwhile the current CI doesn't need any change.

rolandpoulter:
@yyscamper I think the CI should use a custom pm2 config. Generally that is how I would expect it to work. Ideally `on-wss` is running when people test out RackHD incase they want to use the UI

johren:
This PR appears to have gone stale.   Please reopen if ready to move forward.

PR #581, opened 2017-01-18
No need to ```pm2 start``` since Vagrant Box >=1.0.0

panpan0000:
RackHD is running as service(all rackhd package are installed as deb package in packer now) since vagrant box >=1.0.0, so no need to pm2 start

@heckj @anhou @changev

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -39,8 +39,7 @@ Clone RackHD repo to your local machine.
     cd example
     # create the RackHD instance.
     vagrant up dev
-    # start the RackHD services
-    vagrant ssh dev -c "sudo pm2 start rackhd-pm2-config.yml"
+    # then the RackHD services will start during bootup (they are running as service since box version 1.0.0)
 
 ### LOCAL SOURCE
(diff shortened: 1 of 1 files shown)


changev:
:+1: 
So  vagrant box with source-code-installation won't be provided anymore?

panpan0000:
as discussed, we need another vagrant box ```rackhd-dev``` for Development Enviroments and a script helps to clone

rolandpoulter:
Will this information still be relevant when we create another vagrant box `rackhd-dev`? Maybe we should keep it around until we can add the new box so we can move the documentation.

panpan0000:
Hi, @rolandpoulter , thanks for reviewing. you reminded me , I also need to modify the next chapter "LOCAL SOURCE", which used ```rackhd-dev``` box instead of ```rackhd``` box
👎  to myself until I resolve that.

PengTian0:
test this please

PengTian0:
test this please

PR #583, opened 2017-01-18
RAC-3946 Update RackHD test to only reference new configuration

jimturnquist:
- Converted over previous configuration references to new
- Obsoletes global_config.json
- Eliminates all but a few global symbols
- Add config dictionary-assist functions
- vagrant is now default stack configuration

@johren @gpaulos @larry-dean @stuart-stanley @hohene


jimturnquist:
Sample generated configuration shown below:

[fit-config-20170117-193917.txt](https://github.com/RackHD/RackHD/files/714542/fit-config-20170117-193917.txt)

gpaulos:
Tested on Vagrant and ESXi deployments.

test/config/credentials_default.json
```diff
@@ -19,8 +20,8 @@
     ],
     "ora": [
       {
-        "username": "vagrant",
-        "password": "vagrant"
+        "username": "onrack",
+        "password": "onrack"
```
stuart-stanley: Is this right with the default still being "vagrant"? (just curious, because it would be easy to be testing against a stack and miss changing this back!)

test/config/install_default.json
```diff
@@ -1,6 +1,11 @@
 {
-  "test-config": {
-      "template": "ora-ubuntu-1604.ova",
+  "install-config": {
```
stuart-stanley: are we allowed this use of an internal address?

test/config/credentials_default.json
gpaulos: The ora credentials are onrack/onrack in the MN config that we use. Remember this is the public config set.

test/config/install_default.json
gpaulos: No this should not have been changed. Please use "template": "ora-ubuntu-1604.ova" here

jimturnquist: OK.  I will make the change.

test/config/credentials_default.json
jimturnquist: OK.  These should be vagrant/vagrant then as Stuart said.  What about other "onrack" credentials:

  "node": [
      {
        "username": "root",
        "password": "1234567"
      },
      {
        "username": "onrack",
        "password": "onrack"
      }
    ],

jimturnquist: Also:  Our default community is "onrack".  Is that what we want?  Do we want "public" or something else to appear out on github?

gpaulos: I think that's fine.

gpaulos: I think that's fine. 'public' is a common default.

jimturnquist:
This is what I'm considering changing with regards to onrack references:

[diff.this.txt](https://github.com/RackHD/RackHD/files/715074/diff.this.txt)

jimturnquist:
Proposed fixes are out for review.

larry-dean:
+1

jimturnquist:
@hohene Good catch.  I did not intend the following to survive my latest changes:

VERBOSITY = int(os.getenv("VERBOSITY", "1"))
# The global "VERBOSITY" will go away once all usages are removed. At that point,
# the following call can migrate some place that makes more sense (not sure where
# that is yet!)

Fixing...

jimturnquist:
@hohene VERBOSITY has been fixed in most recent commit.

jimturnquist:
Test this please.

jimturnquist:
Rebased changes.

README.md has been cleaned up
Cleaned up global_config.json references.

jimturnquist:
@johren We are clean WRT rackhdci.  Erika and George have approved.  If you are OK with it, I think we are ready to merge.

johren:
Just have a questions regarding test_redfish10_os_bootstrap_base.py.  It looks like all of the test cases are being skipped in this suite.  Is this temporary or are there other plans for this suite?

jimturnquist:
@johren Yes, George is planning on reworking how os bootstrapping works.  We agreed to skipping these in the near term.

gpaulos:
@johren I have a story to fix the bootstrap tests.

PR #584, opened 2017-01-18
rac-3777 verbosity ctl

stuart-stanley:
This adds control of the levels to the new logging infra. This is a first step, and
is focused mainly on the controls via the nose-plugin options, but with some wiring
for (temporary) to run_test.py's -v specturm. (rac-3984 will dig into that further).

Alas, this is a fairly big change-set. The bulk of what was added can be summed up with
the following --help chunk:

  listing log settings:
    --sm-list-logging   list logging settings
    --sm-list-level=LIST_LOGGING_LEVEL
                        detail level if --list-logging is used

  setting log options:
    note: both 'logger-name's and 'handler-name's can contain wild-card
    characters to match multiple items.

    --sm-set-logger-level=('logger-name', 'level-name-or-value')
                        Set a logger's capture threshold. Loggers are
                        evaluated before handlers.
    --sm-set-handler-level=('handler-name[:logger-name]', 'level-name-or-value')
                        Set a handler's capture threshold.
    --sm-set-combo-level=('handler-name[:logger-name]', 'level-name-or-value')
                        Set a handler's capture threshold AND all loggers
                        feeding it.
    --sm-crank-file=('file-pattern', '[handler-name[:logger-name] [level-name-or-value]]')
                        changes the logging level for handler(s) on the given
                        loggers to a new value for matching files

Everything else is to support that and/or test it. I am updating the README.td file, in parallel with
starting the review, and the README will have an overview of the loggers and how loggers + handlers work. Actually, I'll problably start with the overview in this PR!

Overview of whats going on in order to make sense of the changes :) :
In a very brief sense, when you use the "main" logger from the first example conversion (logs = flogging.get_loggers()), you
are writing to the 'test.run' logger when you do a 'logs.debug()' or 'logs.info()' call. The sequence is a like this:

There are four infra-logging loggers: infra.run, infra.data, test.run, test.dat (and technically the root one, but it itself ins't used directly for log-calls)
So, by default, there are six files: 'test_run.log', 'test_data.log', 'infra_run.log', 'infra_data.log', 'combined_all_all.log', and 'console_capture.log', placed in the directory 'log_output/run_<datetime-stamp>.d'. 'log_output/run_last.d' is linked to the most recent of these.
test_run, test_data, infra_run, and infra_data all contain DEBUG level and above output for THEIR loggers.
combined_all_all contains DEBUG level and above from all of those.
console_capture contains only INFO level and above from the four infra-loggers (and INFO and above from any other logger outside of main infra-logging ones)

So, back to the --help stuff:
    --sm-set-logger-level=('logger-name', 'level-name-or-value')
                        Set a logger's capture threshold. Loggers are
                        evaluated before handlers.
This will change the logger level. For example, '--sm-set-logger-level test.run DEBUG_5' will change the logger level from DEBUG(_0) to DEBUG_5. I.E. DEBUG_6 will still be dropped by the test.run logger.
The way things are wired, this only impacts the level of messages flowing to test_run.log and combined_all_all.log. The messages would still be dropped by the root-logger, since it is at INFO.

Note that the logger name can be wild-carded: 'test.*' or 'test*' to change both test.run and test.data, or '*.data' to crank up the volume of data dumps.

    --sm-set-handler-level=('handler-name[:logger-name]', 'level-name-or-value')
                        Set a handler's capture threshold.

But if you wanted to have more than just DEBUG flow to stdout? '--sm-set-handler-level console DEBUG'. This will actually have less impact than you would think, since
both the root logger AND the console handler have a level threshold of 'INFO'.
The actual usefull application of this option would be to lower levels '--sm-set-handler-level console WARNING'. This would me only warnings and above to the console, while still sending INFO and above to the console_capture file.

Mostly, this is here for completeness, and for use 'behind the scenes' on the more usefull next option:

    --sm-set-combo-level=('handler-name[:logger-name]', 'level-name-or-value')
                        Set a handler's capture threshold AND all loggers
                        feeding it.

Doing a '--sm-set-combo-level console DEBUG' will go and change the handler level for 'console' AND all loggers feeding into it. Because 'console' is attached to the
root logger, this ends up override ALL the loggers (that had been created at this point. That is a key limitation).
Of course, this means that the console is being spammed with DEBUG output from everything. '--sm-set-combo-level console*:*.run DEBUG_5' will have the affect of:

    --sm-crank-file=('file-pattern', '[handler-name[:logger-name] [level-name-or-value]]')
                        changes the logging level for handler(s) on the given
                        loggers to a new value for matching files

This last option can be used to "crank up" the output for a specific file or files. The handler and logger name parts are just like in the combo.
The file pattern is a regex applied against the filename seen in the output records.

The changes made to do this:
* adapted run_test.py to be able to pass arguments it didn't use to nosetests (and get help back).
* changed the comment for the -v option in run_test.py and setup argument mappings to implement them.
* changed the wrapper in _LevelLoggerClass, which was wrongly using _log rather than log (_log didn't propagate!)
* changed levels in the dict-config from DEBUG to NOTSET infra handlers
* added a 'reset_config' method in flogging/infra_logging.py to squash configs between tests run in the plugin tester
* added a 'flogging/infra_logopts.py':
** contains a parser-helper class that other entry-points can call to add the infra-logging-opts options to it and then later call to process the args. This also includes the implementation for the list and set options.
** since nose is still on the old optargs, add a converter class to convert setup via argparse to the more limited optargs.
** the _LoggingConfigFilter class defined here implements the handler+logger filtering described above.
* wired the parser-helper setup and processing into the stream_monitor plugin.
* moved most of the file observer logic into its own file, log_stream_Helper.py
* added logic to same to deal with testing for level based logs.
* added a test_log_opts.py inside stream-monitor/test to self-test options.
* rewired imports a little to be more precise about when the "real" flogging gets imported/instantiated.

@RackHD/corecommitters @johren @hohene @gpaulos @larry-dean @jimturnquist @keedya @panpan0000 @changev @anhou

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -39,9 +39,9 @@ The FIT test framework is under RackHD/test
 ## Configuration
 
 Default configuration file values can be found in the following files:
-    'config/rackhd_default.json'
-    'config/credentials_default.json'
-    'config/install_default.json'
+* 'config/rackhd_default.json'
+* 'config/credentials_default.json'
+* 'config/install_default.json'
 
 Stack definitions are set from the 'config/stack_config.json' file.
 An alternate configuration directory can be selected using the -config argument.
@@ -51,66 +51,60 @@ More details in config/README.md.
 
 All FIT tests can be run from the wrapper 'run_tests.py':
 
-    usage: run_tests.py [-h] [-test TEST] [-config CONFIG] [-group GROUP]
-                        [-stack STACK] [-ora ORA] [-version VERSION]
-                        [-template TEMPLATE] [-xunit] [-numvms NUMVMS] [-list]
-                        [-sku SKU] [-obmmac OBMMAC | -nodeid NODEID]
-                        [-http | -https] [-port PORT] [-v V]
+### --help output
+        usage: run_tests.py [-h] [-test TEST] [-config CONFIG] [-group GROUP]
+                            [-stack STACK] [-ora ORA] [-version VERSION]
+                            [-template TEMPLATE] [-xunit] [-numvms NUMVMS] [-list]
+                            [-sku SKU] [-obmmac OBMMAC | -nodeid NODEID]
+                            [-http | -https] [-port PORT] [-v V] [-nose-help]
 
-    Command Help
+        Command Help
 
-    optional arguments:
-      -h, --help          show this help message and exit
-      -test TEST          test to execute, default: tests/
-      -config CONFIG      config file location, default: config
-      -group GROUP        test group to execute: 'smoke', 'regression',
-                          'extended', default: 'all'
-      -stack STACK        stack label (test bed), overrides -ora
-      -ora ORA            OnRack/RackHD appliance IP address or hostname, default:
-                          localhost
-      -version VERSION    OnRack package install version, example:onrack-
-                          release-0.3.0, default: onrack-devel
-      -template TEMPLATE  path or URL link to OVA template or OnRack OVA
-      -xunit              generates xUnit XML report files
-      -numvms NUMVMS      number of virtual machines for deployment on specified
-                          stack
-      -list               generates test list only
-      -sku SKU            node SKU name, example: Quanta-T41, default=all
-      -obmmac OBMMAC      node OBM MAC address, example:00:1e:67:b1:d5:64
-      -nodeid NODEID      node identifier string of a discovered node, example:
-                          56ddcf9a8eff16614e79ec74
-      -http               forces the tests to utilize the http API protocol
-      -https              forces the tests to utilize the https API protocol
-      -port PORT          API port number override, default from
-                          install_config.json
-      -v V                Verbosity level of console output, default=1, Built Ins:
-                          0: No debug, 2: User script output, 4: rest calls and
-                          status info, 6: other common calls (ipmi, ssh), 9: all
-                          the rest
+        optional arguments:
+          -h, --help          show this help message and exit
+          -test TEST          test to execute, default: tests/
+          -config CONFIG      config file location, default: config
+          -group GROUP        test group to execute: 'smoke', 'regression',
+                              'extended', default: 'all'
+          -stack STACK        stack label (test bed), overrides -ora
+          -ora ORA            OnRack/RackHD appliance IP address or hostname, default:
+                              localhost
+          -version VERSION    OnRack package install version, example:onrack-
+                              release-0.3.0, default: onrack-devel
+          -template TEMPLATE  path or URL link to OVA template or OnRack OVA
+          -xunit              generates xUnit XML report files
+          -numvms NUMVMS      number of virtual machines for deployment on specified
+                              stack
+          -list               generates test list only
+          -sku SKU            node SKU name, example: Quanta-T41, default=all
+          -obmmac OBMMAC      node OBM MAC address, example:00:1e:67:b1:d5:64
+          -nodeid NODEID      node identifier string of a discovered node, example:
+                              56ddcf9a8eff16614e79ec74
+          -http               forces the tests to utilize the http API protocol
+          -https              forces the tests to utilize the https API protocol
+          -port PORT          API port number override, default from
+                              install_config.json
+          -v V                Verbosity level of console output, default=1, Built Ins:
+                              0: No debug, 2: Display infra.run and test.run DEBUG, 4:
+                              Display test.data (rest calls and status) DEBUG, 6:
+                              infra.data DEBUG (ipmi, ssh), 9: infra.* and test.* at
+                              DEBUG_9 (max output)
+          -nose-help          display help from underlying nosetests command,
+                              including additional log options
 
-When run_tests.py executes, a generated configuration will be created in the
-config/generated directory.  The name will appear in the run_test.py output.
 
-A previously generated configuration can be used again for a run_tests.py
-invocation by setting the environment variable FIT_CONFIG.
-
-This example will run the RackHD installer onto stack 1 via the wrapper script:
+### Example that will run the RackHD installer onto stack 1 via the wrapper script:
 
     python run_tests.py -stack 1 -test deploy/run_rackhd_installer.py
-    *** Created config file: config/generated/fit-config-20170118-160319
-    *** Using config file: config/generated/fit-config-20170118-160319
 
-The -stack or -ora argument can be omitted when running on the server or appliance. The test defaults to localhost:8080 for API calls.
-
-To run this exact test configuration again, the following could be done:
-    export FIT_CONFIG=config/generated/fit-config-20170118-160319
-    python run_tests.py
-
-This example will run the smoke test from the appliance node or the default Vagrant test bed:
+### Example will run the smoke test from the appliance node or the default Vagrant test bed:
 
     python run_tests.py -test tests -group smoke
 
+The -stack or -ora argument can be omitted when running on the server or appliance. The test defaults to localhost:8080 for API calls.
 
+
+### Example using nosetests
 Alternatively tests can be run directly from nose. Runtime parameters such as ORA address must be set in the environment.
 
 The following example will run all the entire test harness from a third party machine to ORA at 192.168.1.1:
@@ -119,7 +113,7 @@ The following example will run all the entire test harness from a third party ma
     nosetests -s tests
 
 
-## Running individual tests
+### Running individual tests
 
 Individual test scripts or tests may be executed using the following 'Nose' addressing scheme:
 
@@ -130,6 +124,24 @@ For example, to run the test 'test_rackhd11_api_catalogs' in script 'tests/rackh
 
     python run_tests.py -stack 11 -test tests/rackhd11/test_rackhd11_api_catalogs.py:rackhd11_api_catalogs.test_api_11_catalogs
 
+### Example of rerunning a test based on a previous run's configuration:
+
+When run_tests.py executes, a generated configuration will be created in the
+config/generated directory.  The name will appear in the run_test.py output.
+
+A previously generated configuration can be used again for a run_tests.py
+invocation by setting the environment variable FIT_CONFIG.
+
+This shows the config being generated:
+
+     python run_tests.py -stack 1 -test deploy/run_rackhd_installer.py
+     *** Created config file: config/generated/fit-config-20170118-160319
+     *** Using config file: config/generated/fit-config-20170118-160319
+
+This shows re-using the generated config:
+
+     export FIT_CONFIG=config/generated/fit-config-20170118-160319
+     python run_tests.py
 
 ## Running FIT tests on Vagrant RackHD
 
@@ -167,6 +179,12 @@ Use the following commands to initialize the server and run a Smoke Test:
 Note that any previously installed RackHD Vagrant boxes will prevent a new instance from running.
 Please remove any old RackHD VMs prior to executing this routine.
 
+## Hints and background for logging/debuging tests
+
+Please read 'stream_monitor/flogging/README.md' for information on the logging system.
+That file also contains a set of common "if you want this to happen, type this" at the top of the file and how
+the existing '-v' shortcut option maps to the loggers.
+
 ## Test conventions
 
 - Tests should be written using Python 'unittest' classes and methods.
@@ -189,7 +207,7 @@ The setUp and tearDown methods are useful ways to setup and clean out test-speci
     virtualenv .venv
     source .venv/bin/activate
     sudo pip install -r requirements.txt
-    
+
 ## Running the tests
 
 Run Vagrant environment
@@ -224,13 +242,13 @@ Run regression tests
 
     python run.py --config=/home/user/config.ini
 
-    
+
 ## HTTP proxy configuration:
 
 For OS installer related tests, optional HTTP proxies for accessing remote OS image repositories can be defined, see the [RackHD Configuration howto](http://rackhd.readthedocs.io/en/latest/rackhd/configuration.html?highlight=httpProxies#rackhd-configuration)
 
 ## Specifying test groups
-    
+
     To display the entire test plan and available test groups run:
 
     python run.py --show-plan
@@ -246,7 +264,7 @@ For OS installer related tests, optional HTTP proxies for accessing remote OS im
     Run only Redfish 1.0 compliant API related tests:
     python run.py --group=api-redfish-1.0
 
-## To reset the default target BMC user/password 
+## To reset the default target BMC user/password
 
     NOTE: only prompts for user/password when .passwd file is missing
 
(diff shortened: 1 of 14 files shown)


stuart-stanley:
test me please

stuart-stanley:
test this please

stuart-stanley:
test this please

stuart-stanley:
test this please

test/README.md
```diff
@@ -167,6 +201,52 @@ Use the following commands to initialize the server and run a Smoke Test:
...
+* There is a root logger that has these properties:
+    * INFO level as a gate.
+    * Handler to output to stdout.
+    * A handler with a log file named console_capture.log.
+
+The logfiles are generated per-run in 'log_output/run_<timedate-stamp>.d/' under
+this (RackHD/test) directory. In addition, 'log_output/run_last.d' is linked to
+the last test-run.
+
+The following are some common use-cases for wanting to alter logging for some reason:
```
hohene: To me, these examples should be in a separate readme or in the logging readme above.  We don't dig into any of the command line options in this top level RackHD/readme.md.

test/common/fit_common.py
```diff
@@ -267,7 +266,14 @@ def mkargs(in_args=None):
         in_args = sys.argv[1:]
 
     # command line argument parser returns cmd_args dict
-    arg_parser = argparse.ArgumentParser(description="Command Help")
+    arg_parser = argparse.ArgumentParser(
+        description="Command Help", add_help=False,
+        epilog="Note: -v/--verbosity is being deprecated. RAC-3984 is covering that work. " +
```
hohene: See comment above, please remove the notes.

test/common/fit_common.py
```diff
@@ -267,7 +266,14 @@ def mkargs(in_args=None):
...
+               "See README.md for hints on the temporary mapping of -v to the new abilities " +
+               "as well as a few hints on using them directly. " +
+               "NOTE: below is the --help output from nosetests.")
```
hohene: We discussed in the review to remove the output from nosetests when we do "run_tests.py -h", please.

test/common/fit_common.py
```diff
@@ -305,13 +311,46 @@ def mkargs(in_args=None):
...
+
+    # if 'help' was set, WE to handle it as best we can. We use argparse to
+    # display usage and arguments, and then give nose a shot at printing
+    # things out.
+    if parse_results.help:
+        arg_parser.print_help()
+        rcode = _run_nose_help()
+        sys.exit(rcode)
+
+    # now handle mapping -v to infra-logging
```
hohene: The shortcuts are nice, so want to keep that in some form, like we discussed in the review. thanks!

test/common/fit_common.py
```diff
@@ -305,13 +311,46 @@ def mkargs(in_args=None):
...
+    # we want to grab the arguments we want, and pass the rest
+    # into the nosetest invocation.
+    parse_results, other_args = arg_parser.parse_known_args(in_args)
+
+    # if 'help' was set, WE to handle it as best we can. We use argparse to
+    # display usage and arguments, and then give nose a shot at printing
+    # things out.
+    if parse_results.help:
+        arg_parser.print_help()
+        rcode = _run_nose_help()
```
hohene: same comment above, leave off nosetests --help.

test/README.md
```diff
@@ -82,19 +83,51 @@ All FIT tests can be run from the wrapper 'run_tests.py':
...
+
+        --sm-set-logger-level=('logger-name', 'level-name-or-value')
+                            Set a logger's capture threshold. Loggers are
+                            evaluated before handlers.
+        --sm-set-handler-level=('handler-name[:logger-name]', 'level-name-or-value')
+                            Set a handler's capture threshold.
+        --sm-set-combo-level=('handler-name[:logger-name]', 'level-name-or-value')
+                            Set a handler's capture threshold AND all loggers
+                            feeding it.
+        --sm-crank-file=('file-pattern', '[handler-name[:logger-name] [level-name-or-value]]')
```
stuart-stanley: Yeah, agreed. I had a hard time naming/describing this one. I'll try to reword and I like your name. And yeah, the "log level for handlers..." starts to feel like you are reading Dr. Suess.. ;)

stuart-stanley:
Changes pushed up based on all comments before this point!

test/common/fit_common.py
```diff
@@ -305,13 +307,59 @@ def mkargs(in_args=None):
...
+        sys.exit(rcode)
+
+    # now handle mapping -v to infra-logging
+    if parse_results.v >= 9:
+        # Turn them all up to 11.
+        vargs = ['--sm-set-combo-level', 'console*', 'DEBUG_9']
+    elif parse_results.v >= 6:
+        # ends up turning everything up to DEBUG (levels 2, 4, and 6)
+        vargs = ['--sm-set-combo-level', 'console*', 'DEBUG']
+    elif parse_results >= 4:
```
hohene: missing .v - parse_results.v

test/common/fit_common.py
```diff
@@ -305,13 +307,59 @@ def mkargs(in_args=None):
...
+        # infra.run and test.* to DEBUG (levels 2 and 4)
+        vargs = ['--sm-set-combo-level', 'console*:(infra.run|test.*)', 'DEBUG']
```
hohene: missing .v - parse_results.v

test/common/fit_common.py
```diff
@@ -305,13 +307,59 @@ def mkargs(in_args=None):
...
+    elif parse_results >= 2:
+        # infra and test.run to DEBUG
+        vargs = ['--sm-set-combo-level', 'console*:*.run', 'DEBUG']
```
hohene: missing .v - parse_results.v

test/common/fit_common.py
```diff
@@ -305,13 +307,59 @@ def mkargs(in_args=None):
...
+        rcode = _run_nose_help()
+        sys.exit(rcode)
+
+    # now handle mapping -v to infra-logging
+    if parse_results.v >= 9:
+        # Turn them all up to 11.
+        vargs = ['--sm-set-combo-level', 'console*', 'DEBUG_9']
+    elif parse_results.v >= 6:
+        # ends up turning everything up to DEBUG (levels 2, 4, and 6)
+        vargs = ['--sm-set-combo-level', 'console*', 'DEBUG']
```
hohene: Per the comments in the code, it seems like --sm-set-combo-level is supposed to act on both loggers and handlers at the same time.  But I can't get it to behave consistently.
This option:
      - vargs = ['--sm-set-combo-level', 'console*', 'DEBUG_9']
Sets both console and logfile output to DEBUG_9 level.
This option:
      - vargs = ['--sm-set-combo-level', 'console*', 'ERROR']
Only sets console* to ERROR, leaving logging into the files at DEBUG_5.

If consistent, I'd  expect debug to be set to either 9 in both places or 5 in both places.
To get the combined behavior, I instead do this:
    -  vargs = ['--sm-set-handler-level', 'console*:*.run', 'ERROR', '--sm-set-logger-level', '*.run', 'ERROR']
   or
     - vargs = ['--sm-set-combo-level', '*', 'ERROR']
This sets console output to error and all output into the *_run.log files to ERROR.
Sets all levels to ERROR.
Also, can we add a comment in the code what the 'handler' names are?  I only see 'console', what is expected?  I think loggers are the specific file names.... :-)

test/common/fit_common.py
```diff
@@ -305,13 +307,59 @@ def mkargs(in_args=None):
...
+    elif parse_results >= 4:
+        # infra.run and test.* to DEBUG (levels 2 and 4)
+        vargs = ['--sm-set-combo-level', 'console*:(infra.run|test.*)', 'DEBUG']
+    elif parse_results >= 2:
+        # infra and test.run to DEBUG
+        vargs = ['--sm-set-combo-level', 'console*:*.run', 'DEBUG']
+    else:
```
hohene: The default verbosity level in fit_common is a "1".    So, I think we need to use that for our default logging level when running - which would be setting Info_5 for all handlers and loggers.  Currently, I think we have INFO for console and DEBUG for log file output.

Also, to make this easier to use, I tried many more options for mapping verbosity "-v" to the logger options.  Would like to make changes similar to this to make it work closer to the current functionality.
```
   # now handle mapping -v to infra-logging
    if parse_results.v >= 9:
        # Turn them all up to 11.  debug_9
        vargs = ['--sm-set-combo-level', 'console*', 'DEBUG_9']
    elif parse_results.v >= 6:
        # ends up turning everything up to DEBUG debug_5
        vargs = ['--sm-set-combo-level', 'console*', 'DEBUG']
    elif parse_results.v >= 4:
        # infra.run and test.* to DEBUG - debug_5  (looks similar to level 6)
        vargs = ['--sm-set-combo-level', 'console*:(infra.run|test.*)', 'DEBUG']
    elif parse_results.v >= 2:
        # console and test runs to info level, add debug to the rest of the log output
        vargs = ['--sm-set-handler-level', 'console*', 'INFO', '--sm-set-logger-level', 'infra*', 'DEBUG', '--sm-set-logger-level', 'test*', 'INFO']
    elif parse_results.v >= 1:
        # Set console and logs to  INFO default (info_5)  - this should be default, no debug.
        vargs = ['--sm-set-combo-level', '*', 'INFO']
    elif parse_results.v == 0:
        # 0 = no debug and minimal output of just ERROR, CRITICAL to console and logs
        vargs = ['--sm-set-combo-level', '*', 'ERROR']
    else:
 
```

test/common/fit_common.py
```diff
@@ -305,13 +307,59 @@ def mkargs(in_args=None):
     arg_parser.add_argument("-v", default=1, type=int,
                             help="Verbosity level of console output, default=1, Built Ins: " +
                                  "0: No debug, " +
-                                 "2: User script output, " +
-                                 "4: rest calls and status info, " +
-                                 "6: other common calls (ipmi, ssh), " +
-                                 "9: all the rest ")
+                                 "2: Display infra.run and test.run DEBUG, " +
```
hohene: Update these comments to match how we end up setting the -v options below.  0 is no debug output, error and critical only.

test/common/fit_common.py
stuart-stanley: This is exactly correct: it isn't setting the loggers that feed into its level TO ERROR or DEBUG_9, but making sure a feeder-logger won't BLOCK the level you want to set on the higher one. 

I'll put a comment pointing to the readme :). Makes more sense that trying to describe it more than once...

test/common/fit_common.py
stuart-stanley: good catch. And weird, since I tested each level. It must have gotten lost in an edit since then????
Will fix!

test/common/fit_common.py
stuart-stanley: 0 and 1 will be:
   INFO(_5) for 'console' and 'console-capture' handlers, and the root logger (to which those handlers are directly attached to).
  DEBUG(_5) for infra.run, infra.data, test.run, and test.data _loggers_. Their handler levels are 'NOTSET'.

I took the additional options and dropped them into a comment in RAC-3984. I included the '0' level change in this diff, but the others actually change the content from the old -v.

Mainly, -v 2 would start showing more of the test flow on the console, but not the data yet. -v 4 added test-level data (REST calls under test, etc), and -v 6 added in ipmi and ssh (which is basically the infrastructure info).

test/common/fit_common.py
stuart-stanley: done

stuart-stanley:
changes from last round squashed and pushed up.

test/common/fit_common.py
hohene: Thanks for the updates, tested them out.  For verbosity 1 (default), I request NO debug to go to either test_run.log or test_data.log, I just want INFO.  (I'd prefer this also for infra, but compromise).  As a user,  I want to turn up the debug as I go.  To be starting with it ON at DEBUG(_5) by default,  there is not much room to dial it up.    Debug can start on verbosity 2 and on up to infinity. please. :-)

test/common/fit_common.py
```diff
@@ -304,14 +306,66 @@ def mkargs(in_args=None):
...
+                                 "4: Display test.data (rest calls and status) DEBUG, " +
+                                 "6: infra.data DEBUG (ipmi, ssh), " +
+                                 "9: infra.* and test.* at DEBUG_9 (max output) ")
+    arg_parser.add_argument("-nose-help", default=False, action="store_true", dest="nose_help",
+                            help="display help from underlying nosetests command, including additional log options")
+    # we want to grab the arguments we want, and pass the rest
+    # into the nosetest invocation.
+    parse_results, other_args = arg_parser.parse_known_args(in_args)
+
+    # if 'help' was set, WE to handle it as best we can. We use argparse to
```
johren: Typo?  "WE to handle"

test/stream-monitor/flogging/README.md
```diff
@@ -0,0 +1,254 @@
+# (FIT) Infra-structure logging component.
+
+This logging system is part of the overall 'stream_monitor' nose plugin, but can
+in theory work independently of that. This document starts with a set of examples
+for how someone working with run_test.py might accomplish different different
```
johren: Double different

stuart-stanley:
Updated with input from meeting. I left this as it's own commit instead of squashing to help folk home in on the delta from the meeting input.

@jimturnquist : I had to remove the __init__.py you put into test/stream_monitor to enable mkcheck to see the directory. That actually breaks the plugin-self-test abilities. I opened RAC-4144 to steal your work from mkcheck once you have done all the hard work :)

johren:
Approved assuming that @jimturnquist is okay with the postponing the mkcheck in this subdir.

jimturnquist:
Yep.  I'm good with postponing mkcheck in test/stream_monitor.

test/README.md
```diff
@@ -51,50 +51,54 @@ More details in config/README.md.
...
-      -http               forces the tests to utilize the http API protocol
-      -https              forces the tests to utilize the https API protocol
-      -port PORT          API port number override, default from
-                          install_config.json
-      -v V                Verbosity level of console output, default=1, Built Ins:
-                          0: No debug, 2: User script output, 4: rest calls and
-                          status info, 6: other common calls (ipmi, ssh), 9: all
-                          the rest
+### --help output
+        usage: run_tests.py [-h] [-test TEST] [-config CONFIG] [-group GROUP]
```
stuart-stanley: it wasn't doing the markdown right before... looked fine in a text editor, but was messed up elsewhere... (the ### didn't impact it)

However, I did realize that I forgot to recapture the updated output after the most recent changes. Doing that now.

test/README.md
```diff
@@ -82,19 +83,51 @@ All FIT tests can be run from the wrapper 'run_tests.py':
       -http               forces the tests to utilize the http API protocol
       -https              forces the tests to utilize the https API protocol
       -port PORT          API port number override, default from
-                          install_config.json
+                          global_config.json
```
stuart-stanley: it was. I think you were looking at an old diff?

test/common/fit_common.py
```diff
@@ -1075,6 +1123,7 @@ def _noserunner(pathspecs, noseopts):
         argv.append('--xunit-file')
         argv.append(xmlfile)
         argv.extend(pathspecs)
+        argv.extend(fitcfg()['cmd-args-list']['unhandled_arguments'])
```
stuart-stanley: I went with not-changing so I could not-retest again :)

PR #589, opened 2017-01-23
Updated deploy/rackhd_package_install.py for new configuration system

gpaulos:
Fixed package installer for new FIT configuration system.

NOTES:
lines 18-19: removed reference to deprecated -ora arguments
line 82: revision to accommodate docker deployments
lines 157-215: using generated 'config.json' from FIT config.

### DIFF
--- a/test/deploy/rackhd_package_install.py
+++ b/test/deploy/rackhd_package_install.py
@@ -15,8 +15,6 @@ This script performs the following functions:
     - startup and verify operations
 
 usage:
-    python run_tests.py -ova <ip or host> -test deploy/rackhd_package_install.py
-    or
     python run_tests.py -stack <stack ID> -test deploy/rackhd_package_install.py
 '''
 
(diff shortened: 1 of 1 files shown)


test/deploy/rackhd_package_install.py
```diff
@@ -29,9 +27,9 @@
 
 # set proxy if required
 PROXYVARS = ''
-if fit_common.fitproxy() and fit_common.fitproxy['url'] != '':
-    PROXYVARS = "export http_proxy=" + fit_common.fitproxy()['url'] + ";" + \
-              "export https_proxy=" + fit_common.fitproxy()['url'] + ";"
+if fit_common.fitproxy()['host'] != '':
+    PROXYVARS = "export http_proxy=http://" + fit_common.fitproxy()['host'] + ":" + fit_common.fitproxy()['port'] + ";" + \
+                "export https_proxy=http://" + fit_common.fitproxy()['host'] + ":" + fit_common.fitproxy()['port'] + ";"
```
johren: Should this be https://?  Or are we assuming only one proxy URL that is used for both HTTP and HTTPS?

gpaulos: This is correct. One proxy url is used for both. The https_proxy export is required for the ansible installer.

johren: Can you add a comment describing this?  I think this might confuse other developers working in the code.  They might try to change the http to https.

gpaulos:
@johren added comment about proxy settings

PR #590, opened 2017-01-23
Fix the node test after changing the "identifier" property of the nod…

tannoa2:
…e collection to unique

### DIFF
--- a/test/tests/api/v1_1/nodes_tests.py
+++ b/test/tests/api/v1_1/nodes_tests.py
@@ -29,44 +29,50 @@ class NodesTests(object):
         self.__task = None
         self.__discovery_duration = None
         self.__discovered = 0
+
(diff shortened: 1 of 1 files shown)


test/tests/api/v1_1/nodes_tests.py
```diff
@@ -29,44 +29,50 @@ def __init__(self):
...
-            },
+            'identifiers': ["FF:FF:FF:01"],
+            'autoDiscover': 'false',
+            'name': 'test_switch_node',
+            'type': 'switch',
+            'snmpSettings': {
+                'host': '1.1.1.1',
+                'community': 'rackhd'
+            }
+        },
```
anhou: looks like all indents are not matched

tannoa2:
@hohene this change won't cause any problems with older code

PR #591, opened 2017-01-23
Update deploy/rackhd_source_install.py for new FIT configuration

gpaulos:
Fixed source installer for new FIT configuration system.

NOTES:
 lines 20-21: removed reference to deprecated -ora arguments
line 127 - increased ansible installer timeout for Vagrant deployments which often time out
 line 133: revision to accommodate docker deployments
 lines 89-100: removed explicit repo list, instead harvest repo list from config in line 97
 lines 208-266: using generated 'config.json' from FIT config.

### DIFF
--- a/test/deploy/rackhd_source_install.py
+++ b/test/deploy/rackhd_source_install.py
@@ -17,8 +17,6 @@ NOTES:
        If the host is rebooted, the RackHD must be restarted by typing 'sudo nf start' at console.
 
 usage:
-    python run_tests.py -ova <ip or host> -test deploy/rackhd_source_install.py
-    or
     python run_tests.py -stack <stack ID> -test deploy/rackhd_source_install.py
 '''
 
@@ -86,18 +84,6 @@ class rackhd_source_install(fit_common.unittest.TestCase):
 
     def test02_clone_rackhd_source(self):
         print "**** Cloning RackHD source."
-        modules = [
-            "on-core",
-            "on-dhcp-proxy",
-            "on-http",
-            "on-statsd",
-            "on-syslog",
-            "on-taskgraph",
-            "on-tasks",
-            "on-tftp",
-            "on-tools",
-            "on-wss"
-        ]
         # clone base repo
         fit_common.remote_shell('rm -rf ~/rackhd')
         self.assertEqual(fit_common.remote_shell(PROXYVARS + "git clone "
@@ -108,7 +94,7 @@ class rackhd_source_install(fit_common.unittest.TestCase):
                                                  + fit_common.fitinstall()['rackhd']['branch']
                                                  )['exitcode'], 0, "Branch not found on RackHD repo.")
         # clone modules
-        for repo in modules:
+        for repo in fit_common.fitinstall().keys():
             self.assertEqual(fit_common.remote_shell(PROXYVARS
                                                     + "rm -rf ~/rackhd/" + repo + ";"
                                                     + "git clone "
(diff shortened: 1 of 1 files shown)


jimturnquist:
+1

johren:
test this please

test/deploy/rackhd_source_install.py
```diff
@@ -108,7 +94,7 @@ def test02_clone_rackhd_source(self):
                                                  + fit_common.fitinstall()['rackhd']['branch']
                                                  )['exitcode'], 0, "Branch not found on RackHD repo.")
         # clone modules
-        for repo in modules:
+        for repo in fit_common.fitinstall().keys():
```
johren: As far as I can tell, cloning these repos is unnecessary. That gets handled in the playbook. I commented out this code and it worked fine.

gpaulos: @johren  These were intended for future. I had put in some work to create a new playbook to support this deployment but that got derailed. How about I comment this out with a note.

johren: Okay, if this is something you're still planning to do.   Is there an advantage or that deployment vs. the current playbook?

gpaulos: Yes, it will have the ability to select sources from private repos and branches.

gpaulos:
@johren 
Made the following changes:
- revised system startup to use PM2, needed to revise startup logic to accommodate PM2
- commented out repo loader
- added comment to proxy section

test/deploy/rackhd_source_install.py
```diff
@@ -281,14 +213,11 @@ def test05_install_rackhd_config_files(self):
         fit_common.remote_shell('mkdir -p ~/src/on-http/static/swagger-ui')
 
     def test06_startup(self):
-        print "Start services."
-        startup = open('startup.sh', 'w')
-        startup.write('cd ~/;nf start&\n')
-        startup.close()
-        fit_common.scp_file_to_ora('startup.sh')
+        print "**** Start services."
         self.assertEqual(fit_common.remote_shell("chmod 777 startup.sh;/etc/init.d/isc-dhcp-server restart")['exitcode'], 0, "dhcp startup failure.")
```
johren: I think this command line will fail because we no longer have a startup.sh file, correct?  Have you tested with these changes?

test/deploy/rackhd_source_install.py
```diff
@@ -281,14 +213,11 @@ def test05_install_rackhd_config_files(self):
...
-        self.assertEqual(fit_common.remote_shell("nohup ./startup.sh")['exitcode'], 0, "RackHD startup failure.")
+        self.assertEqual(fit_common.remote_shell("cd ~/;pm2 start rackhd-pm2-config.yml > /dev/null 2>&1")['exitcode'], 0, "RackHD startup failure.")
         print "**** Check installation."
+        fit_common.time.sleep(10)
```
johren: Is there a reason to have a sleep here that isn't handled in the retry loop right after this?

gpaulos: Yep, it failed when the sleep was not prior to the startup loop. Looks like an anomaly of the PM2 startup routine.

test/deploy/rackhd_source_install.py
gpaulos: I should change that. It won't fail because the last command in the line generates the exit code.

PR #592, opened 2017-01-23
Update deploy/os_ova_install.py multiple VMs routine

gpaulos:
Fixed OVA installer for new FIT configuration system.
Multiple VM deployments now seed off of the 'ovamac' key in the stack config.


johren:
Please also remove the stack number check around line 114.  This is no longer necessary if we are not using that ID to compose the MAC.

jimturnquist:
@johren Do you the line number correct in your comment above.

johren:
Yep, just double-checked.  It checks to make sure the stack ID is a number if numvms > 1.  This is no longer necessary with George's change.

gpaulos:
Removed number check for stack ID and tested as requested by @johren

PR #595, opened 2017-01-24
Remove SNMP settings from node

dalebremner:
In support of:
RackHD/on-tasks#394
RackHD/on-core#252
RackHD/on-http#570


brianparry:
test this please

PR #598, opened 2017-01-24
Update deploy/rackhd_stack_init.py

gpaulos:
Several updates to deploy/rackhd_stack_init.py:

lines 34-47: resequenced 'set_auth_user' to run first. This will support the secure deployment case where http is not enabled. If 'set_auth_user' is not run first, sku packs would fail in this case.

line 107: passes assert if a 409 return code is received where the default sku is already set. This allows user to run stack init multiple times without error.

lines 229-259: fixed code in the management server setup. Still have this skipped because enabling results in a test failure in test_redfish10_api_chassis.py:redfish10_api_chassis.test_02_redfish_v1_chassis_id_links. Investigating whether failure is a test bug or product bug.

### DIFF
--- a/test/deploy/rackhd_stack_init.py
+++ b/test/deploy/rackhd_stack_init.py
@@ -31,7 +31,20 @@ import pdu_lib
 MAX_CYCLES = 60
 
 class rackhd_stack_init(unittest.TestCase):
-    def test01_preload_sku_packs(self):
+    def test01_set_auth_user(self):
+        print '**** Installing default admin user'
+        fit_common.remote_shell('rm auth.json')
+        auth_json = open('auth.json', 'w')
+        auth_json.write('{"username":"' + fit_common.fitcreds()["api"][0]["admin_user"] + '", "password":"' \
+                        + fit_common.fitcreds()["api"][0]["admin_pass"] + '", "role":"Administrator"}')
+        auth_json.close()
+        fit_common.scp_file_to_ora('auth.json')
+        rc = fit_common.remote_shell("curl -ks -X POST -H 'Content-Type:application/json' https://localhost:" \
+                                     + str(fit_common.fitports()['https']) + "/api/2.0/users -d @auth.json")
+        if rc['exitcode'] != 0:
+            print "ALERT: Auth admin user not set! Please manually set the admin user account if https access is desired."
+
+    def test02_preload_sku_packs(self):
         print "**** Processing SKU Packs"
         # Load SKU packs from GutHub
         subprocess.call("rm -rf temp.sku; rm -rf on-skupack", shell=True)
@@ -79,7 +92,7 @@ class rackhd_stack_init(unittest.TestCase):
             break
         self.assertEqual(errorcount, "", errorcount)
 
-    def test02_preload_default_sku(self):
+    def test03_preload_default_sku(self):
         # Load default SKU for unsupported compute nodes
         print '**** Installing default SKU'
         payload = {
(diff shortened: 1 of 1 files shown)


test/deploy/rackhd_stack_init.py
```diff
@@ -230,36 +230,33 @@ def test11_add_management_server(self):
         print "**** Creating management server."
         usr = ""
         pwd = ""
-        # find correct BMC passwords from global config
+        # find correct BMC passwords from credentials list
         for creds in fit_common.fitcreds()['bmc']:
```
jimturnquist: This could be simplified by using:

for usr, pwd in fit_common.fitcreds()['bmc'].iteritems():

and then just reference usr and pwd from then on.

jimturnquist: Cleared my previous comment.  bmc is a credential array not a dictionary.  Leave code as.

test/deploy/rackhd_stack_init.py
```diff
@@ -31,7 +31,20 @@
...
+    def test01_set_auth_user(self):
+        print '**** Installing default admin user'
+        fit_common.remote_shell('rm auth.json')
+        auth_json = open('auth.json', 'w')
+        auth_json.write('{"username":"' + fit_common.fitcreds()["api"][0]["admin_user"] + '", "password":"' \
+                        + fit_common.fitcreds()["api"][0]["admin_pass"] + '", "role":"Administrator"}')
+        auth_json.close()
+        fit_common.scp_file_to_ora('auth.json')
+        rc = fit_common.remote_shell("curl -ks -X POST -H 'Content-Type:application/json' https://localhost:" \
+                                     + str(fit_common.fitports()['https']) + "/api/2.0/users -d @auth.json")
```
johren: Just a suggestion:  I've seen this pattern a few times throughout the test code of creating a json file for the payload, scp-ing it to the ora and then running an API with the payload.  This would be a good candidate for a common function that takes a payload and a URL, and returns the result.

test/deploy/rackhd_stack_init.py
```diff
@@ -31,7 +31,20 @@
...
+        if rc['exitcode'] != 0:
+            print "ALERT: Auth admin user not set! Please manually set the admin user account if https access is desired."
```
johren: I would say "if authenticated access is desired".  It actually is orthogonal to http vs. https.

test/deploy/rackhd_stack_init.py
```diff
@@ -230,36 +230,33 @@ def test11_add_management_server(self):
...
+                            "config": { "host": fit_common.fitcfg()['bmc'],
+                                        "user": usr,
+                                        "password": pwd
+                                        }
+                                }
+                            ]
+                        }
+            api_data = fit_common.rackhdapi("/api/2.0/nodes", action='post', payload=payload)
+            self.assertEqual(api_data['status'], 201, 'Incorrect HTTP return code, expecting 201, got '
+                             + str(api_data['status']))
```
johren: This can be for later, but we really should have a tests here that the node actually did get created correctly.  The 201 status just tells us that the workflow got started.  Not that it completed successfully.

test/deploy/rackhd_stack_init.py
gpaulos: Yep, I will write a story for that. There are a number of opportunities for useful common functions all over.

gpaulos: @johren Yep, I will write a story for that. There are a number of opportunities for useful common functions all over.

PR #604, opened 2017-01-27
RAC-4005  Pylint cleanup in RackHD/test

jimturnquist:
- added mkcheck.sh script to perform pylint checking (and
  anything else we may eventually want to check).  This
  can be used by Jenkins as a code quality gate.
- Usage: ./mkcheck <html|text>
- pylint output is stored in rackhd_test_pylint.text or rackhd_test_pylint.html
  depending on which pkg_format is requested.  default is text.
- Only actual errors are checked for ("pylint -E").  Cleaning
  up warnings is for another day.
- Updated a handful of errors that existed presently in RackHD/test
- update pylintrc to match more modern pylint

### DIFF
--- a/test/common/api_utils.py
+++ b/test/common/api_utils.py
@@ -165,19 +165,19 @@ def api_validate_node_pollers(client, node_id_list, all_pollers=False):
         # return no good pollers
         return False
 
-    for node_ld in node_id_list:
-        Api().nodes_get_pollers_by_id(node_ld)
+    for node_id in node_id_list:
+        Api().nodes_get_pollers_by_id(node_id)
         node_poller_rsp = client.last_response
         node_poller_list = loads(node_poller_rsp.data)
         if node_poller_rsp.status != 200:
-            print '**** Unable to retrieve node: {} pollers via API. status ({})\n'.format(node.get('id'),
+            print '**** Unable to retrieve node: {} pollers via API. status ({})\n'.format(node_id,
                                                                                            node_poller_rsp.status)
             # unable to get poller for this node, so return a False
             return False
 
         # if the poller return None or an empty list, return a False
         if not node_poller_list:
-            print 'Node {} has no pollers.'.format(node_ld)
+            print 'Node {} has no pollers.'.format(node_id)
             # no poller on this node, so return a False
             return False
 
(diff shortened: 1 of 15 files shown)


test/common/test_api_utils.py
```diff
@@ -404,7 +404,7 @@ def get_supported_pollers(node_id):
 
     if mondata['status'] not in [200,201,202,204]:
         if fit_common.VERBOSITY >= 2:
-            print "Status: {},  Failed to get data for poller id: {}".format(mondata['status'], str(poller_id))
+            print "Status: {},  Failed to get poller data for node: {}".format(mondata['status'], node_id)
```
johren: Sounds picky, but since you are changing it, I think this code be confusing.  I think it should say "Failed to get pollers for node".
The API above is getting the pollers for a node.  Then there is a different API for getting the poller data for a particular poller.

test/mkcheck.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
```
johren: I think this needs a copyright.

jimturnquist: OK.  I actually thought about that.  I will add it although I'm not really sure that we are copyrighting this code.

test/common/test_api_utils.py
jimturnquist: Will fix.

test/mkcheck.sh
jimturnquist: Adding it.

jimturnquist: @johren comments have been addressed and rebased.

test/common/api_utils.py
```diff
@@ -170,7 +170,7 @@ def api_validate_node_pollers(client, node_id_list, all_pollers=False):
         node_poller_rsp = client.last_response
         node_poller_list = loads(node_poller_rsp.data)
         if node_poller_rsp.status != 200:
-            print '**** Unable to retrieve node: {} pollers via API. status ({})\n'.format(node.get('id'),
+            print '**** Unable to retrieve node: {} pollers via API. status ({})\n'.format(node_ld,
```
yyscamper: It's better to rename `node_ld` to `node_id`

yyscamper:
@jimturnquist : I suggest enabling the houndci check for the test files just like other on-xxx repository, so that the linter warning can be show immediately when PR is opened, this can avoid developer forgets the pylint check.

test/common/api_utils.py
jimturnquist: Yep, I can fix.

jimturnquist:
I updated with @yyscamper comments included.

PR #594, opened 2017-01-24
Updated Vagrantfile to conform to new FIT system

gpaulos:
Recent changes to FIT broke the Vagrant installer.

These changes fix Vagrantfile script.

NOTES:
lines 37-38: the echo at the end of the line makes sure that each line returns a zero exitcode. Vagrant stops when a non-zero exit code is detected.
line 33: added package required for Python modules

@hohene @jimturnquist @johren @larry-dean

PR #605, opened 2017-01-31
Fix proxy settings in config/install_default.json

gpaulos:
MN proxy settings inadvertently got included into the default public FIT config. This is causing the deployment scripts to fail outside of MN lab. Please review and merge quickly.

@johren @jimturnquist @hohene

### DIFF
--- a/test/config/install_default.json
+++ b/test/config/install_default.json
@@ -24,8 +24,8 @@
         "https://github.com/RackHD/on-skupack"
         ],
       "proxy": {
-        "host": "web.hwimo.lab.emc.com",
-        "port": "3128"
+        "host": "",
+        "port": ""
       }
   }
 }


johren:
@gpaulos:  Just checking on the testing.  Have you successfully installed with this config outside the MN lab?

jimturnquist:
+1

johren:
Ready to merge after above question is answered.

hohene:
The proxy entry in the install-config.json file is causing a chicken and egg issue for testing.  During the vagrant install, it does a new checkout of the rackhd/rackhd repo, thus pulling in the 'incorrect' configuration.  Then the rest of the install fails, as it tries to use the 'MN lab' proxy.  I can re-instrument the installation to stop and start again...but...being it is broken as is, this fix won't make it worse.  If you can merge, then we can retest it right away....with fixes in #594

gpaulos:
@johren The original default global_config was with the blank proxy settings which worked outside of the MN lab. When we enabled the new config files, MN proxy settings were inadvertently included which broke the Vagrant stuff. This change will not affect the merge gate which does not use the FIT deployments.

johren:
@gpaulos, my concern is just verifying that the original use case was tested.   The description states that it was "causing the deployment scripts to fail outside of MN lab" so I wanted to make sure the change was tested in that scenario.  It sounds like we are not able to test that scenario until the change goes in because of the chicken-egg problem.   Since it is a relatively simple change, I'll merge it to allow the testing.

PR #614, opened 2017-02-04
RAC-4138 - Updated the test template files to match the current framework.

hohene:
Added examples for fit_common api calls, dependencies, and base logging.
Templates can be used as a starting point for creating new fit tests.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,6 +5,4 @@ max-line-length = 128
 include = .
 exclude = .venv
 
-ignore =
-    # E402 module level import not at top of file
-    E402
+ignore = E402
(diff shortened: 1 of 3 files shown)


hohene:
@johren @larry-dean @jimturnquist @diontje

hohene:
@johren @larry-dean @jimturnquist @diontje @gpaulos @derrickostertag @RackHD/corecommitters

jimturnquist:
Lets add an "ignore" for this particular import failure.  We don't have an immediate solution for setting up sys.path with doing this.  .flake8 file should look something like:

[flake8]

max-line-length = 128

include = .
exclude = .venv

ignore =
    # E402 module level import not at top of file		
    E402

PR #615, opened 2017-02-06
Rac 4080 parallel OS tstrap test

gpaulos:
This is a new test script that runs Redfish 1.0 OS bootstrap tests in parallel. This dramatically reduces test time from several hours down to 15-20 minutes, making this test suitable for regression suite.

This test allocates available nodes to test cases then runs all of them. 8 nodes are required to run all of the existing test cases. If there are less than 8 nodes, tests will be skipped. Note that Redfish API currently only supports ESXi, CentOS, and RHEL. Tests for other OSs will be added when Redfish API is updated.

Most of the work in this script is done in the header. The test methods are just polling loops to wait for completion of the associated workflow and assert on fail.


gpaulos:
Found a problem with the script not honoring the group attribute setting. This is because the workflow launch code is outside the scope of the attr decorator. Closing the PR for revision. Will revise and repost later.

PR #616, opened 2017-02-06
Updates to RackHD/test to cleanup module pathing.

jimturnquist:
Standardized how RackHD/test finds test and test/common.  This
eliminates the need for inlined 'git-rev-parse' style statements
throughout the code.  Doing the following in an individual test
will set up the module pathing:

import fit_path

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -6,3 +6,5 @@ include = .
 exclude = .venv
 
 ignore =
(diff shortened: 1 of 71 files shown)


jimturnquist:
Just messing around with hound.  Don't merge.

PR #620, opened 2017-02-07
Rac 4080 parallel OS bootstrap test

gpaulos:
This resubmit of the new test script that runs Redfish 1.0 OS bootstrap tests in parallel. This dramatically reduces test time from several hours down to 15-20 minutes, making this test suitable for regression suite.

This test allocates available nodes to test cases then runs all of them. 8 nodes are required to run all of the existing test cases. If there are less than 8 nodes, tests will be skipped. Note that Redfish API currently only supports ESXi, CentOS, and RHEL. Tests for other OSs will be added when Redfish API is updated.

Most of the work in this script is done in the classSetUp routine. The test methods are just polling loops to wait for completion of the associated workflow and assert on fail.

### DIFF
--- a/test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
+++ b/test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
@@ -54,6 +54,9 @@ import subprocess
 sys.path.append(subprocess.check_output("git rev-parse --show-toplevel", shell=True).rstrip("\n") + "/test/common")
 import fit_common
 
+import flogging
+log = flogging.get_loggers()
+
 # This gets the list of nodes
 NODECATALOG = fit_common.node_select()
 
@@ -70,6 +73,7 @@ def wait_for_task_complete(taskid, retries=60):
     for dummy in range(0, retries):
         result = fit_common.rackhdapi(taskid)
         if result['status'] != 200:
+            log.tdl.error(" " + result['text'])
             return False
         if result['json']['TaskState'] == 'Running' or result['json']['TaskState'] == 'Pending':
             if fit_common.VERBOSITY >= 2:
@@ -78,10 +82,12 @@ def wait_for_task_complete(taskid, retries=60):
         elif result['json']['TaskState'] == 'Completed':
             if fit_common.VERBOSITY >= 2:
                 print "OS Install workflow state: {}".format(result['json']['TaskState'])
+            log.tdl.info_5(" " + result['text'])
             return True
         else:
-            break
-    print "Task failed with the following state: " + result['json']['TaskState']
+            log.tdl.error(" " + result['text'])
+            return False
+    log.tdl.error(" Workflow Timeout: " + result['text'])
     return False
 
 # helper routine for selecting OS image path by matching proxy path
@@ -90,6 +96,13 @@ def proxySelect(tag):
         if tag in entry['localPath']:
             return entry['localPath']
     return ''
+
+# helper routine to return the task ID associated with the running bootstrap workflow
+def node_taskid(osname, version):
+    for entry in status:
+        if status[entry]['os'] == osname and status[entry]['version'] == version:
+            return status[entry]['id']
+    return ""
 
 # run individual bootstrap with parameters
 def run_bootstrap_instance(node, osname, version, path):
@@ -114,8 +127,12 @@ def run_bootstrap_instance(node, osname, version, path):
                                         action='post', payload=payload_data)
     if result['status'] == 202:
         status[node] = {"os":osname, "version":version, "id":result['json']['@odata.id']}
+        log.tdl.info_5(" TaskID: " + result['text'])
+        log.tdl.info_5(" Payload: " + fit_common.json.dumps(payload_data))
     else:
         status[node] = {"os":osname, "version":version, 'id':"/redfish/v1/taskservice/tasks/failed"}
+        log.tdl.error(" TaskID: " + result['text'])
+        log.tdl.error(" Payload: " + fit_common.json.dumps(payload_data))
 
 # run all bootstraps on available nodes
 def launch_bootstraps():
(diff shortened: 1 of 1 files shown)


test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,196 @@
...
+    for item in oslist:
+        # if OS proxy entry exists in RackHD config, run bootstrap against selected node
+        if proxySelect(item['path']) and nodeindex < len(NODECATALOG):
+            #delete active workflows for specified node
+            fit_common.cancel_active_workflows(NODECATALOG[nodeindex])
+            payload_data = {
+                            "osName": item['os'],
+                            "version": item['version'],
+                            "repo": rackhdhost + proxySelect(item['path']),
+                            "rootPassword": "1234567",
```
yyscamper: ESXi password policy is more strict than other OS, see https://communities.vmware.com/docs/DOC-31941, so this rootPassword actually violates the policy, and it will eventually lead to the rootPassword doesn't been set correctly. Of course, this will not fail the workflow so you will not aware of such problem.

To fix this potential problem, I recommend you change the rootPassword to a complex one.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,196 @@
...
+                            "hostname": "rackhdnode",
+                            "dnsServers": [rackhdconfig['apiServerAddress']],
+                            "users": [{
+                                        "name": "rackhd",
+                                        "password": "R@ckHD1!",
+                                        "uid": 1010,
+                                    }]
+                           }
```
yyscamper: If the os name contains `+KVM`, then the payload_data need an additional key-value pair `kvm: true`, otherwise, there is no difference between `xxx` and `xxx+KVM`.

Of course, if the node resource is limited, to skip those os name that ends with `+KVM` is also a choice.

yyscamper:
+1 with some nice-to-have comment.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
gpaulos: Where is this documented?

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
gpaulos: The root password does work like this. Its the user password that needed complexity.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,196 @@
...
+            log.info_5(" " + result['text'])
+            return True
+        else:
+            log.error(" " + result['text'])
+            return False
+    log.error(" Workflow Timeout: " + result['text'])
+    return False
+
+# helper routine for selecting OS image path by matching proxy path
+def proxySelect(tag):
```
johren: just a nit on style:  would be more consistent for this function name to be proxy_select to match the other functions in this file.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,196 @@
...
+
+# helper routine to return the task ID associated with the running bootstrap workflow
+def node_taskid(osname, version):
+    for entry in NODE_STATUS:
+        if NODE_STATUS[entry]['os'] == osname and NODE_STATUS[entry]['version'] == version:
+            return NODE_STATUS[entry]['id']
+    return ""
+
+# run all bootstraps if regression group is specified or no group is specified
+if "regression" in sys.argv or "-a" not in sys.argv:
```
johren: Couldn't this all be part of a test that runs before the rest of the tests such that any failures in kicking off the OS installs are asserts?

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
johren: agree

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,196 @@
...
+# this routine polls a task ID for completion
+def wait_for_task_complete(taskid, retries=60):
+    for dummy in range(0, retries):
+        result = fit_common.rackhdapi(taskid)
+        if result['status'] != 200:
+            log.error(" " + result['text'])
+            return False
+        if result['json']['TaskState'] == 'Running' or result['json']['TaskState'] == 'Pending':
+            if fit_common.VERBOSITY >= 2:
+                print "OS Install workflow state: {}".format(result['json']['TaskState'])
```
gpaulos: Yes, this is Stuart's new logging system

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
gpaulos: I was waiting for this question. I tried my best to get this into a test method, but Nose sequences skips before running any tests. The skips require results from the first test, so there is a chicken-and-egg problem. If anyone has better ideas, I am open to revision.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,196 @@
...
+log = flogging.get_loggers()
+
+# This gets the list of nodes
+NODECATALOG = fit_common.node_select()
+
+# dict containing bootstrap workflow IDs and states
+NODE_STATUS = {}
+
+# download RackHD config from host
+rackhdconfig = fit_common.rackhdapi('/api/2.0/config')['json']
```
gpaulos: This gets the current config from the running system, which may not be the same as what is in the local config.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
johren: This part is not Stuart's new logging system so it should be converted to a log statement.
`if fit_common.VERBOSITY >= 2:
                print "OS Install workflow state: {}".format(result['json']['TaskState'])`

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
johren: Does nosedep @depends solve the problem by making the rest of the tests depend on the first one?  @hohene?

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,196 @@
+'''
+Copyright 2017, Dell EMC
```
johren: If you are making changes to this file, this is the Copyright from legal:
Copyright © 2017 Dell Inc. or its subsidiaries.  All Rights Reserved.

gpaulos: OK, but the copyright statement I put here was from Rod. I think the statement from legal supersedes Rod. I will make the changes you suggest.

gpaulos:
I have revised to respond to all comments above. 
@yyscamper  - Note that the "kvm"=True key is not documented in the Redfish Swagger docs. I included it in the code and it was accepted by Redfish API, so Redfish docs are out of date.

yyscamper:
@gpaulos : Sorry for my fault comment. I just checked the code, if `osName` ends with `+KVM` then Redfish API will automatically add the key-value pair `kvm: true`, so you don't have to do this in your test case. The code is https://github.com/RackHD/on-http/blob/master/lib/api/redfish-1.0/systems.js#L585

gpaulos:
test this please

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
gpaulos: Nosedep does not have conditionals like the skip decorator so it is not a solution to this problem.

stuart-stanley: There is another way to skip that I used for dynamic skipping before...
try putting the nosedep decorator on a test to sequence it, then do the check IN the test. Something like:
{code}
@before(next_test_in_list)
def test_check_and_skip(self):
    if "regression" not in sys.argv and "-a" in sys.argv:
        raise fit_common.unittest.SkipTest('I skipped this because I HATE this this...ok, maybe a better why than that')
{code}

Or, as an alternative, you could put this check into setUp in redfish_bootstrap_base() and have it skip all tests on the given condition? 

Might help, might not! (feel free to ignore if 'not')

gpaulos: Tried the strategy of putting the workflow code into setup, but skips are processed before setUp. The idea is to only skip tests that will not be run, which could be none, some, or all depending upon a number of variables. This is the most flexible solution for running in arbitrary environments and will notify the user when any specific test is skipped. I'm not happy that I have to specify the 'regression' tag twice, but that is the only hackish thing about this method.

gpaulos: @stuart-stanley  Tried the strategy of putting the workflow code into setup, but skips are processed before setUp. The idea is to only skip tests that will not be run, which could be none, some, or all depending upon a number of variables. This is the most flexible solution for running in arbitrary environments and will notify the user when any specific test is skipped. I'm not happy that I have to specify the 'regression' tag twice, but that is the only hackish thing about this method.

stuart-stanley: That's why I was suggesting raising the unittest.SkipTest from WITHIN setUp... I did this for the onserve amqp tests, where setup would probe for amqp, and if it wasn't there raise the exception then.

gpaulos: Don't understand, please send me some examples.

gpaulos: @stuart-stanley I don't quite understand. Please send some examples.

stuart-stanley: @gpaulos 
Hopefully this helps, especially now that I figured out how to actually get a code snippet in here!
```python
import unittest

class test_thing(unittest.TestCase):
    def setUp(self):
        if "regression" not in sys.argv or some_other_check_for_if_we_should_skip_all_these:
            raise unittest.SkipTest('Skipping because regression or all test not selected')
```

You might also be able to directly use attr inside setup, kind of like this: (this is totally theoretical, however. I have not tried this one yet. I have basically on the above one though)
```python
import unittest

class test_thing(unittest.TestCase):
    def setUp(self):
        if not attr(all=True, regression=True):
            raise unittest.SkipTest('Skipping because regression or all test not selected')
```

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,195 @@
...
+                # this is the failure case where there is no task ID
+                NODE_STATUS[NODECATALOG[nodeindex]] = {"os":item['os'], "version":item['version'], 'id':"/redfish/v1/taskservice/tasks/failed"}
+                log.error(" TaskID: " + result['text'])
+                log.error(" Payload: " + fit_common.json.dumps(payload_data))
+            # increment node index to run next bootstrap
+            nodeindex += 1
+
+
+# ------------------------ Tests -------------------------------------
+from nose.plugins.attrib import attr
```
gpaulos: No this was just the template format. I will change it.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
```diff
@@ -0,0 +1,195 @@
...
+        else:
+            log.error(" " + result['text'])
+            return False
+    log.error(" Workflow Timeout: " + result['text'])
+    return False
+
+# helper routine for selecting OS image path by matching proxy path
+def proxy_select(tag):
+    for entry in rackhdconfig['httpProxies']:
+        if tag in entry['localPath']:
```
gpaulos: I debated that myself. I did the partial match to allow some flexibility in naming, but if you think that's a bad idea, then I will change it.

test/tests/bootstrap/test_redfish10_os_bootstrap_parallel.py
gpaulos: Thanks. I'll give that a try

gpaulos: @stuart-stanley The actual problem is not the attribute tag. The problem is reporting. 

The idea of this test is to be as flexible as possible to run in different environments with any number of nodes and any number of OS repos available. If an OS repo is missing or a node not available to run a test, then that test method needs to be skipped, but not the whole test. The skip decorators are executed before all test classes and methods. But the skip decorators need results from the tests to activate. If I dispense with the skip decorators and just let the method fall through without an assert, then Nose will tally that as a PASS. Not acceptable. We need to use Nose tallying to return a correct account of what tests were run and skipped. This is why it needs to be designed like this. The only hackish thing is that the regression tag needs to be specified twice. 

That was the long-winded answer why your suggestion does not address the problem. 
However I may have a different solution, but that also has some hackish elements. 
More info soon.

gpaulos:

All of your comments inspired me to think outside the box and come up with a more elegant solution that meets all of our criteria. The skip decorators don't need test results, just a tally of the OS repos and nodes available. So I created a tally routine that runs in the head, then moved all of the bootstrap launch code to the classSetUp method. This removed the need for the extra regression attribute and moves all of the actual test code back into the class structure.

I put in the 'import fit_path' code to conform to the new pathing convention, but did not rebase the branch because the last time I did that, I broke the branch. So if you want to test the script, please move to the master branch, then it will run.

Also, the ESXi tests are still failing due to product bug RAC-4125.

gpaulos:
@johren Can you please review your change request this per my comments above?

johren:
test this please

johren:
Re-running tests.  Will merge after that passes.

johren:
Test this please

johren:
test this please

PR #624, opened 2017-02-08
Initial commit util/apply_obm_settings.py

gpaulos:
New utility to set OBMs on all or selected nodes using fit_common.apply_obm_settings() routine.

### DIFF
--- a/test/util/apply_obm_settings.py
+++ b/test/util/apply_obm_settings.py
@@ -1,0 +1,32 @@
+'''
+Copyright 2017, Dell EMC
+
+Author(s):
+George Paulos
+
+Script to apply OBM settings to all or selected nodes using BMC credentials in credentials_default.json
+
+Nodes or SKUs can be selected by using the '-sku' or '-nodeid' arguments at command line
+
+'''
+
+import os
+import sys
+import subprocess
+
+
+sys.path.append(subprocess.check_output("git rev-parse --show-toplevel", shell=True).rstrip("\n") + "/test/common")
+import fit_common
+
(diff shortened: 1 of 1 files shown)


test/util/apply_obm_settings.py
```diff
@@ -0,0 +1,28 @@
...
+'''
+
+import sys
+import subprocess
+sys.path.append(subprocess.check_output("git rev-parse --show-toplevel", shell=True).rstrip("\n") + "/test/common")
+import fit_common
+from nose.plugins.attrib import attr
+
+
+@attr(all=True)
```
hohene: Do we want all=True for utiilty script?

test/util/apply_obm_settings.py
```diff
@@ -0,0 +1,28 @@
...
+
+Script to apply OBM settings to all or selected nodes using BMC credentials in credentials_default.json
+
+Nodes or SKUs can be selected by using the '-sku' or '-nodeid' arguments at command line
+
+'''
+
+import sys
+import subprocess
+sys.path.append(subprocess.check_output("git rev-parse --show-toplevel", shell=True).rstrip("\n") + "/test/common")
```
hohene: Jim's PR is addressing this, fyi.

test/util/apply_obm_settings.py
gpaulos: Remember that the utility path is separate from the tests path. I think that the attr stuff can be eliminated from all of the utilities, its just not relevant. What do you think?

PR #626, opened 2017-02-13
Fix a bug in FIT test if use -rackhd_host to specify RackHD IP. 

nortonluo:
Fix a bug in FIT test if use -ora to specify RackHD IP. Now if -ora parameter has IP, it will use this IP instead of stack config file


johren:
@nortonluo We have recently enabled the flake8 linter to be run by Hound.  There are 4 Hound violations currently in this PR.   Could you please take a look at these when you get a chance and apply fixes?

test/common/fit_common.py
```diff
@@ -196,10 +196,11 @@ def apply_stack_config():
...
         mkcfg().add_from_file('stack_config.json', stack)
-        if 'ora' in fitcfg():
-            fitargs()['ora'] = fitcfg()['ora']
-        else:
-            fitargs()['ora'] = 'localhost'
+        if fitargs()['ora'] =='localhost':    
+            if 'ora' in fitcfg():
+               fitargs()['ora'] = fitcfg()['ora']
+            else:
+               fitargs()['ora'] = 'localhost'
```
johren: This else clause seems unnecessary.  It should not get to this line unless figargs()['ora'] is already 'localhost'.

hohene:
This fix is out of date, -ora has been changed to -rackhd_host in fit_common.py.
pull the latest file and see if it still needs the fix.

nortonluo:
@hohene Yes, the issue is still there in the latest code. I just rebased my git origin and updated the fix.
@johren I just correct the logic.
Also I found the https auth is not correct. The auth token should be obtained from 8443 port instead of 8080. I included this fix in this PR.

gpaulos:
Tested and approved. Thanks!

PR #600, opened 2017-01-25
RAC-1984 fixed flawed CIT test. Should expect 404 instead of []

lacarb:
This is the new CIT test for https://github.com/RackHD/on-http/pull/551

The api this CIT test is testing was broken and always returned [] no matter what it was given as input. The above PR corrects for this. The Cit test that this PR changes was built expecting [] which is wrong. The test now expects a 404 response when provided a bad id.

PR #629, opened 2017-02-13
RAC-1984 Removing failing CIT test for dependent PR

lacarb:
Relates to: https://github.com/RackHD/on-http/pull/551
Relates to: https://github.com/RackHD/RackHD/pull/600

This PR is to remove a CIT test which is causing this PR to fail - https://github.com/RackHD/on-http/pull/551

The above PR corrects a broken API. That PR won't pass because this CIT test fails. When making a PR to fix that CIT test the new test fails without the fixed API. I attempted using on-multi to address the mutual dependencies and was unsuccessful.

My hopes for this PR is to comment out the single failing test. See that Jenkins passes. Remove the test and push the PR up. Merge the API PR which behaves correctly. Then add an appropriate CIT test which is represented in this PR - https://github.com/RackHD/RackHD/pull/600

Please let me know if there's a better way forward than this.

### DIFF
--- a/test/tests/api/v2_0/nodes_tests.py
+++ b/test/tests/api/v2_0/nodes_tests.py
@@ -298,7 +298,7 @@ class NodesTests(object):
 
     	Api().nodes_get_workflow_by_id('fooey')
         resps_fooey = self.__get_data()
-    	assert_equal(len(resps_fooey), 0, message='Should be empty')
+    	#assert_equal(len(resps_fooey), 0, message='Should be empty')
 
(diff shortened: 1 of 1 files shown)


lacarb:
@RackHD/corecommitters

lacarb:
test this please

DavidjohnBlodgett:
👍 in the future, I would maybe comment out the test rather then removing it... maybe with a comment that like `TODO: re-add this test after API fix` that's just my opinion though, not a gate

lacarb:
@RackHD/corecommitters, anyone?

PR #634, opened 2017-02-20
rac-261 stream monitor base matcher logic.

stuart-stanley:
This story was pre-dojo and is way too long. It is a good checkpoint of the asynch matching engine to date, however!

Primary feature chunks in here:
* Added the stream-monitor base plugin. This is the classes that specific stream monitors will be based on.
* Added a matcher engine (stream_matchers_base.py)
** ordered and unordered groups of matchers, along with a root group. The groups are the entry point for processing "events" and do a lot of the common logic.
** a StreamMatchBase that protocol-specific matchers will subclass to implement their special sauce. Handles a lot of common logic, leaving as little as possible to be done in the subclass. Mostly implementing '_match' to try and see if an event matches or not.
** a match-batch class to tie events to the results of various matchers in the groups
* Added various matched-types in stream_matchers_results
** MatcherResult (base class)
** MatcherUnderMatch -- indicates a match fail because not enough matches were seen on the given matcher
** MatcherOverMatch -- indicaes a match fail because too many matches were seen on the given matcher
** MatcherCleanHitResult -- indicates success. Yay!
** MatcherOrderedMissMatch -- indicates out-of-order seen in an ordered group
* Created a selftest stream-monitor with a simple exact string match matcher and injector
* And since this IS TDD, there is a test_matchers.py plugin tester.

The bigger pieces of "todo" left are:
* make it possible to do the injections from a different greenlet to actually simulate async
* add a "wait_for" type operation (right now there is a "finish" call that just examines already injected events)

However, I want to get this in and start doing much smaller stories for the pieces. Also, the next step is to add the AMQP stream watcher, which pretty much means those two "todos" happen in that sequence of stories.

Note: also modified "mkcheck.sh" to pass on arguments the flake8 to allow testing of a specific file (or in this case, subdirectory)

One thing to remember is that this is the engine bits. The dev-facing "api" is very open to amendment in the next stories.

@hohene @diontje @gpaulos @RackHD/corecommitters

### DIFF
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,8 +1,11 @@
 on-skupack
 myenv_*
+.vscode
 log_output
 startup.sh
 nohup.out
 *~
 auth.json
 testfile
+rackhd_test_check.txt
+fit_path/fit_test_pathing.egg-info/
--- a/test/mkcheck.sh
+++ b/test/mkcheck.sh
@@ -23,7 +23,7 @@ if [ -f ${output_file} ]; then
 fi
 
 # make changes to flake8 parameters in .flake8 file
-flake8 --config=../.flake8 --statistics --tee --output-file=${output_file}
+flake8 --config=../.flake8 --statistics --tee --output-file=${output_file} $@
 if [ $? -ne 0 ]; then
     if [ "${format}" == "text" ]; then
         cat ${output_file}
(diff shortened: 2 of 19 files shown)


stuart-stanley:
test this please

PR #635, opened 2017-02-21
Adding back lines for CIT test

lacarb:
Relates to RAC-1984, https://github.com/RackHD/on-http/pull/551

Small PR returning a line of code that was removed from a test by mistake.

I made a mistake with a previous PR which removed three lines from this CIT test. The line added to the CIT test by this PR was originally in this test and exposes data which is used in later tests.

### DIFF
--- a/test/tests/api/v2_0/nodes_tests.py
+++ b/test/tests/api/v2_0/nodes_tests.py
@@ -283,6 +283,7 @@ class NodesTests(object):
                 resps.append(self.__get_data())
         for resp in resps:
             assert_not_equal(0, len(resp), message='No Workflows found for Node')
+        Api().nodes_get_workflow_by_id('fooey')
 
     @test(groups=['node_post_workflows-api2'], depends_on_groups=['node_workflows-api2'])
     def test_node_workflows_post(self):


lacarb:
test this please

lacarb:
@RackHD/corecommitters

DavidjohnBlodgett:
perfect, thanks 👍

lacarb:
@RackHD/corecommitters, can someone merge this?

test/tests/api/v2_0/nodes_tests.py
```diff
@@ -283,6 +283,7 @@ def test_node_workflows_get(self):
                 resps.append(self.__get_data())
         for resp in resps:
             assert_not_equal(0, len(resp), message='No Workflows found for Node')
+        Api().nodes_get_workflow_by_id('fooey')
```
yyscamper: @lacarb : Since no assertion after the API calling, so my understanding is that this line can be directly removed, not need to recover it.

lacarb: @yyscamper you can see on lines 274 and 288 that the POST test depends on the GET test. The testing framework being used passes data from one test to the next so the line above is being used in POST.

Also, I'll need that line anyway when I add in a correct CIT test for the fixed api. Possible test can be seen here: https://github.com/RackHD/RackHD/pull/600

I'd test all of this locally and the CIT tests fail on cataloging and don't get to these tests. I'd like a better way to move forward on this and I'm open to suggestions. In the meantime I find this to be simple and harmless.

PR #637, opened 2017-02-22
Update Ubuntu install ISO version. the old versions are not available  any more

panpan0000:
Hmm... the original iso link is no longer valid...
need to find a long term way how to resolve it..

@changev @PengTian0   @yyscamper  @anhou


changev:
:+1:

PR #639, opened 2017-02-22
Add test for amqp message node discovery

nortonluo:
Add test for amqp message node discovery. This test will try remove one node and reboot it to generate a series of AMQP messages. Then it will wait for node get discovered and get the AMQP message.
 WIP

### DIFF
--- a/test/tests/amqp/test_amqp_node_rediscover.py
+++ b/test/tests/amqp/test_amqp_node_rediscover.py
@@ -9,11 +9,15 @@ Norton Luo
 from time import sleep
 import Queue
 import random
+import socket
 import flogging
 import logging
 import pika
 import unittest
+import json
 import threading
+from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+from optparse import OptionParser
 import fit_common
 import test_api_utils
 from nose.plugins.attrib import attr
@@ -22,6 +26,8 @@ amqp_message_received = False
 amqp_queue = Queue.Queue(maxsize=0)
 node_uuid = ""
 nodefound_id = ""
+ebhook_received = False
+webhook_body = ""
 
 
 class AmqpWorker(threading.Thread):
@@ -85,6 +91,33 @@ class AmqpWorker(threading.Thread):
         self.channel.start_consuming()
 
 
+class RequestHandler(BaseHTTPRequestHandler):
+    def do_POST(self):
+        request_headers = self.headers
+        content_length = request_headers.getheaders('content-length')
+        length = int(content_length[0]) if content_length else 0
+        global webhook_received, webhook_body
+        webhook_received = True
+        webhook_body = str(self.rfile.read(length))
+        logs.debug(webhook_body)
+        self.send_response(200)
+
+
+class HttpWorker(threading.Thread):
+    def __init__(self, port, timeout=10):
+        threading.Thread.__init__(self)
+        self.server = HTTPServer(('', port), RequestHandler)
+        self.server.timeout = timeout
+
+    def dispose(self):
+        logs.debug('http service shutdown')
+        self.thread_stop = True
+
+    def run(self):
+        self.server.handle_request()
+        self.dispose()
+
+
 # Check if the node is rediscovered. Retry in every 30 seconds and total
 # 600 seconds.
 @attr(all=True, regression=False, smoke=False)
@@ -113,6 +146,30 @@ class test_node_rediscover_amqp_message(unittest.TestCase):
             timecount = timecount + 1
         logs.debug_2("Wait to rediscover Timeout!")
         return False
+
+    def _set_web_hook(self, ip, port):
+        mondata = fit_common.rackhdapi('/api/current/hooks')
+        self.assertTrue(
+            mondata['status'] < 209,
+            'Incorrect HTTP return code, could not check hooks. expected<209, got:' + str(
+                mondata['status']))
+        hookurl = "http://" + str(ip) + ":" + str(port)
+        for hooks in mondata['json']:
+            if hooks['url'] == hookurl:
+                logs.debug("hook URL alread exist the in RackHD")
+                return
+        response = fit_common.rackhdapi(
+            '/api/current/hooks',
+            action='post',
+            payload={
+                "name": "FITdiscovery",
+                "url": hookurl,
+                "filters": [{"type": "node",
+                            "action": "discovered"}]})
+        self.assertTrue(
+            response['status'] < 209,
+            'Incorrect HTTP return code, expected<209, got:' + str(
+                response['status']))
 
     def _apply_obmsetting_to_node(self, nodeid):
         # usr = ''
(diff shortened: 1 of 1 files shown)


test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,412 @@
...
+                        nodefound_id = check_message[1]
+                        return True
+            sleep(1)
+            timecount = timecount + 1
+        logs.debug_2("Wait to rediscover Timeout!")
+        return False
+
+    def _apply_obmsetting_to_node(self, nodeid):
+        usr = ''
+        pwd = ''
```
sunnyqianzhang: Do we have to declare these variable (usr and pwd) before using them?

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,412 @@
...
+            self.assertNotEquals(
+                amqp_body_json['data'],
+                {},
+                "data field is empty!")
+        except ValueError as e:
+            self.fail(
+                "FAILURE - expected key is missing in the AMQP message!{0}".format(e))
+
+    def test_rediscover(self):
+        node_collection = test_api_utils.get_node_list_by_type("compute")
```
sunnyqianzhang: Add one line like this: "asssert node_collection", in case we get an empty node_collection and the case is not running at all.

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,412 @@
...
+            amqp_message_discover)
+        logs.debug_3("wait for accessbile!")
+        self._wait_amqp_message(100)
+        amqp_message_discover = amqp_queue.get()
+        self._process_message(
+            "accessible",
+            nodefound_id,
+            nodefound_id,
+            "node",
+            amqp_message_discover)
```
sunnyqianzhang: Add the following code to cover RI-48:
amqp_body = amqp_message_discover[1];
assert  "nodeId" in amqp_body["data"], "nodeId is not contained in the discover message"
assert  amqp_body["data"]["nodeId"], "nodeId generated in discovery doesn't include valid data "
assert  "ipMacAddresses" in amqp_body["data"], "ipMacAddresses is not contained in the discover message"
assert amqp_body["data"]["ipMacAddresses"], " ipMacAddresses generated during node discovery doesn't include valid data ",

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,412 @@
...
+        expected_key = messagetype + "." + action + \
+            ".information." + typeid + "." + nodeid
+        expected_payload = {
+            "type": messagetype,
+            "action": action,
+            "typeId": typeid,
+            "nodeId": nodeid,
+            "severity": "information",
+            "version": "1.0",
+            "createdAt": {}}
```
sunnyqianzhang: Remove "createdAt" here, as it isn't used in the compare

nortonluo:
test this please

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,537 @@
...
+                        'status'] < 209, 'Was expecting response code < 209. Got ' + str(result['status']))
+        self._wait_amqp_message(10)
+        amqp_message = amqp_queue.get()
+        self._process_message("removed", nodeid, nodeid, "node", amqp_message)
+        td.dispose()
+
+        # start discovery
+        logs.debug_2("Waiting node reboot and boot into microkernel........")
+        myip = self._get_tester_ip()
+        self._set_web_hook(myip, 9889)
```
anhou: 9889 is hardcoded, please use a configurable variable

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,537 @@
...
+        # Reboot the node to begin rediscover.
+        logs.debug('launch AMQP thread')
+        reset_worker = AmqpWorker(
+            exchange_name="on.events", topic_routing_key="graph.#." + nodeid,
+            external_callback=self.amqp_callback, timeout=100)
+        reset_worker.setDaemon(True)
+        reset_worker.start()
+        # Reboot the node, wait reboot workflow start message.
+        response = fit_common.rackhdapi(
+            '/redfish/v1/Systems/' +
```
anhou: I see many unnecessary line break, please correct it unless it's too long

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,537 @@
...
+import threading
+from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+import fit_common
+import test_api_utils
+from nose.plugins.attrib import attr
+logs = flogging.get_loggers()
+amqp_message_received = False
+amqp_queue = Queue.Queue(maxsize=0)
+node_uuid = ""
+nodefound_id = ""
```
anhou: please avoid use so many global variables unless it has to

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,537 @@
...
+ebhook_received = False
+webhook_body = ""
+
+
+class AmqpWorker(threading.Thread):
+    '''
+    This AMQP worker Class will creat another thread when initialized and runs asynchronously.
```
anhou: I know that you've synced with other guys that it will have refactoring works in this parts in future, please also have a note 'TODO' in annotation for the similar parts so that the developers won't be confused about the duplicated codes.

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,537 @@
...
+    def _get_tester_ip(self):
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        ip = fit_common.fitargs()['rackhd_host']
+        logs.debug("pinging " + ip)
+        s.connect((ip, 0))
+        logs.debug('My IP address is: ' + s.getsockname()[0])
+        return str(s.getsockname()[0])
+
+    def _process_web_message(self, timeout):
+        global webhook_received, webhook_body
```
anhou: try to avoid use global variables unless it has to

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,537 @@
...
+        timecount = 0
+        while webhook_received is False and timecount < timeout:
+            sleep(1)
+            timecount = timecount + 1
+        self.assertNotEquals(
```
anhou: don't need line break, it's not easy to code review for these unnecessary line breaks

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -0,0 +1,537 @@
...
+                "createdAt field is empty!")
+            self.assertNotEquals(
+                amqp_body_json['data'],
+                {},
+                "data field is empty!")
+        except ValueError as e:
+            self.fail(
+                "FAILURE - expected key is missing in the AMQP message!{0}".format(e))
+
+    def test_rediscover(self):
```
anhou: this function is too long, I didn't get the points to review. please break this function to proper functions

PR #641, opened 2017-02-22
rac-44440-unicode-in-test-run-arguments

stuart-stanley:
This change tries to prevent the annoyingly common situation of cutting and pasting something that replaces a '-' with a unicode en-dash. The code does:
* tries to substitute certain known culprits.
* blows up if there are still unicode characters that won't convert to ascii when the substitutions are done.

There are probably more we could substitute, but this at least stops you from spending all day trying to figure out why something is subtely not working!

Removed a few imports from fit_common as well (They were not being used by something importing fit_common either!)
Added a NOQA: imported but unused for "unittest", which IS used by a ton of stuff via the fit_common import.

@RackHD/corecommitters @jimturnquist @gpaulos @hohene @diontje

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -15,16 +15,12 @@ import sys
 import json
 import subprocess
 import time
-import datetime
-import unittest
-import signal
+import unittest   # NOQA: imported but unused
 import re
 import requests
 import pexpect
-import shutil
 import inspect
 
-import nose
 import argparse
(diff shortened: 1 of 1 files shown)


stuart-stanley:
@brianparry -- as the core-comitter-I-know, this has two reviewers and is ready to merge under the test-subdir. I you would not mind pressing the big green button when you have the chance :)

PR #642, opened 2017-02-23
[skip ci]config in kibana docker image has been moved location

changev:
config file of kibana docker image file has been moved from ```/opt/kibana/config/kibana.yml``` to ```/etc/kibana/kibana.yml```

reference: https://hub.docker.com/_/kibana/

@panpan0000 @PengTian0


panpan0000:
👍

PR #645, opened 2017-02-23
Add test for heartbeat AMQP message

nortonluo:
Adding test to monitor AMQP heartbeat message.

### DIFF
--- a/test/tests/amqp/test_amqp_heartbeat.py
+++ b/test/tests/amqp/test_amqp_heartbeat.py
@@ -6,14 +6,11 @@ Norton Luo
 
 '''
 from time import sleep
-from datetime import *
 import threading
 import pika
 import logging
 import flogging
 import fit_common
-import test_api_utils
-import socket
 from nose.plugins.attrib import attr
 
 amqp_message_received = False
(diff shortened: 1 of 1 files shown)


nortonluo:
@sunnyqianzhang @yyscamper

test/tests/amqp/test_amqp_heartbeat.py
```diff
@@ -0,0 +1,179 @@
...
+        logs.debug_7('Pika connection timeout')
+        self.connection.close()
+        exit(0)
+
+    def run(self):
+        logs.debug_7('start consuming')
+        self.channel.start_consuming()
+
+
+@attr(all=True, regression=True, smoke=True)
```
sunnyqianzhang: I'd suggest we set the smoke to false since this test case hasn't been proven to be stable enough to be put into smoke test as a PR gate. Would it be better if we  check in the code first, monitor the result for a while, refactor it when the new amqp testing mechanism is done and set the smoke tag after all these are done?

nortonluo:
test this please

nortonluo:
test this please

sunnyqianzhang:
test this please

nortonluo:
test this please

PR #646, opened 2017-02-23
Rac 4235 fix rackhd source install.py

gpaulos:
This PR addresses technical debt in deploy/rackhd_source_install.py.

◾PM2 startup sometimes fails due to short timeout, removed API polling loop that was causing error
◾removed all IP addresses, harvest and calculate all IP addresses and netmasks from rackhd_config
◾convert function call scp_file_to_ora > scp_file_to_host
◾use FIT logging instead of print statements

@hohene @larry-dean @jimturnquist @stuart-stanley @RackHD/corecommitters

### DIFF
--- a/test/deploy/rackhd_source_install.py
+++ b/test/deploy/rackhd_source_install.py
@@ -73,9 +73,9 @@ class rackhd_source_install(fit_common.unittest.TestCase):
         self.assertEqual(fit_common.remote_shell("git config --global http.sslverify false")
                          ['exitcode'], 0, "Git config failure.")
         if fit_common.fitproxy()['host'] != '':
-            self.assertEqual(fit_common.remote_shell("git config --global http.proxy http://"
-                             + fit_common.fitproxy()['host'] + ':'
-                             + fit_common.fitproxy()['port'])
+            self.assertEqual(fit_common.remote_shell("git config --global http.proxy http://" +
+                             fit_common.fitproxy()['host'] + ':' +
+                             fit_common.fitproxy()['port'])
                              ['exitcode'], 0, "Git proxy config failure.")
         # install Ansible
         self.assertEqual(fit_common.remote_shell(PROXYVARS + "apt-get -y update")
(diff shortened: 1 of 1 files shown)


gpaulos:
@stuart-stanley I understand you concerns about IP processing. It is not resistant to bad input, but putting in a lot of validation code could be more error-prone than just assuming that the inputs are valid.

stuart-stanley:
There are some libs that would help, but I think that would be a next-step story rather than messing with this... So, all good :)

gpaulos:
Pre-checked code for no Hound violations. For some reason, automated Hound check is stalled.

johren:
test this please

test/deploy/rackhd_source_install.py
```diff
@@ -34,197 +34,211 @@
...
+        # create DHCP config
+        fit_common.remote_shell('echo INTERFACES=' + ifslist[1] + ' > /etc/default/isc-dhcp-server')
+        # calculate control LAN IP configuration
+        ipsplit = fit_common.fitrackhd()['dhcpGateway'].split(".")
+        ip_prefix = ipsplit[0] + '.' + ipsplit[1] + '.' + ipsplit[2] + '.'
+        masksplit = fit_common.fitrackhd()['dhcpSubnetMask'].split(".")
+        dhcp_high = \
+            str(int(ipsplit[0]) + (255 - int(masksplit[0]))) + '.' + \
+            str(int(ipsplit[1]) + (255 - int(masksplit[1]))) + '.' + \
+            str(int(ipsplit[2]) + (255 - int(masksplit[2]))) + '.' + '254'
```
johren: We could at least add a catch on each octet to make sure it is between 0 and 255?  And exit the install if it isn't?  If someone enters a subnet of 255.255.0.0, the install will succeed, but the DHCP config will be very invalid.

larry-dean: For what it's worth, monorail.py uses the socket library to validate ip addresses.  It uses  socket.inet_aton(ip) to "convert an IPv4 address from dotted-quad string format (for example, ‘123.45.67.89’) to 32-bit packed binary format, as a string four characters in length".  It will generate a  "socket.error" if something is incorrect.

stuart-stanley:
There was already a discussion on this a ways back... with the idea being to add another story to follow up and make this cidd-safe. I have sample code that will do this at home (I think)... I need to check tonight to see...

johren:
Yeah, I agree about making the algorithm more robust in a later story.  However, before we put this code in, I'd like it to at least do a simple check so that it catches the case when it is incorrect rather than failing silently.  There are a few simple ways that could be one.  Larry pointed out one of them.

gpaulos:
stuart-stanley  Yes please share sample code.

stuart-stanley:
@gpaulos I'll send a tgz file via slack...

gpaulos:
Combined and resubmitted in https://github.com/RackHD/RackHD/pull/671

PR #647, opened 2017-02-23
RAC-4326-fix-rackhd_package_install.py

gpaulos:
This PR addresses technical debt in deploy/rackhd_package_install.py.

◾removed all IP addresses, harvest and calculate all IP addresses and netmasks from rackhd_config
◾convert function call scp_file_to_ora > scp_file_to_host
◾use FIT logging instead of print statements

@hohene @larry-dean @jimturnquist @stuart-stanley @RackHD/corecommitters


gpaulos:
Combined and resubmitted in https://github.com/RackHD/RackHD/pull/671

PR #652, opened 2017-03-01
RAC-4289_FIT_https_port_bug

gpaulos:
'fit_common' fixed for two bugs:

- When rackhd_host is localhost it was always setting port to 8080. This is old code from the pre 'fit-config' days. Just removed that code.
- -http/-https arguments to force protocol were not working, Just changed argument check from string to Boolean.

@nortonluo @hohene @jimturnquist @johren

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -217,12 +217,12 @@ def add_globals():
     if fitargs()['port'] != "None":
         API_PORT = fitargs()['port']
 
-    if fitargs()['http'] == True:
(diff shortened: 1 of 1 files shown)


gpaulos:
For some reason PR gate has a default port set of 9090. This can be fixed by putting argument -port 8080 on the command line that launches the FIT test.

gpaulos:
test this please

gpaulos:
Going to do more surgery and resubmit

PR #654, opened 2017-03-02
RAC-4289 FIT https port/protocol bug

gpaulos:
'fit_common.py' fixed for port and protocol selection bugs:

- When rackhd_host is localhost it was always setting port to 8080. This is old code from the pre 'fit-config' days. I removed that code. Now the default port setting is pulled from the install_default.json config file.
- -http/-https arguments to force protocol were not working because the properties were changed to Boolean. 
- I removed the automatic protocol selection code in rackhdapi(), which was an OnRack leftover that does not work any more in the RackHD world. This code was supposed to check to see if the http endpoints are available, if not then select https. The default protocol is now set to http. If https is desired, then use the -https parameter on the command line. I think that this is simpler and less error-prone than the automatic method that can be difficult to control and diagnose.
- there were conditions where the automatic auth key generator was not working because it was always using https (Onrack leftover). Made the auth key generator use selected protocol.

This is the way it works now:

- If no command-line overrides are used, the default protocol is http and the default http port is set from install_default.json in the ports section.
- If -https option is used, then https is active protocol and https port is set from install_default.json in the ports section.
- if -port override option is used, then this is the port used for the selected protocol, whether http or https

I hope this meets everybody's needs. 

@nortonluo @hohene @jimturnquist @johren

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -214,18 +214,19 @@ def add_globals():
     global VERBOSITY
 
     # set api port and protocol from command line
-    if fitargs()['port'] != "None":
-        API_PORT = fitargs()['port']
 
     if fitargs()['http'] is True:
         API_PROTOCOL = "http"
-        if API_PORT == "None":
-            API_PORT = fitports()['http']
-
-    if fitargs()['https'] is True:
+        API_PORT = str(fitports()['http'])
+    elif fitargs()['https'] is True:
         API_PROTOCOL = "https"
-        if API_PORT == "None":
-            API_PORT = fitports()['https']
+        API_PORT = str(fitports()['https'])
+    else:  # default protocol is http
+        API_PROTOCOL = "http"
+        API_PORT = str(fitports()['http'])
+
+    if fitargs()['port'] != "None":  # port override via command line -port
+        API_PORT = fitargs()['port']
 
     # add globals section to base configuration
     TEST_PATH = fit_path.fit_path_root + '/'
@@ -240,7 +241,7 @@ def add_globals():
         }
     })
 
-    # set OVA template from command line
+    # set OVA template from command line -template
     if fitargs()["template"] == "None":
         fitargs()["template"] = fitcfg()['install-config']['template']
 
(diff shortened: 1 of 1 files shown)


test/common/fit_common.py
```diff
@@ -214,24 +214,19 @@ def add_globals():
...
-        API_PORT = fitargs()['port']
 
-    if fitargs()['http'] == "True":
+    if fitargs()['http'] is True:
         API_PROTOCOL = "http"
-        if API_PORT == "None":
-            API_PORT = fitports()['http']
-
-    if fitargs()['https'] == "True":
+        API_PORT = str(fitports()['http'])
```
anhou: 'http' should be 'https'?

test/common/fit_common.py
```diff
@@ -214,24 +214,19 @@ def add_globals():
...
+    elif fitargs()['https'] is True:
         API_PROTOCOL = "https"
-        if API_PORT == "None":
-            API_PORT = fitports()['https']
+        API_PORT = str(fitports()['https'])
+    else:  # default protocol is http
+        API_PROTOCOL = "http"
```
anhou: duplicate logic with 'if'

test/common/fit_common.py
```diff
@@ -246,7 +241,7 @@ def add_globals():
         }
     })
 
-    # set OVA template from command line
+    # set OVA template from command line -template
```
anhou: from command line paramter '-template' is better?

test/common/fit_common.py
```diff
@@ -559,17 +554,17 @@ def get_auth_token():
...
     try:
-        restful("https://" + fitargs()['rackhd_host'] + ":" + str(API_PORT) +
+        restful(API_PROTOCOL + "://" + fitargs()['rackhd_host'] + ":" + str(API_PORT) +
                 "/login", rest_action="post", rest_payload=api_login, rest_timeout=2)
     except:
         AUTH_TOKEN = "Unavailable"
         return False
     else:
-        api_data = restful("https://" + fitargs()['rackhd_host'] + ":" + str(API_PORT) +
+        api_data = restful(API_PROTOCOL + "://" + fitargs()['rackhd_host'] + ":" + str(API_PORT) +
```
nortonluo: I had another PR "https://github.com/RackHD/RackHD/pull/626" to set get auth token by https only. If 8080 http auth is note enabled in /opt/monorail/config.json , it will failed to get auth token by http.

gpaulos:
Resubmitted here:
https://github.com/RackHD/RackHD/pull/657

PR #656, opened 2017-03-04
Revert "use newly build rackhd/files image instead of download directly" 

keedya:
This reverts commit 1d8b2c5c40d121336644678298727741c9f42edd.

Add micro docker dependency images to DockerFile in files

### DIFF
--- a/docker/docker-compose-centos.yml
+++ b/docker/docker-compose-centos.yml
@@ -15,6 +15,7 @@ services:
 
   files:
     image: rackhd/files:${TAG}
+    build: "./files"
     network_mode: "host"
     privileged: true
     volumes:
--- a/docker/docker-compose-mini.yml
+++ b/docker/docker-compose-mini.yml
@@ -30,6 +30,7 @@ services:
 
   files:
     image: rackhd/files:${TAG}
+    build: "./files"
     network_mode: "host"
     privileged: true
     volumes:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -29,6 +29,7 @@ services:
       - "./dhcp/defaults:/etc/defaults"
 
   files:
+    build: "./files"
     image: rackhd/files:${TAG}
     network_mode: "host"
     privileged: true
--- a/docker/files/Dockerfile
+++ b/docker/files/Dockerfile
@@ -1,0 +1,41 @@
+FROM alpine:latest
+
+RUN mkdir -p /RackHD/downloads
+
+ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail-efi32-snponly.efi \
+    /RackHD/downloads/monorail-efi32-snponly.efi
+ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail-efi64-snponly.efi \
+    /RackHD/downloads/monorail-efi64-snponly.efi
+ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail.ipxe \
+    /RackHD/downloads/monorail.ipxe
+ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail-undionly.kpxe \
+    /RackHD/downloads/monorail-undionly.kpxe
+ADD https://bintray.com/artifact/download/rackhd/binary/ipxe/monorail.intel.ipxe \
+    /RackHD/downloads/monorail.intel.ipxe
+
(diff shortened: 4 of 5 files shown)


keedya:
@RackHD/corecommitters @geoff-reid @tannoa2 @uppalk1

keedya:
test this please

anhou:
@changev to have a review

keedya:
@changev sorry I forgot to tag you here. Thanks @anhou

changev:
@keedya 
-1 for me
This revert is to keep the docker images latest valid, am I right?
But this isn't the root solution and will make daily build and release build unavailable to users.

Keep rackhd/files:latest with the latest master is the root solution IMO. 
But dockerhub can't do this, so I'm setting a jenkins job to monitor on-imagebuilder and update rackhd/files:latest.
Thanks!.

I'll put up the job url later.

keedya:
@changev I totally disagree with the above. U are handling the update of a Docker image in a private jenkins Job and hiding the source of this image from the user.
As a user, i would like to have visibility over this image. on-image-builder should be building and updating the images with the latest on post merge. If this is broken then this what we need to fix IMO.

keedya:
@changev one last comment; **daily build and release build unavailable to users** if u need any modification for our CI system, u can customize your own docker-compose.yml and maintain the files here for user use only and make sure rackhd setup is user friendly

changev:
@keedya 
Thanks for your commets, about
1. "Current solution decrease the visibility" I partially agree, in dockerhub users can't see this image build details, this is disadvantage that different from other repo. But we have release page and rackhd/docker/readme, dockerhub page is not the first suggested entry for user. But "decreased visibility" in docker is truly a disadvantage.

The old solution(you are reverting) can show where this image comes from(bintray files) if users take a look at yml file and rackhd/files Dockerfile.
But in the current solution, if users take a look at the at yml file and on-imagebuilder Dockerfile, they also can know details of this image.

2. "on-image-builder should be building and updating the images with the latest on post merge"
We have build/release strategy, so I don't think users always want the latest static files.  Most of the time they want the static files that support their specific versions' on-xxx image .

3. "customize rackhd/files" I think this is the **advantage of current solution** but not the old one.
Current solution teach users how to build rackhd/files from on-imagebuilder source code, this is "root customize". The old rackhd/files has no relevance with on-imagebuilder explicitly, and in bintray there were only latest static files, how can users to customize rackhd/files?

----------------------------------------------
The root cause is that on-imagebuilder static files can't be built in dockerhub. 
Current solution is not perfect and we haven't perfect solution so far, it's just a results after **weigh the cross**. Compare to limit visibility loss, version control is much more important. 
**After this revert, users will be unable to get rackhd/files of specific version, but just latest**  no matter what other versions on-xxx image they are using. That will be a disaster.
More details in https://rackhd.atlassian.net/browse/RAC-4356, and for solving this issue, we've discussed with cc, version control is the key point, we also wrote detailed docs to increase the visibility.

I agree to increase the visibility and usability of rackhd/files, but version control mustn't be dropped.

keedya:
@changev thanks for your comment. Let us leave the decision to core committers to decide.
I strongly think Dockerfile for rackhd files should remain in source control and loads images from bintray. Docker users in rackhd loved the simplicity of the setup and I prefer to keep it this way.

keedya:
@RackHD/corecommitters

brianparry:
@changev @keedya suggested to me that we leave the dockerfile in docker/files but add the capability to parse the bintray index and build from the latest images.  This seems like it would be a good compromise that would allow us to retain the usability of the Dockerfile without the added complexity of updating it every time a version is updated in bintray.  @changev do you agree?

changev:
@brianparry 
I get your point, do you and @keedya  mean:
* For rackhd/files:latest, don't maintain it in dockerhub, but keep a Dockerfile in  docker/files to build it when docker-compose up. Then we won't need to update rackhd/files:latest every time(now by jenkins job) when on-imagebuilder updates.

This is fine, current docs needn't to be modified too, because we have *-dev.yml now.

The only problem is that:
* Now we are reducing the times of downloading static files from bintray directly. Bintray has data exchange limitation and it's quite expensive. For example, **Old PR Gate** used to download  static files from bintray directly and it exhausted almost the amounts. So we setup a local static file server to solve this problem. If not do this the new PR Gate(add 3 os install tests) will exhaust all data  exchange amounts.

If this is not a problem you think, **+1 for this PR**.

keedya:
@brianparry @changev  I will close this PR to avoid any further issues with the pipeline and try to introduce again in the future once the release pipeline is stabilized

PR #657, opened 2017-03-06
RAC-4289 FIT port/protocol selection bugs

gpaulos:
Resubmit of RAC-4289 with review changes:
•When rackhd_host is localhost it was always setting port to 8080. This is old code from the pre 'fit-config' days. I removed that code. Now the default port setting is pulled from the install_default.json config file.
•-http/-https arguments to force protocol were not working because of type mismatch from string to Boolean.
•I removed the automatic protocol selection code in rackhdapi(), which was an OnRack leftover that does not work any more in the RackHD world. This code was supposed to check to see if the http endpoints are available, if not then select https. The default protocol is now set to http. If https is desired, then use the -https parameter on the command line. I think that this is simpler and less error-prone than the automatic method that can be difficult to control and diagnose.

This is the way it works now:
•If no command-line overrides are used, the default protocol is http and the default http port is set from install_default.json in the ports section.
•If -https option is used, then https is active protocol and https port is set from install_default.json in the ports section.
•if -port override option is used, then this is the port used for the selected protocol, whether http or https

I hope this meets everybody's needs.

@nortonluo @hohene @jimturnquist @johren @anhou

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -212,24 +212,18 @@ def add_globals():
     global VERBOSITY
 
     # set api port and protocol from command line
-    if fitargs()['port'] != "None":
-        API_PORT = fitargs()['port']
-
-    if fitargs()['http'] == "True":
+    if fitargs()['http'] is True:
         API_PROTOCOL = "http"
-        if API_PORT == "None":
-            API_PORT = fitports()['http']
-
-    if fitargs()['https'] == "True":
+        API_PORT = str(fitports()['http'])
+    elif fitargs()['https'] is True:
         API_PROTOCOL = "https"
-        if API_PORT == "None":
-            API_PORT = fitports()['https']
+        API_PORT = str(fitports()['https'])
+    else:  # default protocol is http
+        API_PROTOCOL = "http"
+        API_PORT = str(fitports()['http'])
 
-    if fitargs()['rackhd_host'] == "localhost":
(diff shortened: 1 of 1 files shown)


test/common/fit_common.py
```diff
@@ -212,24 +212,18 @@ def add_globals():
...
-        API_PORT = fitargs()['port']
-
-    if fitargs()['http'] == "True":
+    if fitargs()['http'] is True:
         API_PROTOCOL = "http"
-        if API_PORT == "None":
-            API_PORT = fitports()['http']
-
-    if fitargs()['https'] == "True":
+        API_PORT = str(fitports()['http'])
```
jimturnquist: So, the logic says that -http will supersede -https if both are specified. Is this intentional?  arg_parse has a mutually exclusive capability that will prevent this case from happening at all.

gpaulos: @jimturnquist  No, they are already mutually exclusive as defined in the argument parser below:
    group2 = arg_parser.add_mutually_exclusive_group(required=False)
    group2.add_argument("-http", default="False", action="store_true",
                        help="forces the tests to utilize the http API protocol")
    group2.add_argument("-https", default="False", action="store_true",

Only one protocol can be specified in the command line.

jimturnquist: Ah, I didn't notice that mutual exclusion is already in play.  +1

PR #658, opened 2017-03-07
Rac 4443 Revise API 1.1 to 2.0 in Redfish tests

gpaulos:
This PR revises two Redfish test scripts that still use API 1.1 to use API 2.0. These are the only non-API-1.1 scripts that needs revision for the imminent deprecation of API 1.1.

Both scripts have had references to API 1.1 revised to 2.0 and have had Hound violations fixed.

test_redfish10_system_reset_base.py used API 1.1 via a routine in common/test_api_utils.py which is entirely API 1.1. Erika and I decided that since this was the only reference to that library in the test harness, the easiest and most direct solution was to pull that one routine into the test file and convert to 2.0. 

The future of common/test_api_utils.py is still undetermined but it may be deprecated. If not, it needs a complete rewrite.

@hohene @jimturnquist @johren @larry-dean @stuart-stanley

### DIFF
--- a/test/tests/redfish10/test_redfish10_api_taskservice.py
+++ b/test/tests/redfish10/test_redfish10_api_taskservice.py
@@ -6,14 +6,12 @@ Copyright 2016, EMC, Inc.
 '''
 
 import fit_path  # NOQA: unused import
-import os
-import sys
-import subprocess
 import json
 import fit_common
+from nose.plugins.attrib import attr
 
 
-#Utiities for this script
+# Utiities for this script
 def print_taskid_data(taskid, taskid_json):
     """
     This utility displays the taskjson data for the user
@@ -59,7 +57,7 @@ def get_node_tasklist(nodeid):
         List of task ids on success
         Otherwise empty on failure or error
     """
-    on_url = "/redfish/v1/TaskService/Oem/Tasks/"+nodeid
+    on_url = "/redfish/v1/TaskService/Oem/Tasks/" + nodeid
     on_data = fit_common.rackhdapi(url_cmd=on_url)
 
     tasklist = []
@@ -69,7 +67,7 @@ def get_node_tasklist(nodeid):
             tasklist.append(member['Id'])
     else:
         if fit_common.VERBOSITY >= 2:
-            print "Error in API command. TaskService/Oem/Tasks/"+nodeid+" returned error."
+            print "Error in API command. TaskService/Oem/Tasks/" + nodeid + " returned error."
 
     return tasklist
 
@@ -85,7 +83,7 @@ def get_taskid_data(taskid):
     """
 
     taskid_json = {}
-    on_url = "/redfish/v1/TaskService/Tasks/"+taskid
+    on_url = "/redfish/v1/TaskService/Tasks/" + taskid
     on_data = fit_common.rackhdapi(url_cmd=on_url)
     if on_data['status'] == 200:
         try:
@@ -94,15 +92,13 @@ def get_taskid_data(taskid):
             print "No TaskID data returned"
     else:
         if fit_common.VERBOSITY >= 2:
-            print "Error in API command. TaskService/Oem/Tasks/"+taskid+" returned error."
+            print "Error in API command. TaskService/Oem/Tasks/" + taskid + " returned error."
 
     return taskid_json
 
-# Test Cases
-from nose.plugins.attrib import attr
+
 @attr(all=True, regression=True, smoke=True)
 class redfish10_api_task_suite(fit_common.unittest.TestCase):
-
     def test_redfish_v1_taskservice_tasklist(self):
         # The API /redfish/v1/TaskService will display the list of tasks in the system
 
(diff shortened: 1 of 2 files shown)


test/tests/redfish10/test_redfish10_api_taskservice.py
```diff
@@ -136,8 +132,7 @@ def test_redfish_v1_taskservice_tasklist(self):
         for member in members:
             tasklist.append(member['Id'])
         if fit_common.VERBOSITY >= 2:
-           print ("Task Service contains {0} tasks.".format(len(tasklist)))
-
+            print ("Task Service contains {0} tasks.".format(len(tasklist)))
```
stuart-stanley: If you are using the python3 print() style, you will need to add the following import to the top of the file:
from __future__ import print_function

Otherwise, this will pass flake8, but print out 
```('Task Service contains 2 tasks.', )``` 
Alternative would be to convert to logs.stuff.

Comment (gpaulos) [inline test/tests/redfish10/test_redfish10_api_taskservice.py]: I want to convert this to flogger, but not in this PR. I think that is too much to review and this is time-critical. There is a lot if this stuff all through the legacy code base. Let's make this its own story and chalk it up to technical debt.

File: test/tests/redfish10/test_redfish10_system_reset_base.py [hunk 2]
```diff
@@ -152,8 +145,33 @@ def workflow_tasklist_status_poller(tasklist, tasktype, timeout=180):
...
+            skudata = fit_common.rackhdapi("/api/2.0/skus/" + sku)
+            if skudata['status'] == 200:
+                nodetype = skudata['json'].get("name")
+            else:
+                if fit_common.VERBOSITY >= 2:
+                    errmsg = "Error: SKU API failed {}, return code {} ".format(sku, skudata['status'])
+                    print errmsg
+        else:
+            if fit_common.VERBOSITY >= 2:
+                errmsg = "Error: nodeid {} did not return a valid sku in get_rackhd_nodetype{}".format(nodeid, sku)
```
jimturnquist: I think that get_rackhd_nodetype will be smashed together with the sku.  Need a space?

test/tests/redfish10/test_redfish10_system_reset_base.py
```diff
@@ -152,8 +145,33 @@ def workflow_tasklist_status_poller(tasklist, tasktype, timeout=180):
...
+                print errmsg
+    return nodetype
```
johren: Should we return something like "NOT_FOUND" or "UNKNOWN" or "??????" if we don't find the node ID or the SKU?   It seems like it would be hard to detect when something went wrong if we are just returning "".  In most cases, it will just print out "Node:  xxxxxxx" instead of "Node:  xxxxxxx Quanta-T41".    I don't know that I would detect that there is a problem there.

test/tests/redfish10/test_redfish10_system_reset_base.py
```diff
@@ -152,8 +145,33 @@ def workflow_tasklist_status_poller(tasklist, tasktype, timeout=180):
 
     return task_errorlist
 
+
+def get_rackhd_nodetype(nodeid):
```
anhou: it seems that many files have the same function, and the details are similar but not the same completely
```
$ grep -rn get_rackhd_nodetype * | grep def
test/tests/switch/test_rackhd20_switch_telemetry.py:43:def get_rackhd_nodetype(nodeid):
test/tests/switch/test_rackhd20_switch_pollers.py:42:def get_rackhd_nodetype(nodeid):
test/tests/switch/test_rackhd11_switch_pollers.py:42:def get_rackhd_nodetype(nodeid):
test/tests/switch/test_rackhd11_switch_telemetry.py:43:def get_rackhd_nodetype(nodeid):
test/tests/compute/test_rackhd20_computenode_pollers.py:26:def get_rackhd_nodetype(nodeid):
test/tests/compute/test_rackhd11_computenode_pollers.py:26:def get_rackhd_nodetype(nodeid):
test/common/test_api_utils.py:49:def get_rackhd_nodetype(nodeid):
```

hohene: @anhou This PR is just for pulling the 1.1 APIs out of these two specific scripts that are redfish tests and included in FIT smoke. We are adding more PRs to address non-smoke FIT scripts and another story exists to convert the test_api_utils.py script - RAC-4452.  :-)  
Also, if you dig into the script, the three rackhd20 scripts listed above use api/2.0/nodes within get_rackhd_nodetype.  This function is in test_api_utils, and once updated to 2.0, we should fix the other scripts to use it I think.

gpaulos: There is a limited scope to this PR. We can't fix everything here. This PR needs to be merged before API 1.1 is deprecated, which is imminent. Lets note all of the shortcomings in these files and build stories to address them.

test/tests/redfish10/test_redfish10_system_reset_base.py
gpaulos: Now returns "unknown" when the sku or node is unidentified. Note that this routine is not used in the test logic, it is only for reporting node type during test.

PR #665, opened 2017-03-10
RAC-4396 - Remove rackhd 1.1 tests from smoke.

hohene:
This PR is to turn off the rackhd api 1.1 tests that are executed during FIT smoke.
For now, removed the smoke and regression flags, and created an api_1_1 flag.
Once 1.1 is fully removed from RackHD, we can delete the tests scripts.

Tested on vagrant with one virtual node and hardware rack with multiple nodes.
Run python run_tests.py -stack xx -config xx -test tests -group smoke


hohene:
Test Run:  http://rackhdci.lss.emc.com/job/RackHD/1498/
@johren @gpaulos @keedya @jimturnquist

test/tests/compute/test_rackhd20_computenode_pollers.py
```diff
@@ -23,26 +23,6 @@
...
-            skudata = fit_common.rackhdapi("/api/2.0/skus/" + sku)
-            if skudata['status'] != 200:
-                print "Error: Incorrect HTTP return code on sku, expected 200, received: {}".format(skudata['status'])
-            else:
-                nodetype = skudata['json'].get("name")
-        else:
-            print "Error: nodeid {} did not return a valid sku in get_rackhd_nodetype{}".format(nodeid, sku)
-    return nodetype
-
 from nose.plugins.attrib import attr
```
johren: Nit for later:  Could this be moved up with the other imports?

hohene: sounds good. need to also clean up the calls in this script that are outside of the class, so will do both.

PR #666, opened 2017-03-10
Rac 4453 migrate utils to api2.0

gpaulos:
This PR updates the test/utils files to be API 2.0 compliant. These are not tests and are not used in the PR gate. They are just convenience scripts for utility functions.

- migrated all API 1.1 to 2.0
- fixed Hound violations
- updated copyright notice
- added VERBOSITY = 1 to clean up report output

@hohene @jimturnquist @stuart-stanley @larry-dean @johren

### DIFF
--- a/test/util/display_SYSINFO_settings.py
+++ b/test/util/display_SYSINFO_settings.py
@@ -17,15 +17,15 @@ import test_api_utils
 class display_sysinfo(fit_common.unittest.TestCase):
     def test_01_get_product_info(self):
         print "============== Displaying Product Info"
-        nodes = test_api_utils.monorail_get_node_list(fit_common.fitargs()['ora'])
+        nodes = fit_common.node_select()
         if len(nodes) == 0:
-            print "No Nodes found on Onrack server "+ fit_common.fitargs()['ora']
+            print "No Nodes found on RackHD server "
         else:
             inode=0
             while inode<len(nodes):
                 nn=nodes[inode]
                 print "Node: "+nn
-                monurl = "/api/1.1/nodes/"+nn+"/catalogs/dmi"
+                monurl = "/api/2.0/nodes/"+nn+"/catalogs/dmi"
                 mondata = fit_common.rackhdapi(monurl)
                 catalog = mondata['json']
                 result = mondata['status']
@@ -42,9 +42,9 @@ class display_sysinfo(fit_common.unittest.TestCase):
 
     def test_02_get_catalog_source(self):
         print "============== Displaying Catalog Sources"
-        nodes = test_api_utils.monorail_get_node_list(fit_common.fitargs()['ora'])
+        nodes = fit_common.node_select()
         if len(nodes) == 0:
-            print "No Nodes found on Onrack server "+ fit_common.fitargs()['ora']
+            print "No Nodes found on RackHD server "
         else:
             inode=0
             while inode<len(nodes):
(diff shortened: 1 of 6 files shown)


test/util/display_SYSINFO_settings.py
```diff
@@ -1,31 +1,28 @@
...
 
 import fit_path  # NOQA: unused import
-import os
-import sys
-import subprocess
-
 import fit_common
-import test_api_utils
+
+fit_common.VERBOSIY = 1
```
stuart-stanley: just talked with George and these are intentional and are needed to _keep_ all the debug stuff out of here. If there are other changes to do, I would say add a comment along the lines of:
"# This a CLI utility and not a test. It is only run from a terminal, so make verbosity _not_ show debug information"

If there isn't any other changes, never mind :)

gpaulos: These scripts are just reports and the default verbosity makes the reports unreadable. They are not imported or used in any testing.

johren: Yes, I would agree with Stuart to add a comment to the code so we don't keep rehashing it.

gpaulos:
@johren @stuart-stanley  Added comment line about verbosity=1 to each file

test/util/display_SYSINFO_settings.py
```diff
@@ -1,31 +1,28 @@
-# Copyright 2016, EMC, Inc.
+'''
+Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
```
johren: Not to hold up the approval of the PR, but just a note for PRs going forward.  When modifying the copyright in an existing file, the previous copyright should be preserved.  IOW  2016-1017.

johren:
Ready to merge after @hohene changes request-for-change to approval.

test/util/display_SYSINFO_settings.py
anhou: +1,
@gpaulos  by the way, this file name has a uppercase 'SYSINFO' which is not consistent with other file names unless it has to, at present, on README, Dockerfile has uppercase in file name.

gpaulos: @anhou I don't like the file neames either, but that's the way it has been for a long time.

gpaulos: @anhou I don't like the file names either, but that's the way it has been for a long time.

hohene: feel free to do a mv of the files.  mv dispplay_SYSINFO_settings.py display_sysinfo_settings.py and same for display_node_FIRMWARE_versions.py.
That keeps the log history and easy fix.

johren: Get me know if you want to do the "git mv".  I agree that should be an easy fix.  I'll hold off merge until I hear.

johren:
test this please

PengTian0:
test this please

PengTian0:
test this please

test/util/display_SYSINFO_settings.py
hohene: @gpaulos Did you do a git mv or a copy?  it looks like the files got added as new instead of just moved?

gpaulos:
For some unknown reason, 'git mv' duplicated files rather than renamed them. We suspect its because it is a change of case rather than a name change. In any case, the problem should be resolved now even thought the log looks a bit confused.

jimturnquist:
Just looked at the commit with George.  One of the files was moved and changed in the same PR.  Thats why there is a diff line that looks like the following:

test/util/display_node_FIRMWARE_versions.py → test/util/display_node_firmware_versions.py

Followed by difference lines...

johren:
Any ideas on the hound violation?

gpaulos:
The Hound violation happened after a file rename. I can't reproduce the error using mkcheck.sh. I retested all of the files and they work properly. Maybe a Python 3 artifact? My opinion is that this should just be merged. These are not test files and are not part of any test harness, they are just convenience utilities.

johren:
Based on similar hound errors that I just saw in another PR, my guess is that this is related to changes to "print" in python 3.  I can merge after the Jenkins build is green.

johren:
test this please

PR #671, opened 2017-03-15
Rac-4236 Rac-4325 fix technical debt in RackHD installers

gpaulos:
This PR addresses technical debt in deploy/rackhd_package_install.py and deploy/rackhd_source_install.py .

This is a resubmit of RAC-4236 and RAC-4325 to address concerns about invalid configuration entries

- removed all hard-coded IP addresses, harvests all IP addresses and netmasks from rackhd_config
- convert function call scp_file_to_ora > scp_file_to_host
- use FIT logging instead of print statements
- refactored and moved all shared network routines to new common library common/network_lib.py
- added validation to detect invalid or missing network configuration
- fixed Hound violations

@hohene @larry-dean @jimturnquist @stuart-stanley @johren

### DIFF
--- a/test/common/network_lib.py
+++ b/test/common/network_lib.py
@@ -15,22 +15,44 @@ log = flogging.get_loggers()
 
 
 def get_host_nics(host=fit_common.fitargs()['rackhd_host']):
-    # this routine returns an array of valid network ports on the specified host, default rackhd_host
+    """
+    This routine returns an array of valid network ports on the specified host, default rackhd_host
+    """
     # collect nic names
-    getifs = fit_common.remote_shell("ifconfig -s -a |tail -n +2 |grep -v -e Iface -e lo -e docker", host)
+    getifs = fit_common.remote_shell("ip -o addr |grep -v -e lo -e docker", host)
     # split into array
     splitifs = getifs['stdout'].split('\n')
     niclist = []  # array of valid network ports
-    # clean out login artifacts
+    # clean out login artifacts and verify ports
     for item in splitifs:
-        if "assword" not in item and item.split(" ")[0]:
-            niclist.append(item.split(" ")[0])
+        if "assword" not in item and "inet" in item and item.split(" ")[1]:
+            if fit_common.remote_shell("ip -o addr show " + item.split(" ")[1])['exitcode'] == 0:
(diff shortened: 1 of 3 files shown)


PengTian0:
test this please

PengTian0:
test this please

test/common/network_lib.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    if not ipaddress or not split_ipv4(ipaddress) or not split_ipv4(netmask):
+        return None  # return None if ipaddress is missing or invalid
+    ipsplit = split_ipv4(ipaddress)
+    ip_prefix = ipsplit[0] + '.' + ipsplit[1] + '.' + ipsplit[2] + '.'
+    masksplit = split_ipv4(netmask)
+    dhcp_high = \
+        str(int(ipsplit[0]) + (255 - int(masksplit[0]))) + '.' + \
+        str(int(ipsplit[1]) + (255 - int(masksplit[1]))) + '.' + \
+        str(int(ipsplit[2]) + (255 - int(masksplit[2]))) + '.' + '254'
```
johren: I still have the same concern here that I did in https://github.com/RackHD/RackHD/pull/646 and https://github.com/RackHD/RackHD/pull/647.  If the netmask provided is, for example, 255.255.0.0, this will result in a very invalid dhcpd.conf.  How will this be handled?

johren: I bring that up not as a general error catching issue, but because this specific piece of new code adds that potential for a silent error.

gpaulos:
@johren That's a good catch. I put in some additional code to correct dhcp config when using netmasks such as 255.255.0.0. However, this will NOT fail silently. I did some tests with invalid dhcp configurations and it asserted in several places. With these fixes, it should always create a valid config file.

In addition to the above fixes I also included some additional dhcp config parameters that are required for Ubuntu OS installation.

Note that these deployment scripts are only used in the MN lab, they are NOT used in the PR gate or regression. I suspect that these scripts will eventually be deprecated when a more universal installation routine is developed.

gpaulos:
test this please

gpaulos:
test this please

hohene:
last failure looked like issue with vmslave25 again, rerunning.
test this please

PR #672, opened 2017-03-16
[skip ci]packer ansible: add rackhd install log and version check

changev:
This PR is for:
* Put the version check ahead of time (right after aptitude install)
* Ansible log is not enough. Should increase a bit

@panpan0000 @PengTian0


panpan0000:
Thank you for doing this , @changev 

👍

PR #680, opened 2017-03-20
Add run os install on node per index

keedya:


keedya:
@RackHD/corecommitters @hohene @johren @tannoa2 @geoff-reid 
- This code is tested in jenkins Sandbox under Regression-vagrant where rackHD is connected to 8 compute nodes of baremetal Quantat41.
- this PR will help running different tests concurrently against one instance of rackhd.

keedya:
test this please

panpan0000:
:+0 for this good idea!

keedya:
image

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
     def __get_data(self):
         return loads(self.__client.last_response.data)
 
+    def __get_compute_nodes(self):
+        Api().nodes_get_all()
```
yyscamper: I am not sure whether the on-http python client support query, because `/api/2.0/nodes?type=compute` will directly return all compute nodes, so you don't have to do filter by yourself.

Anyway, your current code is still OK.

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+        LOG.info('compute nodes count {0}'.format(len(compute_nodes)))
+        return compute_nodes
+
+    def __wait_for_completion(self, node, graph_name, graph_instance):
+        id = node.get('id')
+        start_time = date.datetime.now()
+        current_time = date.datetime.now()
+        while True:
+            if (current_time - start_time).total_seconds() > DEFAULT_TIMEOUT_SEC:
+                raise TypeError('Timed out after {0} seconds'.format(DEFAULT_TIMEOUT_SEC))
```
yyscamper: This is not TypeError (see TypeError definition https://docs.python.org/2/library/exceptions.html#exceptions.TypeError), please choose another proper Error type or even a generic Error will be better than TypeError.

keedya: @yyscamper i changed it

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -195,20 +266,45 @@ def install_suse(self, version, nodes=[], options=None, payloadFile=None):
...
         # load the payload from the specified file
         body = {}
         body = self.__get_os_install_payload(payloadFile)
+        kargs = {}
+        # Ubuntu installer requirs from the dhcp both options:
+        # - routers
+        # - domain-name-servers
+        # using static ip instead
+        try:
+            # if node index is set then we can use hard coded static IP
```
changev: Minor question.
Only installing ubuntu using static ip, is this for increasing test coverage? Thanks.

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+
+    def __wait_for_completion(self, node, graph_name, graph_instance):
+        id = node.get('id')
+        start_time = date.datetime.now()
+        current_time = date.datetime.now()
+        while True:
+            if (current_time - start_time).total_seconds() > DEFAULT_TIMEOUT_SEC:
+                raise TypeError('Timed out after {0} seconds'.format(DEFAULT_TIMEOUT_SEC))
+                break
+            Api().workflows_get()
```
yyscamper: isn't the `Api().workflows_get_by_instance_id` is enough? The manual filter is not necessary

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+            index = int(os.environ['NODE_INDEX'])
+        except:
+            LOG.info('NODE_INDEX env is not set')
+            workflows().post_workflows(graph_name, timeout_sec=DEFAULT_TIMEOUT_SEC, nodes=nodes, data=body)
+            return
+
+        # check if index is in the array range
+        nodes = self.__get_compute_nodes()
+        if index >= len(nodes):
+            raise TypeError('index is outside the array range index: {0} vs nodes len {1}'.format(index, len(nodes)))
```
yyscamper: This is not TypeError

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+        id = node.get('id')
+        # delete active workflow on the selected node
+        try:
+            Api().nodes_workflow_action_by_id(id, {'command': 'cancel'})
+        except ApiException as e:
+            assert_equal(404, e.status, message='status should be 404')
+        Api().nodes_post_workflow_by_id(id, name=graph_name, body=body)
+        log_context = self.__get_data().get('logContext')
+        if log_context is None:
+            raise TypeError('Could not find logContext in {0}'.format(self.__get_data()))
```
yyscamper: This is not TypeError

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+        nodes = self.__get_data()
+        compute_nodes = []
+        for n in nodes:
+            type = n.get('type')
+            if type == 'compute':
+                compute_nodes.append(n)
+        LOG.info('compute nodes count {0}'.format(len(compute_nodes)))
+        return compute_nodes
+
+    def __wait_for_completion(self, node, graph_name, graph_instance):
```
yyscamper: If the `graph_instance` is specified, actually the argument `node` and `graph_name` is redundant. because when the graph instance is created, the node identifier and graph name is known.

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+        start_time = date.datetime.now()
+        current_time = date.datetime.now()
+        while True:
+            if (current_time - start_time).total_seconds() > DEFAULT_TIMEOUT_SEC:
+                raise TypeError('Timed out after {0} seconds'.format(DEFAULT_TIMEOUT_SEC))
+                break
+            Api().workflows_get()
+            workflows = loads(self.__client.last_response.data)
+            for w in workflows:
+                # LOG.info('print w : {0}'.format(w))
```
yyscamper: Please remove these unnecessary code

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -195,20 +266,45 @@ def install_suse(self, version, nodes=[], options=None, payloadFile=None):
...
-    def install_ubuntu(self, version, payloadFile, nodes = []):
+    def install_ubuntu(self, version, payloadFile, nodes=[]):
         graph_name = 'Graph.InstallUbuntu'
         os_repo = defaults.get('RACKHD_UBUNTU_REPO_PATH', \
             self.__base + '/repo/ubuntu')
         # load the payload from the specified file
         body = {}
         body = self.__get_os_install_payload(payloadFile)
+        kargs = {}
+        # Ubuntu installer requirs from the dhcp both options:
```
yyscamper: typo "requirs"

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+            time.sleep(10)
+            current_time = date.datetime.now()
+            # LOG.info('current time {0} vs start time {1}'.format(current_time, start_time))
+
     def __post_workflow(self, graph_name, nodes, body):
-        workflows().post_workflows(graph_name, timeout_sec=DEFAULT_TIMEOUT_SEC, nodes=nodes, data=body)
+        # check if NODE_INDEX is set
+        index = None
+        try:
+            index = int(os.environ['NODE_INDEX'])
```
yyscamper: I remembered we had a cleanup of using environment variable (Joel still worked in RackHD at that time) because we discussed and agreed the environment variable was a kind of global variable even across processes, it is error-prone and hard to maintain. And eventually the config.init is created (https://github.com/RackHD/RackHD/blob/master/test/config/config.ini).

However, your code is trying to introduce the environment variable again. Base on our previous conclusion, please move the `NODE_INDEX` into config file, if you want to dynamically change the `NODE_INDEX` via command line argument, you need some code to merge the command line argument and config.ini.

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+        except:
+            LOG.info('NODE_INDEX env is not set')
+            workflows().post_workflows(graph_name, timeout_sec=DEFAULT_TIMEOUT_SEC, nodes=nodes, data=body)
+            return
+
+        # check if index is in the array range
+        nodes = self.__get_compute_nodes()
```
yyscamper: Actually, there is no guarantee for RackHD API that multiple call for the same API will always return data in the same order. 
Without the guarantee, different index may return the same node document and fail the workflow testing.
To ensure the node document order, you could manually sort the `nodes`.

@brianparry please correct me if my above understanding is wrong.

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+            Api().nodes_workflow_action_by_id(id, {'command': 'cancel'})
+        except ApiException as e:
+            assert_equal(404, e.status, message='status should be 404')
+        Api().nodes_post_workflow_by_id(id, name=graph_name, body=body)
+        log_context = self.__get_data().get('logContext')
+        if log_context is None:
+            raise Exception('Could not find logContext in {0}'.format(self.__get_data()))
+            return
+        # load graph instance id
+        graph_instance = log_context.get('graphInstance')
```
yyscamper: Why not directly use the `instanceId` but rather than turn to use `logContext.graphInstance`? the `instanceId` is defined in graph object model, but logContext is a dynamic property.

yyscamper:
@keedya : Could you please detail how to use this feature? Run multiple CIT instances in RackHD host environment's multiple shell? If so, I believe there should be a Jenkins pipeline change, but I don't find it.

yyscamper (changes requested):
Please submit the missing pipeline change PR, and clean up the environment variable usage.

test/tests/api/v2_0/os_install_tests.py
keedya: Global variables aren't global accross processes

test/tests/api/v2_0/os_install_tests.py
keedya: There are not actualy redundant. This function tracks workflows based on node id, graph name and graph instance.

test/tests/api/v2_0/os_install_tests.py
keedya: Ubuntu requires router option and dns server in the dhcp seerver . Adding these options breaks the installation of ESXI5.5.
the solution here is to remove the options from dhcp server and use static IP for Ubuntu

keedya:
@yyscamper the pipeline is not part of my task. The plan is to run multiple instance of the test with different NODE_INDEX. There is a job as I mentioned earlier in the sandbox(I posted a snapshot in this PR )
I know @panpan0000 team is been working on the pipeline scripts and they are more expert in this field and they probably can help me convert the job in jenkins to pipeline.
I will deploy the job from jenkins Sandbox to production once this PR is merged

Finally there reason behind this PR is to run all flavors of supported OSs on baremetal in parallel and reduce the job time to 1 hour. This can be so beneficial to our release pipeline

geoff-reid (approved):

test/tests/api/v2_0/os_install_tests.py
keedya: @yyscamper i haven't encountered this issue before. @brianparry if true, i can sort the list.

test/tests/api/v2_0/os_install_tests.py
```diff
@@ -46,8 +50,75 @@ def teardown(self):
...
+            workflows = loads(self.__client.last_response.data)
+             for w in workflows:
+                # LOG.info('print w : {0}'.format(w))
+                if (w.get('node') == id and w.get('injectableName') == graph_name
+                   and w.get('instanceId') == graph_instance):
+                   status = w.get('status')
+                   # LOG.info('{0} - target: {1}, status: {2}'.format(w.get('injectableName'), id, status))
+                   if status == 'succeeded' or status == 'failed' or status == 'canceled':
+                       LOG.info('{0} - target: {1}, status: {2}'.format(w.get('injectableName'), id, status))
+                       if status == 'failed' or status == 'canceled':
```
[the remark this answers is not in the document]
WeiyuanSong: Checked against the repo's own rule: docs/wiki/guardrails/security/grpc-error-sanitization.md governs only the boundary between internal gRPC and *external* callers (BFF resolvers, GraphQL, public REST) and explicitly states internal gRPC-to-gRPC calls within the same trust domain are out of scope. GetExecutePolicyCallData is called by internal product services, so this path does not cross that boundary. The message also intentionally mirrors the identical, already-merged pattern in maybeBuildAddOwnerCall (install_calldata_common.go — same wording, same %v) so the two drift-heal paths stay identical, and the wallet address is server-derived from FAS, not reflected user input. No change.

🤖 Reply generated with Toshi

PR #681, opened 2017-03-20
Rac 4234 obsolete fit common routines

gpaulos:
This PR fixes some technical debt in fit_common.py:

- removed obsolete function appliance_reset() which used BMC to reset the appliance in the old MN lab. No longer relevant, because it doesn't work with Vagrant or with ESXi stacks. Can just use remote_shell('reboot') for a reset.

- removed obsolete function apply_obm_settings_seq(), apply_obm_settings() is the current supported version

- fixed mongo_reset() which was set up for OnRack configuration and broken for current, Now works with RackHD source and package installations.

@hohene @larry-dean @jimturnquist @stuart-stanley @johren

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -203,6 +203,7 @@ def apply_stack_config():
         if 'ucs_port' in fitcfg():
             fitargs()['ucs_port'] = fitcfg()['ucs_port']
 
+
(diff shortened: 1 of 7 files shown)


gpaulos:
Sorry, wrong branch

PR #690, opened 2017-03-23
RAC-4629

jimturnquist:
Adds the --extra option to run_tests.py.  Allows for addition
configuration files to be added to the generated configuration
file.  All added files should be relative to config-mn.

  -extra EXTRA   comma separated list of extra config files (found in
                 'config' directory)

@hohene @gpaulos @johren @nortonluo

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -169,6 +169,10 @@ def compose_config(use_sysargs=False):
             # stack overlay configuration
             apply_stack_config()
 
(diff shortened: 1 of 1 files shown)


johren:
test this please

johren:
I started a test to try out slack notifications and then aborted it.  The write-back stage doesn't run when the test is aborted so it looks like it is still running.  However, https://rackhdci.lss.emc.com/job/RackHD/112/ did succeed right before that.

PR #699, opened 2017-04-05
RAC-4744 Parallel API 2.0 OS install test for local or NFS repos

gpaulos:
This script implements the 'parallel' OS install algorithm for locally-installed or NFS-mounted repos using the API 2.0 interface. This test pulls repo information from config/install_default.json to run minimum-payload installation tests. This test script is suitable for use in PR gate because it uses minimum resources and executes in under 20 minutes. 

To run all tests, a stack must contain 12 nodes, virtual or physical but can be run with fewer. As with the proxy-based tests, this test will skip any test where there are insufficient nodes to run or where there is no repo defined. The order of repos defined in config/install_default.json will determine test priority. 

This test is NOT tagged to run under Smoke or Regression runs and must be executed at the run_test.py command line for use in its own pipeline.

CoreOS test may fail due to this product bug:
https://rackhd.atlassian.net/browse/RAC-4755

ESXi tests fail in MN lab due to this product bug:
https://rackhd.atlassian.net/browse/RAC-4125

@hohene @jimturnquist @johren @keedya


test/config/install_default.json
```diff
@@ -30,6 +30,27 @@
       "proxy": {
         "host": "",
         "port": ""
-      }
+      },
+      "os-install": [
+        {"workflow": "Graph.InstallESXi", "version": "6.0", "path": "/repo/esxi/6.0", "kvm": false},
+        {"workflow": "Graph.InstallCentOS", "version": "6.5", "path": "/repo/centos/6.5", "kvm": false},
+        {"workflow": "Graph.InstallUbuntu", "version": "trusty", "path": "/repo/ubuntu", "kvm": false},
+        {"workflow": "Graph.InstallESXi", "version": "5.5", "path": "/repo/esxi/5.5", "kvm": false},
```
yyscamper: Is there any reason to not put the entries of the same OS together? Put same OS entries together will make it more clear to know the parameter difference for the same OS.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
```diff
@@ -0,0 +1,292 @@
...
+            return NODE_STATUS[entry]['id']
+    return ""
+
+
+# Match up tests to node IDs to feed skip decorators
+index = 0  # node index
+for item in OSLIST:
+    if index < len(NODECATALOG):
+        NODE_STATUS[NODECATALOG[index]] = \
+            {"workflow": item['workflow'], "version": item['version'], "kvm": item['kvm'], "id": "Pending"}
```
yyscamper: initialize the `id` as `pending` is confusing, sine `pending` is a kind of workflow status.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
```diff
@@ -0,0 +1,292 @@
...
+
+# collect repo information from config files
+OSLIST = fit_common.fitcfg()["install-config"]["os-install"]
+
+# download RackHD config from host
+rackhdconfig = fit_common.rackhdapi('/api/2.0/config')['json']
+rackhdhost = "http://" + str(rackhdconfig['apiServerAddress']) + ":" + str(rackhdconfig['apiServerPort'])
+
+
+# this routine polls a task ID for completion
```
yyscamper: In rackhd namespace, the `task` has special meaning. But at here, the `task` is actually 'workflow', so to avoid confusion, I'd suggest changing the word from `task` to `workflow`. The same for whole rest of code.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
```diff
@@ -0,0 +1,292 @@
...
+                         'id': "failed"}
+                    log.error(" TaskID: " + result['text'])
+                    log.error(" Payload: " + fit_common.json.dumps(payload_data))
+                # increment node index to run next bootstrap
+                nodeindex += 1
+
+    @fit_common.unittest.skipUnless(node_taskid("Graph.InstallESXi", "5.", False) != '',
+                                    "Skipping ESXi5.5, repo not configured or node unavailable")
+    def test_api20_bootstrap_esxi5(self):
+        self.assertTrue(wait_for_task_complete(node_taskid("Graph.InstallESXi", "5.", False)), "ESXi5.5 failed.")
```
yyscamper: Why the version is "5." rather than "5.5"?

gpaulos: This allows 5.x for possibilities such as 5.51 etc.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
gpaulos: OK good point.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
gpaulos: There is no workflow submitted yet, but this key needs to be initialized so I used 'pending'. Any other ideas?

test/config/install_default.json
gpaulos: The order was selected to run the three PR gate OS installs first, for example in the case of only three nodes available.

Comments are not supported in JSON files. I can add that to the comment section in the test file.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
```diff
@@ -0,0 +1,297 @@
...
+
+# ------------------------ Tests -------------------------------------
+
+
+@attr(all=True)
+class api20_bootstrap_base(fit_common.unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # run all OS install workflows first
+        nodeindex = 0
```
hohene: Add a check in here to make sure NODECATALOG is not [] and if so, log an error for the user and be done.
if NODECATALOG:  for item in OSLIST: ...  else log.error("no nodes exists in testbed") ....along that line.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
```diff
@@ -0,0 +1,297 @@
...
+NODE_STATUS = {}
+
+# global timer
+START_TIME = time.time()
+
+# collect repo information from config files
+OSLIST = fit_common.fitcfg()["install-config"]["os-install"]
+
+# download RackHD config from host
+rackhdconfig = fit_common.rackhdapi('/api/2.0/config')['json']
```
hohene: do a quick check that this api call worked, else log an error and can't do much more.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
```diff
@@ -0,0 +1,297 @@
...
+                         "id": result['json']['instanceId']}
+                    log.info_5(" TaskID: " + result['json']['instanceId'])
+                    log.info_5(" Payload: " + fit_common.json.dumps(payload_data))
+                else:
+                    # if no task ID is returned put 'failed' in ID field
+                    NODE_STATUS[NODECATALOG[nodeindex]] = \
+                        {"workflow": item['workflow'],
+                         "version": item['version'],
+                         "kvm": item['kvm'],
+                         'id': "failed"}
```
hohene: Can you log an error message here - log.error(" os install {} on node {} failed" + NODECATALOG[nodeindex] + workflow) here?  Makes it easier to see/debug a failed workflow.  thanks.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_local.py
gpaulos: If NODECATALOG is [] then it will skip all tests.

johren:
Could you include a sample test run since this is not being run currently in the PR gate?

gpaulos:
@johren 
This is the console output from a run of three OS installs: CentOS 7, Suse, Ububtu14
[os_install.txt](https://github.com/RackHD/RackHD/files/914500/os_install.txt)

gpaulos:
test this please

johren:
I'm waiting on the "Changes Requested" to be cleared before merging.

johren:
test this please

gpaulos:
please test this

gpaulos:
test this please

panpan0000:
@gpaulos  @johren 
I found the command line from the txt, but still confused for "  which node will run which OS install, do you configure it somewhere?
"
```
python run_tests.py -config config-mn -stack 4 -test tests/bootstrap/test_api20_os_bootstrap_parallel_local.py -sku Unidentified-Compute
```

PR #700, opened 2017-04-05
Rac 4327 Parallel API 2.0 OS install test for proxy repos

gpaulos:
This script implements 'parallel' OS install via proxy using API 2.0 commands. As with the Redfish proxy test, the script harvests test repo information from the config.json 'http-proxies' section then runs a minimum-payload installation test on each.  This test script is suitable for use in PR gate because it uses minimum resources and executes in under 20 minutes.

To run all tests, a stack must contain 11 nodes, virtual or physical but can be run with fewer. As with the proxy-based tests, this test will skip any test where there are insufficient nodes to run or where appropriate repo info is not present in config.json.

This test does NOT contain a Windows OS install which cannot be performed via proxy.

This test is NOT tagged to run under Smoke or Regression runs and must be executed at the run_test.py command line for use in its own pipeline.

CoreOS test may fail due to this product bug:
https://rackhd.atlassian.net/browse/RAC-4755

ESXi tests fail in MN lab due to this product bug:
https://rackhd.atlassian.net/browse/RAC-4125

@hohene @jimturnquist @johren @keedya

### DIFF
--- a/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
+++ b/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
@@ -150,7 +150,7 @@ for item in OSLIST:
 # ------------------------ Tests -------------------------------------
 
 
-@attr(all=True)
+@attr(all=False)
 class api20_bootstrap_base(fit_common.unittest.TestCase):
     @classmethod
     def setUpClass(cls):
--- a/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
+++ b/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
@@ -65,6 +65,7 @@ import fit_path  # NOQA: unused import
 from nose.plugins.attrib import attr
 import fit_common
 import flogging
+import sys
 import time
 log = flogging.get_loggers()
 
--- a/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
+++ b/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
@@ -84,7 +84,8 @@ rackhdhost = "http://" + str(rackhdconfig['apiServerAddress']) + ":" + str(rackh
 
 # this routine polls a workflow task ID for completion
 def wait_for_workflow_complete(taskid):
-    while time.time() - START_TIME < 1800:  # limit test to 30 minutes
+    result = None
+    while time.time() - START_TIME < 1800 or result is None:  # limit test to 30 minutes
         result = fit_common.rackhdapi("/api/2.0/workflows/" + taskid)
         if result['status'] != 200:
             log.error(" HTTP error: " + result['text'])
--- a/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
+++ b/test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
@@ -4,7 +4,7 @@ Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
 Author(s):
 George Paulos
 
-This script tests base case of the RackHD API 2.0 OS bootstrap workflows.
+This script tests minimum payload base case of the RackHD API 2.0 OS bootstrap workflows using proxy method.
 This routine runs OS bootstrap jobs simultaneously on multiple nodes.
 For 11 tests to run, 11 nodes are required in the stack. If there are less than that, tests will be skipped.
 This test takes 15-20 minutes to run.
@@ -65,6 +65,7 @@ import fit_path  # NOQA: unused import
 from nose.plugins.attrib import attr
 import fit_common
 import flogging
+import time
 log = flogging.get_loggers()
 
 # This gets the list of nodes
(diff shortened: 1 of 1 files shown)


test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
```diff
@@ -0,0 +1,255 @@
...
+          "server": "http://os-mirror-server/mirrors/sles/12"
+        },
+        {
+          "localPath": "/Ubuntu/14",
+          "remotePath": "/",
+          "server": "http://os-mirror-server/mirrors/ubuntu/boot/14.04.4"
+        }
+      ],
+
+For each OS type, the "localpath" string must conform to the convention above.
```
yyscamper: localpath => localPath

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
```diff
@@ -0,0 +1,255 @@
...
+                         "id": result['json']['instanceId']}
+                    log.info_5(" TaskID: " + result['text'])
+                    log.info_5(" Payload: " + fit_common.json.dumps(payload_data))
+                else:
+                    # this is the failure case where there is no task ID
+                    NODE_STATUS[NODECATALOG[nodeindex]] = \
+                        {"workflow": item['workflow'],
+                         "version": item['version'],
+                         "kvm": item['kvm'],
+                         'id': "failed"}
```
yyscamper: set `id` as failed? isn't `status`?

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
```diff
@@ -0,0 +1,255 @@
...
+    {"workflow": "Graph.InstallCentOS", "version": "6.5", "path": "/CentOS/6.5", "kvm": True},
+    {"workflow": "Graph.InstallCentOS", "version": "7", "path": "/CentOS/7.0", "kvm": True}
+]
+
+# Match up tests to node IDs to feed skip decorators
+index = 0  # node index
+for item in OSLIST:
+    if proxy_select(item['path']) and index < len(NODECATALOG):
+        NODE_STATUS[NODECATALOG[index]] = \
+            {"workflow": item['workflow'], "version": item['version'], "kvm": item['kvm'], "id": "Pending"}
```
yyscamper: set "id" as "Pending"? I think this is a typo, `id` need be changed to `status` or something else.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
gpaulos: At this point there is no ID because the workflow failed to initialize. I could have put in another key pair but this was simpler, so this is intentional.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
gpaulos: At this point, there is no workflow yet submitted. This sets up the dict for later processing. This routine is to feed the skip decorators with the valid tests. Once again, I could have put in another key pair for this condition but this is simpler, so this is intentional.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
```diff
@@ -0,0 +1,260 @@
...
+
+# download RackHD config from host
+rackhdconfig = fit_common.rackhdapi('/api/2.0/config')['json']
+rackhdhost = "http://" + str(rackhdconfig['apiServerAddress']) + ":" + str(rackhdconfig['apiServerPort'])
+
+
+# this routine polls a workflow task ID for completion
+def wait_for_workflow_complete(taskid):
+    result = None
+    while time.time() - START_TIME < 1800 or result is None:  # limit test to 30 minutes
```
hohene: nice!

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
```diff
@@ -0,0 +1,260 @@
...
+
+# ------------------------ Tests -------------------------------------
+
+
+@attr(all=True)
+class api20_bootstrap_base(fit_common.unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # run all OS install workflows first
+        nodeindex = 0
```
hohene: Add a check here to see if NODECATALOG is empty,  if so, log an obvious error for the user.
```
 ie.  if NODECATALOG:
          for item in OSLIST:
              ....
      else:
          log.error("no nodes exist, skipping it all")   
```

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
```diff
@@ -0,0 +1,260 @@
...
+NODECATALOG = fit_common.node_select()
+
+# dict containing bootstrap workflow IDs and states
+NODE_STATUS = {}
+
+# global timer
+START_TIME = time.time()
+
+# download RackHD config from host
+rackhdconfig = fit_common.rackhdapi('/api/2.0/config')['json']
```
hohene: Should probably check if rackhdconfig came back with valid data, ie. status = 200.  relying on it further below in proxy_select.

gpaulos: I suppose. if not, It will just error out anyways.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
gpaulos: It will skip all tests anyways if there are no nodes. 

if nodeindex < len(NODECATALOG):

johren:
Could you include a sample test run since this is not being run currently in the PR gate?

gpaulos:
Console output from OS install run here in MN. Not all tests can be executed here in a single run due to resource limits and some hardware problems: 
[os_install_proxy.txt](https://github.com/RackHD/RackHD/files/917797/os_install_proxy.txt)

gpaulos:
test this please

johren:
I'm waiting on the "Changes Requested" to be cleared and an approval before merging.

test/tests/bootstrap/test_api20_os_bootstrap_parallel_proxy.py
```diff
@@ -0,0 +1,260 @@
...
+def wait_for_workflow_complete(taskid):
+    result = None
+    while time.time() - START_TIME < 1800 or result is None:  # limit test to 30 minutes
+        result = fit_common.rackhdapi("/api/2.0/workflows/" + taskid)
+        if result['status'] != 200:
+            log.error(" HTTP error: " + result['text'])
+            return False
+        if result['json']['status'] == 'running' or result['json']['status'] == 'pending':
+            log.info_5("{} workflow status: {}".format(result['json']['injectableName'], result['json']['status']))
+            fit_common.time.sleep(30)
```
gpaulos: Instead I switched to fit_common.time.time() and not import time.

johren:
test this please

gpaulos:
test this please

PR #613, opened 2017-02-03
Clean up for flake8 checker.  Only import warnings remain.

jimturnquist:

PR #701, opened 2017-04-06
Rac 1984 cit test fix2

lacarb:
Correction for an earlier CIT test modification.
This should make CIT green for both master and this API fix - https://github.com/RackHD/on-http/pull/613

### DIFF
--- a/test/tests/api/v2_0/nodes_tests.py
+++ b/test/tests/api/v2_0/nodes_tests.py
@@ -269,7 +269,7 @@ class NodesTests(object):
             assert_equal(200,resp.status, message=resp.reason)
         assert_raises(rest.ApiException, Api().nodes_get_catalog_source_by_id, 'fooey','bmc')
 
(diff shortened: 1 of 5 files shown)


test/tests/api/v2_0/nodes_tests.py
```diff
@@ -294,6 +294,7 @@ def test_node_workflows_post(self):
         resps = []
         Api().nodes_get_all()
         nodes = self.__get_data()
+        print nodes
```
VulpesArtificem: did you want to keep this print statement?

lacarb: I'll cut it out

hohene:
We need PR #613 to merge before this, correct?

lacarb:
@VulpesArtificem, print has been removed. I'm not sure what's up with #613

lacarb:
@RackHD/corecommitters

PR #705, opened 2017-04-07
Correction for CIT test modification.

lacarb:
Correction for an earlier CIT test modification.
This should make CIT green for both master and this API fix - https://github.com/RackHD/on-http/pull/613


DavidjohnBlodgett:
👍 after tests!

lacarb:
@RackHD/corecommitters, Super small PR. Please give it a look

PR #712, opened 2017-04-11
RAC-1984 (somewhat) Removal of PUT node/id/workflows CIT test

lacarb:
This relates to #707. I did not consider that the PUT test would have the same issue as POST. They both use and fail when posting a workflow when the GET API fix is present.

This CIT test fails when run against the fix for GET node/:id/workflows (RackHD/on-http#613)

The PUT for node/id/workflows works with master and the above GET fix. However, the CIT test breaks and blocks the GET fix from getting merged. I propose that this test be temporarily removed so the GET API fix can be moved in. Meanwhile a new ticket will be added for correcting the CIT test for PUT.

What I have in the pipes at the moment is this test removal, the API fix, and a new GET CIT test.


DavidjohnBlodgett:
👍

lacarb:
test this please

PR #716, opened 2017-04-12
Updated Vagrantfile and README

pscharla:
@panpan0000 @DavidjohnBlodgett @RackHD/corecommitters 
Updates to the RackHD example directory to get the Vagrantfile and README in sync with the most recent RackHD releases

### DIFF
--- a/example/Vagrantfile
+++ b/example/Vagrantfile
@@ -5,100 +5,134 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-
-    # RackHD SERVER
+    
+    # MONORAIL SERVER
     config.vm.define "dev" do |target|
         target.vm.box = "rackhd/rackhd"
         target.vm.provider "virtualbox" do |v|
             v.memory = 4096
             v.cpus = 4
-            v.customize ["modifyvm", :id, "--nicpromisc2", "allow-all", "--ioapic", "on"]
-        end
-
-        # NOTE: virtualbox shared folder syncing can be quite slow, but shared folders
-        # can be configured as NFS mounts to the host instead.
-        # To do so, append: `, type: "nfs"` to all of the target.vm.synced_folder commands
-        # below, then on the host system, run `sudo nfsd start`.
-        # Additionally, uncomment the below configuration:
-
-        # For NFS mounts, a host only network is needed. Feel free to change the IP.
-        # target.vm.network "private_network", ip: "192.168.50.4"
-
-        # Sync optional local workspace directory to the vagrant instances src/ path.
-        # REPO is optional when WORKSPACE is defined, used for synchronizing single RackHD repos
-        # otherwise it assumes all service repos have been cloned and built under the WORKSPACE path.
-        # NOTE: See further below for instructions on how to use an NFS mount for faster synced file performance
-        if ENV['WORKSPACE']
-          if ENV['REPO'] == "on-tasks"
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/on-taskgraph/node_modules/on-tasks/"
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/on-http/node_modules/on-tasks/"
-          elsif ENV['REPO'] == "on-core"
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/on-taskgraph/node_modules/on-core/"
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/on-http/node_modules/on-core/"
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/on-tftp/node_modules/on-core/"
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/on-dhcp-proxy/node_modules/on-core/"
-          end
-          target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['REPO']}", "/home/vagrant/src/#{ENV['REPO']}"
-          if ENV['CONFIG_DIR']
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['CONFIG_DIR']}", "/opt/onrack/etc/"
-            target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['CONFIG_DIR']}", "/home/vagrant/opt/monorail/"
-          end
+            v.customize ["modifyvm", :id, "--nicpromisc2", "allow-all"]
         end
 
         # Create a public network, which generally matched to bridged network.
         # Bridged networks make the machine appear as another physical device on
         # your network.
         # target.vm.network :public_network
+        if ENV['WORKSPACE']
+          target.vm.synced_folder "#{ENV['WORKSPACE']}/build-deps", "/home/vagrant/src/"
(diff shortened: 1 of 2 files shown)


DavidjohnBlodgett:
@pscharla so I'm noticing a lot of changes happening inside the vagrant file, large amounts of it I would not have expected and wondered where they came from... i.e. there seems to be a lot of code for `pm2` and I thought the new builds no longer have pm2 because we just start the services with `service` in addition I noticed a lot of changes around the `env` variables... was that not working and needed all the new edits to function correctly?

pscharla:
@DavidjohnBlodgett : yes that newer Vagrantfile is the one from on-build-config (with a couple edits) that works with the newer RackHD releases.  My understanding for `pm2` was that it used to be part of the setup instructions but now incorporated into the Vagrantfile. The `env` variable differences appear to be at least part of the reason the new environment behaves quite differently compared to how it used to be in the older releases (i.e. 0.15, 0.16)

My hope in this PR was to at least get the Vagrantfile back to a working state and the docs updated appropriately. I think we can continue to refine the Vagrantfile as we go forward.  As things stand now, it's completely broken.

DavidjohnBlodgett:
@pscharla 

Do note that the build tools your using as a source for vagrant file changes is set to not use current builds (see this link): https://github.com/RackHD/on-build-config/blob/master/vagrant/Vagrantfile#L12 

Hence my concerns that the code is expecting tools to exist that I am not seeing on the VMs produced from this deployment process (i.e. pm2 is not packaged with our example VMs)

pscharla:
@DavidjohnBlodgett : Understood about the other Vagrantfile calling out a specific version of the rackhd. I removed that line so it should grab the latest.

DavidjohnBlodgett:
👍 for doc changes, 👎 for vagrant file changes... I'll add my concerns directly to the code.

example/Vagrantfile
```diff
@@ -5,100 +5,134 @@
...
+         end
+          list = ENV['PYTHON_REPOS'].split(' ')
+          list && list.each do |repo|
+            config.vm.provision "shell" do |s|		
+            s.inline = "cd /home/vagrant/src/#{repo} && pip install -r requirements.txt"		
+            s.privileged = true
+            end
+          end
+        end
+
```
DavidjohnBlodgett: the whole section (65-88) above seems like its for testing and python repos related directly to the deployment process, it is unclear to me why these would be included or if they are even documented.

example/Vagrantfile
```diff
@@ -5,100 +5,134 @@
...
+            s.privileged = true
+            end
+          end
+        end
+        
         # Specify optional interface to bridge when wanting to expose the instance to an
         # external network.  Useful for when discovering nodes externally.
         if ENV['BRIDGE_INTF']
-           target.vm.network "public_network", ip: "172.31.128.1", bridge: "#{ENV['BRIDGE_INTF']}"
+          target.vm.network "public_network", ip: "172.31.128.1", bridge: "em1"
```
DavidjohnBlodgett: this seems to be a local change, and not something that we'd want to move away from using an environment variable for.

example/Vagrantfile
```diff
@@ -5,100 +5,134 @@
...
         else
-           target.vm.network "private_network", ip: "172.31.128.1", virtualbox__intnet: "closednet"
+          target.vm.network "private_network", ip: "172.31.128.1", virtualbox__intnet: "closednet"
```
DavidjohnBlodgett: not sure what the change here was, but ruby can be very funny about tabs and spaces... may want to make sure this doesn't pull that `target.vm.network` def out of scope

example/Vagrantfile
```diff
@@ -5,100 +5,134 @@
...
+          service isc-dhcp-server start
+          service rsyslog stop
+          echo manual | sudo tee /etc/init/rsyslog.override
+          if [ "#{ENV['MULTI']}" ]; then
+            pm2 logs > /home/vagrant/src/build/vagrant.log &
+          else
+            pm2 logs > /home/vagrant/src/"#{ENV['REPO_NAME']}"/vagrant.log &
+          fi
+          waitForPM2Daemon
+          pm2 start rackhd-pm2-config.yml
```
DavidjohnBlodgett: 108-133 this whole sections seems to expect `pm2` to exist... since 2.0.0 I have not seen pm2 packaged with the example VM. I would suggest keeping the start of isc-dhcp, and removing the rest of this code.

example/Vagrantfile
```diff
@@ -5,100 +5,134 @@
...
-            v.customize ["modifyvm", :id, "--nicpromisc2", "allow-all", "--ioapic", "on"]
-        end
-        #target.vm.network :public_network
-        target.vm.network "private_network", virtualbox__intnet: "closednet", auto_config: false, nic_type: "82540EM"
-        target.vm.provision "shell", inline: <<-SHELL
-            sudo ifconfig eth1 up
-            sleep 2
-            sudo udhcpc -i br1
-        SHELL
-    end
```
DavidjohnBlodgett: del of 83-101, this removes the creation of our HW simulator... this should not be removed or there will be no way to demo the functioning of this system.

example/Vagrantfile
```diff
@@ -5,100 +5,134 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-
-    # RackHD SERVER
+    
+    # MONORAIL SERVER
```
anhou: any reasons modify RackHD to MONORAIL, I'm afraid users will be confused when they saw monorail

example/Vagrantfile
changev: agree to @DavidjohnBlodgett 

These steps is to volume host file to vagrant box for testing, in test the base vagrant box is v0.16, which is a box just for test.

With this no specific box version Vagrantfile, the latest released box will be pulled. Nothing need to do for a released box. RackHD services start automatically.

example/Vagrantfile
```diff
@@ -5,100 +5,134 @@
...
+              target.vm.synced_folder "#{ENV['WORKSPACE']}/build", "/home/vagrant/src/on-taskgraph/node_modules/on-core/"  
+              target.vm.synced_folder "#{ENV['WORKSPACE']}/build", "/home/vagrant/src/on-http/node_modules/on-core/" 
+              target.vm.synced_folder "#{ENV['WORKSPACE']}/build", "/home/vagrant/src/on-tftp/node_modules/on-core/" 
+              target.vm.synced_folder "#{ENV['WORKSPACE']}/build", "/home/vagrant/src/on-dhcp-proxy/node_modules/on-core/" 
+            end
+          end
+        end
+        
+        if ENV['CONFIG_DIR']
+          target.vm.synced_folder "#{ENV['WORKSPACE']}/#{ENV['CONFIG_DIR']}", "/opt/monorail/"
```
changev: This line of code can be kept.
A RackHD config.json file for users can be put in the same directory for replacing default  rackhd config.

changev:
Thanks for catching this issue. Our CD has proceeded a lot and this section should be modified.

Some tips for Vagrantfile change:
* on-build-config/Vagrantfile pulls v0.16 rackhd box, which is just for test, so lots of extra work need to do when vagrant up.

*Vagrantfile with no specific box version will pull latest released box, every thing is good in this box.

example/Vagrantfile
pscharla: yes agreed. changing now

example/Vagrantfile
pscharla: agreed. will put this back

example/Vagrantfile
pscharla: agreed. changing back

example/Vagrantfile
pscharla: Thanks for the heads up. I think the original Vagrantfile uses 3 spaces of indentation inside the if block and the one from on-build-config uses 2. Not sure if it matters?

pscharla:
After further discussion with @DavidjohnBlodgett I'm going to close out this PR and open a new one for just the documentation updates. It appears that the Vagrantfile from on-build-config has a bunch of extra logic in it that is no longer necessary (i.e. pm2).

PR #720, opened 2017-04-18
Packer: update hostname and hosts together

panpan0000:
otherwise, in the following script firstuser.sh (run after dep.sh)
it will be very **slow** when doing any shell command with "sudo"
and packer will dump warning message like ```sudo: unable to resolve host perfl133```

@yyscamper  @anhou  @PengTian0  @changev

### DIFF
--- a/packer/scripts/dep.sh
+++ b/packer/scripts/dep.sh
@@ -9,6 +9,9 @@ pip install --upgrade setuptools
 pip install ansible==2.2.0.0
 
 # set a friendly hostname
+# by default, the hostname in both /etc/hosts & /etc/hostname are obtained from DHCP server during install.
 rm -f /etc/hostname
 echo "rackhd" > /etc/hostname
-
+#Below two lines were moved from scripts/cleanup.sh. I can't recall why they didn't go with /etc/hostname change together...
+NEW_HOST_NAME=$(cat /etc/hostname)
+sed -i  "/127.0.1.1/,/$/c127.0.1.1\t${NEW_HOST_NAME}" /etc/hosts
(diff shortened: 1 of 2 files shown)


packer/scripts/dep.sh
```diff
@@ -9,6 +9,9 @@ pip install --upgrade setuptools
 pip install ansible==2.2.0.0
 
 # set a friendly hostname
+# by default, the hostname in both /etc/hosts & /etc/hostname are obtained from DHCP server during install.
 rm -f /etc/hostname
 echo "rackhd" > /etc/hostname
-
+#Below two lines were moved from scripts/cleanup.sh. I can't recall why they didn't go with /etc/hostname change together...
+NEW_HOST_NAME=$(cat /etc/hostname)
+sed -i  "/127.0.1.1/,/$/c127.0.1.1\t${NEW_HOST_NAME}" /etc/hosts
```
yyscamper: Please confirm this is not a typo as we usually use `127.0.0.1`

panpan0000: @yyscamper , of coz no.

typically, the /etc/hosts will be looked like below, the 127.0.1.1 is what we want to modify.
```
cat /etc/hosts
127.0.0.1       localhost
127.0.1.1       <Your_Machine_Name>

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters

```

panpan0000:
@anhou

panpan0000:
@iceiilin

PR #722, opened 2017-04-18
Add functional test for the redfish alert feature

tannoa2:


tannoa2:
test this please

tannoa2:
test this please

tannoa2:
test this please

tannoa2:
test this please

tannoa2:
test this please

tannoa2:
test this please

tannoa2:
test this please

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -0,0 +1,265 @@
...
+        threading.Thread.__init__(self)
+        pika_logger = logging.getLogger('pika')
+        if fit_common.VERBOSITY >= 8:
+            pika_logger.setLevel(logging.DEBUG)
+        elif fit_common.VERBOSITY >= 4:
+            pika_logger.setLevel(logging.WARNING)
+        else:
+            pika_logger.setLevel(logging.ERROR)
+        if fit_common.API_PORT == "9090":
+            amqp_port = fit_common.fitports()['amqp-vagrant']
```
hohene: heads up for this -  changes for 'amqp-vagrant', and 'amqp' ports variables are coming PR 715, so ampq_port will need to change to this.
 amqp_port = fit_common.fitports()['amqp_ssl']
We found the tests shouldn't need to check if api-port is 9090 inside the script to get the correct values it needs, so config files being updated to work properly.  9090 is just how our vagrant is set up.

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -0,0 +1,265 @@
...
+    MONGO_PORT = None
+    CATALOGS_COUNT = None
+    NODES_COUNT = None
+    OBMS_COUNT = None
+
+    @classmethod
+    def setUpClass(self):
+        if fit_common.API_PORT == "9090":
+            MONGO_PORT = 37017
+            IP = "10.0.2.2"  # vagrant's ip for localhost
```
hohene: See previous comment regarding amqp-port.  We should have values set up in the config file so the script doesn't need to check if we are running internal vs external.
Where does 10.0.2.2 come from - shouldn't it be fit_args().rackhd_host?
Otherwise, this test only runs against that IP looks like.

tannoa2:
test this please

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -0,0 +1,267 @@
...
+class test_alert_notification(unittest.TestCase):
+    CLIENT = None
+    MONGO_PORT = None
+    CATALOGS_COUNT = None
+    NODES_COUNT = None
+    OBMS_COUNT = None
+
+    @classmethod
+    def setUpClass(self):
+        if fit_common.API_PORT == "9090":
```
jimturnquist: We shouldn't be selecting ports based on API_PORT deployment.  Running tests outside of vagrant (off-box) should be using the stack configuration 'vagrant-remote'.  Otherwise the stack configuration should be 'vagrant'.  At this point in the test code all port assignments that are based on deployment should be selected.   For example:

I think it should look something like:
MONGO_PORT = fit_common.fitports().get('mongo_port')
IP = fit_common.fit_common.fitcfg().get('rackhd_host')

No selection based on if fit_common.API_PORT == "9090"

mongo_port would be defaulted in install_config.  It could be overridden then in the stack_config.json.

The vagrant_remote needs to be populated accordingly.

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def run(self):
+        logs.debug_7('start consuming')
+        self.channel.start_consuming()
+
+
+@attr(all=True, regression=False, smoke=True)
+class test_alert_notification(unittest.TestCase):
+    CLIENT = None
+    MONGO_PORT = None
```
jimturnquist: This declaration of MONGO_PORT does not serve a purpose.   There is no use of self.MONGO_PORT.  I believe you can just remove this.

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -0,0 +1,271 @@
...
+        self.db = self.CLIENT.pxe
+        self.node = {
+            "identifiers": ["FF:FF:FF:FF:FF:FF"],
+            "name": "FF:FF:FF:FF:FF:FF",
+            "relations": [],
+            "tags": [],
+            "type": "compute"
+        }
+        if fit_common.fitargs()['stack'] == 'vagrant_guest':
+            IP = "127.0.0.1"
```
jimturnquist: This assignment of IP won't live outside of setUpClass since IP is not declared as a global.

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -0,0 +1,271 @@
...
+        elif fit_common.fitargs()['stack'] == 'vagrant':
+            IP = "10.0.2.2"
+        else:
+            env_vagrant = False # NOQA
```
jimturnquist: This assignment of env_vagrant won't live outside of setUpClass since env_vagrant is not declared as a global.

Small python example to demostrate:
```
env_vagrant=True

class test_alert_notification():

    @classmethod
    def someClassMethod(self):
        # global env_vagrant
        env_vagrant=False

tan = test_alert_notification()

print env_vagrant
tan.someClassMethod()
print env_vagrant

```

jimturnquist: This will print True True with the commented line and True False with line uncommented.

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -0,0 +1,269 @@
+'''
+Copyright 2017 Dell Inc. or its subsidiaries.  All Rights Reserved.
+Author(s):
+'''
+
```
hohene: Please add a description as to what the test is doing here.  There isn't anything in the PR nor here to call it out.
thanks!!

PR #725, opened 2017-04-25
RAC-4793 connect fit deployed amqp ports into amqp stream monitor

stuart-stanley:
* Added ssh-helper inside plugin to be able to go on-box and add an amqp user.
* Wired handle_finalize call-all-sources and added a handler in amqp stream monitor to cause delete of amqp user
* Altered ports in install_default.json and stack_config.json get the amql_ssl port and ssh port correct for both on box (vagrant) and offbox (vagrant_remote) stack types.
* Altered nose-invoke in fit-common to add all the required ssh and amqp configure information to be passed into plugin.

Signed-off-by: Erika Hohenstein <Erika.Hohenstein@dell.com>

Note: the port names may have gotten stale, but if the PR gate passes, we can merge and rebase into the story Erika and I are on right now and fix there worse case. (i.e., no impact for anyone other than us)

@hohene @johren

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -1088,7 +1088,12 @@ def run_nose(nosepath=None):
 
     exitcode = 0
     # set nose options
-    noseopts = ['--exe', '--with-nosedep', '--with-stream-monitor']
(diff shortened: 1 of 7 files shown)


stuart-stanley:
test this please

PR #726, opened 2017-04-27
RAC-4957 FIT ssh failure in OVA Post Test

stuart-stanley:
Problem is actually misconfiguration. However, the ssh-helper should not kill the whole test suite, but should instead should only disable AMQP based Stream Monitors

This change:
* explicitly calls out the EOF exception and logs a warning indicating that this is likely a misconfiguration
* broadens the existing exception catch to prevent -any- failure from stopping the entire suite (it should only disable this particular stream-monitor)

@johren @hohene

### DIFF
--- a/test/stream-monitor/stream_sources/amqp_source.py
+++ b/test/stream-monitor/stream_sources/amqp_source.py
@@ -9,7 +9,7 @@ import optparse
 import uuid
 import gevent
 import gevent.queue
-from pexpect.pxssh import ExceptionPxssh
+from pexpect import EOF
 from .monitor_abc import StreamMonitorBaseClass
 from .stream_matchers_base import StreamMatchBase
 from .amqp_od import RackHDAMQPOnDemand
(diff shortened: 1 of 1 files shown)


hohene:
Tested the fix using vagrant and rackhd_source_install.py, which will bring out the issue.  Tests continue with the following warning now:
```
2017-04-27 15:50:51,385 INFO     Start Of Test Block: 1                                                                    root
2017-04-27 15:50:53,480 WARNING unable to set up amqp user. AMQP monitors disabled: rabbitmqctl delete_user tdd_amqp_user_b602cc6b-ddb9-4442-9ad2-f21f62c9c027 failed with something other than 'no_such_user': infra.run **.** 15676 MainProcess stream-monitor/stream_sources/amqp_source.py:__setup_generated_amqp@232 gl-main [any tests]
2017-04-27 15:50:53,483 WARNING ^^^^ if this is a deploy test, this is probably ok. If it is a real test, this is a problem. infra.run **.** 15676 MainProcess stream-monitor/stream_sources/amqp_source.py:__setup_generated_amqp@233 gl-main [any tests]
```

test/stream-monitor/stream_sources/amqp_source.py
```diff
@@ -215,13 +215,22 @@ def __setup_generated_amqp(self, generate_string):
...
             return 'amqp://{}:{}@{}:{}'.format(auser, apw, fixed.dut_ssh_host, port), auser
-        except ExceptionPxssh as ex:
+        except EOF as ex:
             # the PR in-queue behind with one sets up better logging in the sources.
             # Rather than try to pull that forward, manually grab flogging and
             # update this block in next merge.
             from flogging import get_loggers
             logs = get_loggers()
-            logs.irl.debug('unable to set up amqp user. AMQP monitors disabled: %s', ex)
+            logs.irl.warning('unable to connect ot instance to setup AMQP user. AMQP monitors disabled: %s', ex)
```
johren: Typo:  ot -> to
I won't hold up for the typo but we should probably fix this in a soon-to-follow PR since it is user visible.

PR #727, opened 2017-04-27
Turning off the notification test for now until we fix RAC-4965

tannoa2:
@hohene @johren


panpan0000:
Woo..  I link those kind of feature flag in FIT !

PR #729, opened 2017-04-28
RAC-4794-wiring-match-all-amqp-matcher-engine

stuart-stanley:
* pass in blank dns settings to rabbitmq docker instance (prevents timeouts if no dns available)
* added a retry loop in on-demand-amqp to allow container time to stabalize.
* changed 'log_facility' attribute as name of greenlet to 'greenlet_name' in flogging (log_facility was left over from orig code used as template)
* added mechanism ('handle_set_flogging') to allow sub-plugins to be updated with loggers
* removed 2nd call to handle_after_test on each sub-plugin. This was causing the test # to go up by two for each test.
* In log_stream_helper.py
** taught check code to ignore logs messages coming -from- the plugin infrastructure
** altered diag info in asserts to be more useful
* in monitor_abc.py:
** converted to gevent.queue.Queue for inject-to-finish transport (allows self-test stream-source to run asnych)
** added handle_set_flogging method and debug statements
* self_test_source.py
** added a relay greenlet to make async
* stream_matchers_results.py
** changed definition of "is_ok" from at least one ok and no errors to just no errors
* test_amqp_matchers.py:
** made tester-class to handle pre-setup test in makeSuite (allows the tests to check state from setUpClass in suite)
** tests for:
*** injecting and seeing 5 events on a tracker/processor pair
*** testing "re-seeing" of same 5 events in a second test (same tracker, new processor)
*** test re-seeing of prior-tracked events plus one "live" one
*** test NOT seeing 5 prior-tracked events when new processor created w/ "start-at" of "now"
*** test two side-by-side processor on single tracker.
* test_amqp_tracker.py
** renamed from test_amqp_mon.py
** adjusted to tracker-match records instead of old "hand coded" result dictionaries
* added .vscode to global .gitignore
* amqp_source.py -- main code for sourcing AMQP data into the matching engine
** re-did content of AMQPQueueMonitor to: (AMQPQueueMonitor is gone)
*** _AMQPQueueTracker that handles:
**** recording all messages to a exchange/routing-key
**** handles keeping track of processors (see below) attached to tracker
**** injects recorded messages into new processors (unless told not to by processor via 'start_at' param)
*** _AMQPProcessor that handles:
**** being a greenlet to watch for messages
**** dealing with timeouts for failure (waiting 10 seconds and never met criteria to complete matcher set)
**** dealing with tail-timeouts for success (waiting for 1 second after a message that allows the matcher set to complete)
*** Adjusted AMQPStreamMonitor to use trackers and processors
* updated "real test space" test to use new trackers.

Signed-off-by: Stuart Stanley <stuart.stanley@dell.com>

@hohene @johren @RackHD/corecommitters

### DIFF
--- a/test/stream-monitor/stream_sources/amqp_source.py
+++ b/test/stream-monitor/stream_sources/amqp_source.py
@@ -3,6 +3,11 @@ Copyright (c) 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
 
 A set of super-simple matchers to use to self-test the matching framework.
 """
+from gevent import monkey
+monkey.patch_dns()
+monkey.patch_time()
+monkey.patch_builtins()
+monkey.patch_select()
 import sys
 import time
 import optparse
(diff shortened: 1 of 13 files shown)


hohene:
https://rackhdci.lss.emc.com/job/RackHD/207/
this build ran and passed, started by upstream.

hohene:
test this please

test/stream-monitor/stream_sources/amqp_source.py
```diff
@@ -108,47 +126,165 @@ def _match(self, other_event):
...
+                    return results
+
+        self._logs.irl.debug('---exiting because of time---: %s -> %s', self, results)
+        res = self._match_groups.check_ending()
+        results.add_result(res)
+        self._logs.irl.debug('  final results from %s is %s', self, results)
+        return results
+
+    def start_finish(self, timeout, tail_timeout=1.0):
+        self._logs.irl.debug('start finish on %s called. timeout=%s, tail-timeout=%s', self, timeout, tail_timeout)
```
johren: Nit:  Should this be start_finish in the debug message?

stuart-stanley: It was actually more of a verb (the method name start_finish is a property built into log records). But if it jumps out, I'll change it.

If you are ok with the cp0 being removed in the next story, I'll put this fix off till then also: that way I can skip a round of rebases/re-tests.

johren: Yes, that sounds fine since the next story is coming in on the heels.

PR #731, opened 2017-05-02
Rac 4181 - FIT api/v2_0 staging of CIT scripts being converted.

hohene:
This PR is for staging the CIT api/v2_0 test scripts into a temp directory for follow on PRs.
This will allow for doing diffs between original and converted scripts, without losing the master.
The scripts in this PR are a copy of the scripts in /test/tests/api/v2_0 only, NO updates,
They won't run as part of the CIT or FIT smoke tests in this form.

Plan is:

   1. copy the original scripts into this staging area (this PR)
   2. mv the staged script and apply the converted script on top do show the diffs only for review
   3. allow the converted script to run in FIT smoke suite.

For this PR, please ignore the hound errors. This PR is copying existing scripts into a staging directory, and the hound errors will be fixed when we apply the NEW converted scripts.

### DIFF
--- a/test/tests/api/fit-staging/v2_0/catalogs_tests.py
+++ b/test/tests/api/fit-staging/v2_0/catalogs_tests.py
@@ -1,0 +1,67 @@
+from config.api2_0_config import *
+from modules.logger import Log
+from on_http_api2_0 import ApiApi as Api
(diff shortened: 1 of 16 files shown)


hohene:
@brianparry @larry-dean @johren

PR #732, opened 2017-05-02
Removing the redfish_1.0 group from the CIT smoke-test suite.

hohene:
This will stop it from running as part of the PR gate.
The redfish scripts have been convert to the FIT format and exist in the directory /test/tests/api-cit/redfish_1.0 and are running as part of the PR gate under the 'FIT' bubble.

### DIFF
--- a/test/run.py
+++ b/test/run.py
@@ -9,11 +9,9 @@ import sys
 def run_tests(group=['smoke-tests']):
 
     import tests.api.v2_0 as api_2_0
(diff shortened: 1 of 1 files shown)


johren:
At what point are these CIT tests going to be removed from source?

hohene:
@panpan0000 @brianparry @johren @derrickostertag @larry-dean @manfrednde

PR #742, opened 2017-05-08
docker: add dns for ubuntu os install, 

changev:
1. ubuntu os installation need dns-server configuration.

2. change INTERFACES to em1 for test stack.

This PR just fix this issue in our CD system.
However, how to do this configuration for users is still a issue.
@panpan0000

### DIFF
--- a/docker/dhcp/config/dhcpd.conf
+++ b/docker/dhcp/config/dhcpd.conf
@@ -9,6 +9,9 @@ max-lease-time 7200;
 ignore-client-uids true;
 deny duplicates;
 
+option domain-name-servers 172.31.128.250;
+option routers 172.31.128.254;
+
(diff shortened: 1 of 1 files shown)


pengz1:
@anhou, shouldn't a dns requirement on RackHD be documented in RackHD user guide. I never succeeded to install Ubuntu, I suspect it is because I don't have correct dns settings.

pengz1:
@anhou @iceiilin I am thinking if we should add dns requirement to our RackHD document?

anhou:
agree @pengz1  
http://rackhd.readthedocs.io/en/latest/rackhd/ubuntu_package_installation.html  http://rackhd.readthedocs.io/en/latest/rackhd/ubuntu_source_installation.html
http://rackhd.readthedocs.io/en/latest/rackhd/npm_based_installation.html
also need to be updated

PR #746, opened 2017-05-10
Add com console parameters for esxi

nortonluo:
In many rack mount servers or customized servers, there is no monitor connected, UART port are usually used as console. VMware ESXi use VGA card as console. User need pass VM kernel To support this use, I add three options: "gdbPort","logPort","debugLogToSerial". 
And there is also "comport" options. I add it to this sample file.

Jenkins depends on https://github.com/RackHD/on-tasks/pull/443

### DIFF
--- a/example/samples/install_esx_payload_full.json
+++ b/example/samples/install_esx_payload_full.json
@@ -57,6 +57,10 @@
                 }
             ],
             "installDisk": "t10.ATA_____SATADOM2DSH_TYPE_C_3SE___________________20151007AA8511725043",
+            "gdbPort": "default",
+            "logPort": "default",
+            "comport": "com1",
+            "debugLogToSerial": "1",
             "switchDevices": [
                 {
                     "switchName": "vSwitch0",


anhou:
@nortonluo I have some comment in https://github.com/RackHD/docs/pull/390, the same here

example/samples/install_esx_payload_full.json
```diff
@@ -57,6 +57,10 @@
                 }
             ],
             "installDisk": "t10.ATA_____SATADOM2DSH_TYPE_C_3SE___________________20151007AA8511725043",
+            "gdbPort": "default",
+            "logPort": "default",
+            "comport": "com1",
+            "debugLogToSerial": "1",
```
yyscamper: the literal "1" is not friendly for user to understand its meaning, I see each literal have good description in https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-esxi.json#L78, so I suggest enhancing it to some meaningful word(s)

PR #747, opened 2017-05-11
API 2.0 FIT OS bootstrap test for Linux distros

gpaulos:
New API 2.0 FIT OS bootstrap test for Linux distros. Payload can be specified by loading a file using the -extra parameter. Examples in the file header comments.

There is a default payload for CentOS 7, but the repo spec is, as always, site specific. There is no generic repo spec that will work in all situations unless some conventions are adopted. Erika and I are unsure whether there should be a default payload at all. Any thoughts?

@hohene @johren @derrickostertag @stuart-stanley @jimturnquist

### DIFF
--- a/test/tests/bootstrap/test_api20_linux_bootstrap.py
+++ b/test/tests/bootstrap/test_api20_linux_bootstrap.py
@@ -1,0 +1,132 @@
+'''
+Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
+
+Author(s):
+George Paulos
+Manfred Nde
+Erika Hohenstein
+
+This script tests arbitrary payload of the RackHD API 2.0 OS bootstrap workflows.
+The default case is running a minimum payload CentOS 7 OS install.
+Other Linux-type OS install cases can be specified by creating a payload file and specifiying it using the '-extra' argument.
+This test takes 15-20 minutes to run.
+
+Example payload file (installed in configuration dir):
+
+{"bootstrap-payload":
+    {"name": "Graph.InstallRHEL",
+        "options": {"defaults": {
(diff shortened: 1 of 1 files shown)


johren:
test this please

test/tests/bootstrap/test_api20_linux_bootstrap.py
```diff
@@ -0,0 +1,132 @@
...
+                                      '/workflows',
+                                      action='post', payload=PAYLOAD)
+        if result['status'] == 201:
+            # workflow running
+            log.info_5(" TaskID: " + result['json']['instanceId'])
+            log.info_5(" Payload: " + fit_common.json.dumps(PAYLOAD))
+            workflowid = result['json']['instanceId']
+        else:
+            # workflow failed with response code
+            log.error(" TaskID: " + result['text'])
```
johren: Is this really a TaskID?

test/tests/bootstrap/test_api20_linux_bootstrap.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@attr(all=False)
+class api20_bootstrap_linux(fit_common.unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # Get the list of nodes
+        NODECATALOG = fit_common.node_select()
+        # Select one node at random
+        cls.__NODE = NODECATALOG[random.randint(0, len(NODECATALOG) - 1)]
```
gpaulos: Need to specify a node ID or node index when running in parallel.

gpaulos:
Requested changes pushed up.

johren:
Paired with @hohene

PR #752, opened 2017-05-16
Added -nodeindex parameter to fit_common.py

gpaulos:
Added new argument -nodeindex to fit_common.py as an additional modifier to the node_select() method. 

This was requested as a tool to select target nodes for Jenkins OS installations when nodeid or MAC address is undefined.

@larry-dean @johren @hohene

### DIFF
--- a/test/common/fit_common.py
+++ b/test/common/fit_common.py
@@ -336,6 +336,8 @@ def mkargs(in_args=None):
     arg_parser.add_argument("-sku", default="all",
                             help="node SKU name, example: Quanta-T41, default=all")
     group = arg_parser.add_mutually_exclusive_group(required=False)
+    group.add_argument("-nodeindex", default="all",
+                       help="node index, integer 0 or greater")
     group.add_argument("-obmmac", default="all",
                        help="node OBM MAC address, example:00:1e:67:b1:d5:64")
     group.add_argument("-nodeid", default="None",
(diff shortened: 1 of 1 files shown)


test/common/fit_common.py
```diff
@@ -867,7 +869,7 @@ def node_select():
             if str(fitargs()['sku']) in json.dumps(skuentry):
                 skuid = skuentry['id']
         # Collect node IDs
-        catalog = rackhdapi('/api/2.0/nodes')
+        catalog = rackhdapi('/api/2.0/nodes?type=compute')
```
hohene: We need to do a sort on the returned nodes for using the 'index' command.
Different calls of run_tests.py that use the passed in index needs to use the node list in a specific order, otherwise, index is random.
ie. sort the return of api/2.0/nodes into 'catalog'

test/common/fit_common.py
```diff
@@ -867,7 +869,7 @@ def node_select():
...
         if skumap['status'] != 200:
```
gpaulos: Yes, skumap['status'] is an error. Looks like that's been there since the beginning. Will revise.

catalog['json'] is called several times, but 'catalog' variable is the result from one API call. We could create yet another variable to hold catalog['json'] , but this is not really gaining us anything. I say leave it.

gpaulos:
Added sort routine to -nodeindex function and fixed catalog['status'] bug.

PR #753, opened 2017-05-16
RAC-4994-amqp-rediscovery-conversion

stuart-stanley:
* Common-changes:
** remote-shell was not honoring the ssh-port. Changed the code to do so AND include that in the check for a truly local command
* Added requirements.txt items:
** updated kombu to latest (matches plugin!)
** added ipaddress and netifaces to help making detection of actual test-host IP address even possible
* monitor base-class changes:
** threaded logging throughout
** threaded in "allow complete miss" logic
* stream_matchers_base and monitor_abc
** taught add_result recorder to understand lists
** enhanced is_empty to check -into- a list of results and call it empty if it is just a list of empties.
** threaded logging in
** added __str__ to a number of items to make diag easier
** added an _validate step. _match still handles "does this apply?" while _validate handles "and does it make sense..."
** taught entire check_event chain to handle lists of results returned (a list occurs when validation failuers happen)
** taught entire check_event chain to handle allow_complete_miss concept. The default is that NOT an event in an ordered grouping causes an out-of-order result to be attached to the next matcher. But sometimes we only care that x.y.z occur in order and don't care if anything else drops between x or y or z.
* test_amqp_node_rediscovery.py nows works with the stream monitor. Big changes:
** includes a sample class to determine the test-host IP address. I'll be writing up a story to move this to common and fully test across all known deploys.
** removed all use of globals etc etc.
** removed linear amqp-worker.
** made tests for each block-progress-step:
*** test_rediscover_pick_node
*** test_rediscover_check_obm
*** test_rediscover_reboot_kickoff
*** test_rediscover_reboot_graph_amqp_flow -- this is the core of graph progress tracking
*** test_rediscover_discover_rebooted_node
*** test_rediscover_discover_sku_assign
*** test_rediscover_discover_node_registration
*** test_rediscover_discover_hook -- this checks the rackhd post-back results
*** test_rediscover_discover_obm_settings
** moved validation into a common routine that can report multiple failures via _verify
** added nosedeps were sequencing was needed (a fair amount in this case)
** used a class-level run-context along with the nosedeps to handoff critical data
* stream-monitor test/test_matchers.py
** added a preceding whitespace to the match result type check (the addition of the __str__ diag code was causing an extra hit here)
** assert_errors in __assert_Basic_results when an ok is expected (ends up being way more useful than just the the assert saying 'ok' wasn't expected)
** added a test for the allow_complete_miss flag
** fixed a couple if in-text typeos
* stream_matchers_results.py:
** added a dump() to the stream-results for diag
** added __str__ to same
** added Validation matcher-results (Missmatch and MissingField)
* amqp_source got more changes than expected:
** Added a _KeyedConsumerHandler rather than having the ampq-server-wrapper keep track of all queues. Turns out you can either get everything on an unnamed consumer as a Kombu message (which you can ack) OR you can set the consumer-id... and get a raw amqplib message, which you CAN'T ack! Pulling this out gives the equiv of consumer-id to event callback lists w/o using the consumer-ids
** added a _validate method to the AMQP-matcher to check fields etc. This could be a template validation test-harness wide.
** we now wait till we move to finishing mode before even looking at a single message. There is other wiring that would have to happen to add matchers "on the fly" and this isn't the time for them!
** change from time.time to datetime.now and timedeltas (much pretties/readable time-deltas)
** lots more logging!

@hohene @johren

### DIFF
--- a/test/stream-monitor/stream_sources/amqp_source.py
+++ b/test/stream-monitor/stream_sources/amqp_source.py
@@ -10,37 +10,94 @@ monkey.patch_builtins()
 monkey.patch_select()
 import re
 import sys
-import time
 import optparse
 import uuid
 import gevent
 import gevent.queue
-import json
 from pexpect import EOF
-from datetime import datetime
+from datetime import datetime, timedelta
 from .monitor_abc import StreamMonitorBaseClass
 from .stream_matchers_base import StreamMatchBase
-from .stream_matchers_results import StreamRunResults
+from .stream_matchers_results import StreamRunResults, MatcherValidationMissmatch, MatcherValidationMissingField
 from .amqp_od import RackHDAMQPOnDemand
 from .ssh_helper import SSHHelper
 from kombu import Connection, Producer, Queue, Exchange, Consumer
 
 
+class _KeyedConsumerHandler(object):
+    _keyed_consumers = {}
+
+    @classmethod
+    def get_keyed_consumer(cls, logs, connection, exchange, routing_key, queue_name, event_cb):
+        mname = "ex={} rk={} qn={}".format(exchange, routing_key, queue_name)
+        if mname not in cls._keyed_consumers:
+            new_one = _KeyedConsumerHandler(logs, connection, mname, exchange, routing_key, queue_name)
+            cls._keyed_consumers[mname] = new_one
+        cls._keyed_consumers[mname].add_new_event_handler(event_cb)
+        return cls._keyed_consumers[mname]
+
+    @classmethod
+    def test_helper_finalize_cleanup(cls):
+        cls._keyed_consumers = {}
+
+    def __init__(self, logs, connection, name, exchange, routing_key, queue_name):
+        self.__logs = logs
+        self.__ignore_some_stuff = False
+        self.name = name
+        self.__event_callbacks = []
+        if queue_name is None:
+            queue_name = ''
+            exclusive = True
+        else:
+            exclusive = False
+        chan = connection.channel()
+        ex = Exchange(exchange, 'topic', channel=chan)
+        queue = Queue(exchange=ex, routing_key=routing_key, exclusive=exclusive)
+        consumer = Consumer(chan, queues=[queue], callbacks=[self.__message_cb])
+        consumer.consume()
+        self.exchange = ex
+
+    def add_new_event_handler(self, event_cb):
+        self.__event_callbacks.append(event_cb)
+
+    def __message_cb(self, body, msg):
+        if self.__ignore_some_stuff:
+            if "heartbeat" in msg.delivery_info['routing_key']:
+                msg.ack()
+                return
+            if msg.delivery_info['routing_key'].startswith('http'):
+                msg.ack()
+                return
+            if msg.delivery_info['routing_key'].startswith('polleralert'):
+                msg.ack()
+                return
+
+        self.__logs.idl.debug(
+            'Inbound AMQP msg. %s (delivery_info=%s, content_type=%s, properties=%s, body=%s)',
+            msg, msg.delivery_info, msg.content_type, msg.properties, body)
+        for event_cb in self.__event_callbacks:
+            try:
+                event_cb(msg, body)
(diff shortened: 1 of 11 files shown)


stuart-stanley:
please test me

stuart-stanley:
test me please

stuart-stanley:
test this please

johren:
test this please

stuart-stanley:
test this please

johren:
test this please

johren:
test this please

stuart-stanley:
@johren : it looks like the cit stuff is using some really low-level amqp direct-lib stuff. I don't have time this instant, but I'll look at backing the kombu-update out as soon as I can though and see if the new stuff still works with it. If so, I can push kombu back to 3.0.x until we can unwire the low-level calls for new stuff.

stuart-stanley:
@hohene and @johren : CIT based pip package conflicts have been resolved and the new conversion was retested.

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -9,138 +9,222 @@
...
-        logs.debug_7('Pika connection timeout')
-        if self.connection.is_closed is False:
-            self.channel.stop_consuming()
-            self.connection.close()
-        self.thread_stop = True
 
-    def run(self):
-        logs.debug_7('start consuming')
-        self.channel.start_consuming()
+class _TstHostInterfacer(object):
```
hohene: Let's move this into fit_common.py so other test scripts can make use of it.  I've made updates already in my local sandbox for doing this so can review them with you. :-)

test/tests/amqp/test_amqp_node_rediscover.py
```diff
@@ -277,185 +298,202 @@ def _process_web_message(self, timeout):
...
-        td.setDaemon(True)
-        td.start()
+        self.__set_run_context('reboot_graphid', graphid)
+
+    @depends(after='test_rediscover_reboot_kickoff', before='test_rediscover_node_delete')
+    def test_rediscover_reboot_graph_ampq_flow(self):
+        nodeid = self.__get_run_context('start_nodeid')
+        graphid = self.__get_run_context('reboot_graphid')
+        self.__qproc.push_group(ordered=True)
+        # todo ^^^^ erika to help name that method!
```
hohene: Add a comment above the "push_group" - # following amqp messages should be received in order, so setting push_group ordered = True.  
Suggested names "push_ordered_matching", "set_message_ordering"  or leave as is, if we add a comment to explain.

test/tests/amqp/test_amqp_node_rediscover.py
stuart-stanley: Let's open another story to do that. It makes total sense, but making this routine "public" is really a different thing than this PR.

test/stream-monitor/stream_sources/amqp_source.py
```diff
@@ -246,15 +349,15 @@ def process_tracked_record(self, tracked_record):
     def match_any_event(self, description=None, min=1, max=1):
         if description is None:
             description = "match-any(rk={},min={},max={}".format(None, min, max)
-        m = _AMQPMatcher(route_key=None, description=description, min=min, max=max)
+        m = _AMQPMatcher(self._logs, route_key=None, description=description, min=min, max=max)
         self._add_matcher(m)
 
-    def match_on_routekey(self, description=None, routing_key=None, min=1, max=1):
+    def match_on_routekey(self, description, routing_key=None, min=1, max=1, validation_block=None, match_CB=None):
```
stuart-stanley: Ah. Actually I need to fix test_amqp_heartbeat.py (forcing the use of description was an important step -- otherwise it is impossible to tell what is matching :( )

test/tests/amqp/test_amqp_node_rediscover.py
stuart-stanley: We can also push another unordered group (at least in theory!). The comment may have to do, since I'm coming up with nothing for how to name this right :(

stuart-stanley:
test this please

PR #757, opened 2017-05-17
API 2.0 FIT OS bootstrap test for ESXi

gpaulos:
New API 2.0 FIT OS bootstrap test for ESXi. Payload can be specified by loading a file using the -extra parameter. Examples in the file header comments.

There is a default payload for ESXi 6, but the repo spec is, as always, site specific. There is no generic repo spec that will work in all situations unless some conventions are adopted. 

Paired with Larry Dean.

@hohene @johren @larry-dean

### DIFF
--- a/test/tests/bootstrap/test_api20_esxi_bootstrap.py
+++ b/test/tests/bootstrap/test_api20_esxi_bootstrap.py
@@ -51,9 +51,9 @@ if config:
 
 
 # this routine polls a workflow task ID for completion
-def wait_for_workflow_complete(instanceid, start_time, waittime=900, cycle=30):
+def wait_for_workflow_complete(instanceid, start_time, waittime=1800, cycle=30):
     log.info_5(" Workflow started at time: " + str(start_time))
-    while start_time - time.time() < waittime:  # limit test to 30 minutes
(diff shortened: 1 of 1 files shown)


test/tests/bootstrap/test_api20_esxi_bootstrap.py
```diff
@@ -0,0 +1,128 @@
...
+# this routine polls a workflow task ID for completion
+def wait_for_workflow_complete(instanceid, start_time, waittime=900, cycle=30):
+    log.info_5(" Workflow started at time: " + str(start_time))
+    while start_time - time.time() < waittime:  # limit test to 30 minutes
+        result = fit_common.rackhdapi("/api/2.0/workflows/" + instanceid)
+        if result['status'] != 200:
+            log.error(" HTTP error: " + result['text'])
+            return False
+        if result['json']['status'] in ['running', 'pending']:
+            log.info_5("{} workflow status: {}".format(result['json']['injectableName'], result['json']['status']))
```
larry-dean: Flogger messages should not use format or " + ". Standard "C" formatting is to be used. I'm not going to hold up the PR for this though.

johren:
test this please

PR #760, opened 2017-05-18
Fix start_time bug in Linux OS Install

gpaulos:
Found a math problem with the workflow timer in the Linux OS install test. We had the expression backwards and the timer wasn't timing out. 

Single line fix.

@hohene @johren


johren:
test this please

johren:
test this please

johren:
test this please

PR #762, opened 2017-05-19
RAC-5000 Windows OS Install test script

gpaulos:
Windows OS Install FIT test script for arbitrary payloads.

The Windows OS install requires additional configuration of the RackHD server as outlined in the header notes.

Payload can be specified by loading a file using the -extra parameter. Examples in the file header comments.

There is a default payload for Windows 2012 but the repo spec is, as always, site specific. There is no generic repo spec that will work in all situations unless some conventions are adopted.

The default payload contains a dummy activation key. We cannot publish this script with a valid activation key. If the default payload is used, Windows Install will ask for a valid key on the console, which will continue if entered manually.. If fully automated operation is required, then a valid activation key must be included in an external payload file. 

The Windows Install workflow execution time varies from 1800 to 2900 seconds in the several runs we made. We set the timeout at 3200. We presume this will be enough for regression testing but may need to be increased if its really slow in the production environment.

Paired with Vaidehi   Pathak

@hohene @johren @larry-dean

### DIFF
--- a/test/tests/bootstrap/test_api20_windows_bootstrap.py
+++ b/test/tests/bootstrap/test_api20_windows_bootstrap.py
@@ -1,0 +1,137 @@
+'''
+Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
+
+This script tests arbitrary payload of the RackHD API 2.0 OS bootstrap workflows.
+The default case is running a minimum payload Windows OS install.
+Other Windows-type OS install cases can be specified by creating a payload file and specifiying it using the '-extra' argument.
+This test takes 30-45 minutes to run.
+
+Example payload file (installed in configuration dir):
+
+{"bootstrap-payload":
+    {"name": "Graph.InstallWindowsServer",
+           "options": {"defaults": {"version": "2012",
+                                    "repo": "http://172.31.128.1:8080/repo/winpe",
+                                    "smbRepo": "\\\\172.31.128.1\\windowsServer2012",
+                                    "productkey": "XXXXX-XXXXX-XXXXX-XXXXX-XXXXX",
+                                    "username": "rackhduser",
+                                    "password": "RackHDRocks",
+                                    "smbUser": "vagrant",
+                                    "smbPassword": "vagrant"}}}
+}
+
+Example command line using external payload file:
+
+python run_tests.py -stack 4 -test tests/bootstrap/test_api20_windows_bootstrap.py -extra base_windows_2012_install.json
+
+RackHD Windows installation workflow requires special configuration of the RackHD server:
+- A customized WinPE environment installed on RackHD server as documented here:
+        https://github.com/RackHD/on-tools/tree/master/winpe
+- Samba installed on the RackHD server and configured as documented here:
+        http://rackhd.readthedocs.io/en/latest/rackhd/install_os.html?highlight=os%20install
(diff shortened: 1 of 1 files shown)


test/tests/bootstrap/test_api20_windows_bootstrap.py
```diff
@@ -0,0 +1,137 @@
...
+def wait_for_workflow_complete(instanceid, start_time, waittime=3200, cycle=30):
+    log.info_5(" Workflow started at time: " + str(start_time))
+    while time.time() - start_time < waittime:  # limit test to waittime seconds
+        result = fit_common.rackhdapi("/api/2.0/workflows/" + instanceid)
+        if result['status'] != 200:
+            log.error(" HTTP error: " + result['text'])
+            return False
+        if result['json']['status'] in ['running', 'pending']:
+            log.info_5("{} workflow status: {}".format(result['json']['injectableName'], result['json']['status']))
+            fit_common.time.sleep(cycle)
```
johren: Remove the fit_common.  Not needed.

test/tests/bootstrap/test_api20_windows_bootstrap.py
```diff
@@ -0,0 +1,137 @@
...
+        # launch workflow
+        workflowid = None
+        result = fit_common.rackhdapi('/api/2.0/nodes/' +
+                                      self.__class__.__NODE +
+                                      '/workflows',
+                                      action='post', payload=PAYLOAD)
+        if result['status'] == 201:
+            # workflow running
+            log.info_5(" InstanceID: " + result['json']['instanceId'])
+            log.info_5(" Payload: " + fit_common.json.dumps(PAYLOAD))
```
johren: Don't use json from fit_common.  Import json directly here.

test/tests/bootstrap/test_api20_windows_bootstrap.py
```diff
@@ -0,0 +1,137 @@
...
+            workflowid = result['json']['instanceId']
+        else:
+            # workflow failed with response code
+            log.error(" InstanceID: " + result['text'])
+            log.error(" Payload: " + fit_common.json.dumps(PAYLOAD))
```
johren: Don't use json from fit_common.  Import json directly here.

test/tests/bootstrap/test_api20_windows_bootstrap.py
```diff
@@ -0,0 +1,137 @@
...
+# this routine polls a workflow task ID for completion
+def wait_for_workflow_complete(instanceid, start_time, waittime=3200, cycle=30):
+    log.info_5(" Workflow started at time: " + str(start_time))
+    while time.time() - start_time < waittime:  # limit test to waittime seconds
+        result = fit_common.rackhdapi("/api/2.0/workflows/" + instanceid)
+        if result['status'] != 200:
+            log.error(" HTTP error: " + result['text'])
+            return False
+        if result['json']['status'] in ['running', 'pending']:
+            log.info_5("{} workflow status: {}".format(result['json']['injectableName'], result['json']['status']))
```
johren: The comment that Larry made in the earlier PR.  Flogger messages should not use format or " + ". Standard "C" formatting is to be used.  Rather than propogating further, should fix it here.

johren:
The fit_common comments haven't been addressed.  I'll make a new PR to clean these all up.

PR #763, opened 2017-05-19
Rac 5182 mongo db role fix

michaelsteven:
Not the most elegant fix, but will probably do the job.  Just throwing this out there as a possible fix, take it or close it.  I will build off of my fork in the meantime until this bug is resolved.

It successfully installs MongoDB (default version 2.6.10) on Ubuntu 16.04 without failing.  I can't install the version desired because it is not available on 16.04.


michaelsteven:
Closing... changing my build to use ubuntu 14.04 until newer mongo is supported.

PR #764, opened 2017-05-22
remove ucsm-ip to disable UCS tests

geoff-reid:
This is a temporary fix for https://rackhd.atlassian.net/browse/RAC-5184.  The Docker and OVA tests are failing the UCS tests because the UCS-service is not included in either of these builds.  This change will disable the UCS tests.  They will be re-enabled once the UCS-service is include in the RackHD Docker and OVA builds.  

@RackHD/corecommitters


panpan0000:
👍

PR #768, opened 2017-05-24
Ansible fix

michaelsteven:
This enhancement allows a developer to specify an "organization" in addition to the branch variable.  Previously the ansible role would only pull from the rackhd organization.  

It is written so that a developer can fork only the repos they want to work on into their "organization" (i.e. http://www.githhub.com/michaelsteven - where michaelsteven is the organization).  

The role now does a check to see if the repo exists in the specified organization, and if not it will fall back pull from the "rackhd".  It also does a check to see if the branch exists, and if not it will fall back to use "master".

Also changed is the dependence on "submodules" is removed.  It no longer needs them.

### DIFF
--- a/packer/ansible/roles/remoterepos/tasks/main.yml
+++ b/packer/ansible/roles/remoterepos/tasks/main.yml
@@ -6,7 +6,7 @@
 # Same is true for branches.  A developer may choose to make a branch on only one or more repos 
 # If the passed in branch is not found in the github repo, it will know to pull from the master branch
 
- 
+
 # Clone down the RackHD Repo (without submodules)
 # NOTE: fall back to default "rackhd" repo if it does not exist in the specified repo
 
@@ -139,6 +139,14 @@
 
 
 # Continue with the NPM installation of the repos
+
+- name: install dev libraries for NPM installs
+  apt: pkg={{ item }} state=installed
+  with_items:
+    - libkrb5-dev
+    - unzip
+  sudo: yes
+
 
 - name: Npm install Repos
   npm: path="{{ ansible_env.HOME }}/src/{{ item }}"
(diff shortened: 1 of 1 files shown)


iceiilin:
add BRI experts for review @panpan0000 @PengTian0 @changev

michaelsteven:
Ok I discovered a problem when deploying with these changes to Ubuntu 14.04.  The ansible version for trusty is 1.5.4, and the git "recursive" option isn't available to until ansible 1.6.   This means while it works fine when using it with Ubuntu 16.04 but not 14.04 (without backports).   I haven't been able to find out how to tell git to NOT pull the submodules in the old ansible.    Looking at options, don't approve yet.

packer/ansible/roles/remoterepos/tasks/main.yml
```diff
@@ -1,46 +1,152 @@
 ---
+# This role supports developers who want to pull code from one or more repositories forked into their github organization.
```
panpan0000: can you please write a **usage** guide on the top, to tell which parameters does this new playbook accept ?
- organization 
- branch
- repos ?

packer/ansible/roles/remoterepos/tasks/main.yml
```diff
@@ -1,46 +1,152 @@
...
-    - on-tftp
-    - on-dhcp-proxy
-    - on-taskgraph
-    - on-syslog
-    - on-wss
-    - on-tools
-    - ucs-service
-
-- name: Reset to the branch specified
-  shell: git reset --hard {{ branch }}
```
panpan0000: I noticed this step is gone. although a brand new clone will ensure local HEAD aligns master HEAD, so I think this step is not necessary.. just raise this up in case you delete by accident.

michaelsteven:
Do not commit this PR yet.  Still working out the problems with using on 14.04 that carries a really old version of Ansible.  I feel it is worth having once this is fixed for build system flexibility.

anhou:
@michaelsteven  is this PR still valid, if not, we could close it

PR #775, opened 2017-06-01
RAC-4965 - add additional stack configurations for running pipelines

hohene:
Updated the stack-config.json file with additional stacks.
Added docker, ova, and vagrant_ucs test stacks.
This allows us to configure the test beds differently for the deployments and run tests using proper setup.
Also allows easier skipping of tests, if we need, if not running in an expected deployment.

updates to on-build-config in separate PRs.

### DIFF
--- a/test/tests/rackhd20/test_rackhd20_api_notifications.py
+++ b/test/tests/rackhd20/test_rackhd20_api_notifications.py
@@ -77,7 +77,7 @@ class test_alert_notification(unittest.TestCase):
     def setUpClass(self):
         if fit_common.fitargs()['stack'] == 'vagrant_guest':
             IP = "127.0.0.1"
-        elif fit_common.fitargs()['stack'] == 'vagrant':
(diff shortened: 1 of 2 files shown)


hohene:
@gpaulos @johren @patelb10 @geoff-reid @tannoa2 @panpan0000

test/tests/rackhd20/test_rackhd20_api_notifications.py
```diff
@@ -25,11 +25,14 @@
 from pymongo import MongoClient
 from bson.objectid import ObjectId
 
-
+# Check the running test environment
+if fit_common.fitargs()['stack'] in ['vagrant_guest', 'vagrant']:
```
johren: Does this need to include vagrant_ucs?

hohene: yep, I can add that.

PR #792, opened 2017-06-12
Fix: timeout to run test_api_20_ucs_discover_and_catalog_all

PengTian0:
I'm running FIT with docker and sometimes the case "test_api_20_ucs_discover_and_catalog_all" failed with message: timeout for 180s.
Actually, the graph "ucs discovery" is succeed.
But the graph takes about 187s and the test case only waits for 180s.
@geoff-reid  @anhou @iceiilin @panpan0000

### DIFF
--- a/test/tests/ucs/test_rackhd_api_ucs.py
+++ b/test/tests/ucs/test_rackhd_api_ucs.py
@@ -196,7 +196,7 @@ class rackhd_ucs_api(unittest.TestCase):
 
         return len(api_data["json"]["ServiceProfile"]["members"])
 
-    def wait_utility(self, id, counter, name):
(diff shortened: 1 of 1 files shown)


test/tests/ucs/test_rackhd_api_ucs.py
```diff
@@ -196,7 +196,7 @@ def get_service_profile_count(self):
 
         return len(api_data["json"]["ServiceProfile"]["members"])
 
-    def wait_utility(self, id, counter, name):
+    def wait_utility(self, id, counter, name, max_wait=self.MAX_WAIT):
```
pengz1: Have you tried this? It should not work. You should use MAX_WAIT.

PengTian0:
test this please

pengz1:
+1

PR #781, opened 2017-06-05
RAC-5128 - Modify sel alert test to use fit

larry-dean:
* Modify to support new AMQP

PR #793, opened 2017-06-13
RAC-5128 - Modify sel alert test to use fit

larry-dean:
* Modify to support new AMQP
* Moved sel_alert_poller_tests.pytests/api-cit/v2_0
* Required changes to fit_common to allow multiple ports for scp
* Correct ssh_helper.py to use StrictHostKeyChecking=no
* Correct ssh_helper.py to use UserKnownHostsFile=/dev/null
* Change ipmi SEL log inject error to use button push and added event to skupack
* Fixed mispelled impi to ipmi.
* Added method to check and calculate the number of available sel log entries on the node.
  Physical hardware works differently than the simulated nodes and needs to be handled properly.
* Added a clear of the inserted SEL entries at completion of script


larry-dean:
This is a new version of PR: https://github.com/RackHD/RackHD/pull/781

@stuart-stanley @hohene @johren

larry-dean:
Correction for Jenkins failure pushed up

PR #797, opened 2017-06-14
Configuration modification for dell SMI services

prashanthlgowda:
Configuration modification for dell SMI services


michaelsteven:
Changes look good.  These changes are to fix where we were maintaining two sets of config files for the SMI microservices (internal and external/github) and as a result the github ones became stale.  We will now begin using github config files exclusively.  This change should not have any impact on RackHD itself, just affects the SMI microservices.

akashkw:
Changes look good.

PR #798, opened 2017-06-14
Moving SMI rackhd config file to opt monorail directory

michaelsteven:
This PR moves the SMI microservice Rackhd config settings file underneath the /opt/monorail directory.  There are two copies of the file.  One for RackHD installed via ansible, and one for RackHD installed as docker containers. 

This PR should go around the same time as a PR in on-core that changes the configuration.js to point to the new file location.


michaelsteven:
This is the related pull request in on-core:
https://github.com/RackHD/on-core/pull/282

PR #800, opened 2017-06-15
Adding missing callback/webhook router

michaelsteven:
Adding callback/webhook router that is missing from the docker version of monorail/config.json.  It is present in the ansible version of the monorail/config.json.  This is needed for the SMI inventory microservices to asynchronously call back with data when an inventory completes.  Not having this is breaking SMI Inventory callbacks when running the wsman discovery/inventory taskgraphs from a dockerized RackHD.


docker/monorail/config.json
```diff
@@ -29,6 +29,14 @@
           "proxiesEnabled": true,
           "authEnabled": false,
           "routers": "southbound-api-router"
+      },
+	  {
+          "address": "0.0.0.0",
```
anhou: In our old discussion, the IP for southbound is limited to 172.31.128.1 which communicate with nodes, and northbound could be from all NICs. I found https://github.com/RackHD/RackHD/blob/master/packer/ansible/roles/monorail/files/config.json also use '0.0.0.0' for 9988 port, could you also help update to '172.31.128.1' there.

PR #802, opened 2017-06-16
Add config for running docker locally

PengTian0:


PengTian0:
@hohene @anhou @iceiilin @panpan0000

test/config/stack_config.json
```diff
@@ -64,6 +64,23 @@
...
+    "ucs_service_uri" : "https://localhost:7080",
+    "ucsm_config_file" : "./tests/ucs/ucsm_config.xml",
+    "install-config": {
+      "ports": {
+        "mongo_port": 27017,
+        "http": 9090,
+        "https": 8443,
+        "amqp_ssl": 5672,
+        "amqp_nossl": 5671,
+        "ssh": 22
```
stuart-stanley: Is this really the ssh-port we reach the docker container with? When "ssh -p 22 localhost" is done from the host executing the tests, you need to log into the _container_....

PengTian0: Thanks for your reviewing!
"22" is not the ssh-port to reach the docker container, it's the ssh-port to log into the host which is running the container.
 Usually, we log into a container by ```docker exec -it <container> /bin/bash``` .
 If we really need to log into a container with "ssh", we need to install a ssh server in the container. But I just don't think it's necessary for RackHD container.

I guess it's enough for FIT to log into the host which is running a RackHD container.
Because I tried to run our continuous test pipeline with the config and it passed: http://rackhdci.lss.emc.com/job/on-build-config/463

PengTian0: Thank you for your review!
"22" is not the ssh-port to reach the docker container, it's the ssh-port to log into the host which is running the container.
And I'm running FIT with below command and it works well :
```
python run_tests.py -test tests -group smoke -stack docker_local_run --sm-amqp-url amqp://guest:guest@localhost:5672 -v 4 -xunit
```

From my understanding , the reason for FIT and CIT to log into the rackhd host  is that they need to run commands like: ```ipmitool, rabbitmqctl```
And RackHD depends on the tools ipmitool and rabbitmq server.
FIT and CIT can reuse them.

But I don't think it's a good idea to run smoke test of RackHD depending  on the environment of RackHD.
FIT and CIT should run with only the access of the API , such as http://localhost:9090, amqp://guest:guest@localhost:5672.
About ipmitool, maybe we can setup a docker image for FIT and CIT which installed ipmitool.

 If we really want to log into a container with "ssh", we need to install a ssh server in the container. But I just don't think it's a good idea. There is a document ["If you run SSHD in your Docker containers, you're doing it wrong!"](https://jpetazzo.github.io/2014/06/23/docker-ssh-considered-evil/
)

test/tests/api-cit/v2_0/sel_alert_poller_tests.py
```diff
@@ -268,9 +268,9 @@ def test_full_sel(self):
 
         sel_file = self.__create_selEntries_file(available_sel_entries)
         fit_common.remote_shell('ls')
-        fit_common.scp_file_to_host(sel_file)
+        sel_file_path = fit_common.scp_file_to_host(sel_file)
 
-        self.__run_ipmitool_command(bmc_ip, "sel add ~/selError.txt")
+        self.__run_ipmitool_command(bmc_ip, "sel add {0}".format(sel_file_path))
```
pengz1: sel_file_path is parsed by os.path.basename, you may want "~/{}".format(sel_file_path)

test/config/stack_config.json
stuart-stanley: "FIT and CIT should run with only the access of the API , for example:
http://localhost:9090, amqp://guest:guest@localhost:5672" -- first, there is only FIT, no CIT. We've been trying to collapse the two together, but honestly spend maybe a 3rd of our time dealing with unify the testing.

More importantly, the FIT framework is meant to be able to DO all the environment mapping for our testing. Continuous testing like PR gates, deployment spreads, per-merge, extended and so on. Being able to go into the box under any deployment to allow tests to execute anywhere as well as do things like inject errors and so on. (virtualenv isn't part of this, to my awareness). I TOTALLY AGREE ABOUT RUNNING SSH IN DOCKER, but instead of removing a bunch of our capabilities, look at providing primitives to do "on box" operations that do ssh when ssh is there, and "docker exec ..." for (example) docker. 

At the very least in this case, 'ssh' is the wrong name for the field. It should be something like 'image_controller_ssh_port' (along with 'image_controller_host' (and auth data) -- other examples might be the v-center host for deploying OVAs). I.E 'ssh' is used throughout the testing environment to mean 'how to ssh into the RackHD-under-test"

PR #805, opened 2017-06-16
Adding .env file for smi docker-compose file

michaelsteven:
This .env file is at the RackHD/docker/dell service so it is used by the docker-compose.yml at that level, setting some default values.


panpan0000:
:+1

PR #806, opened 2017-06-19
rac-5400 - Failed at case SELPollerAlertTests

larry-dean:
* Added logic skip test of AMQP server not active


larry-dean:
@johren @jimturnquist @BrentHiggins

PR #809, opened 2017-06-21
Change node_select default behavior to always select node with sku's assigned

larry-dean:
* allow_unknown_nodes option allows nodes without sku assignment to be selected


larry-dean:
@jimturnquist @BrentHiggins @hohene @derrickostertag

jimturnquist:
This is being set aside.  Closing.

PR #811, opened 2017-06-21
Use port 8080 instead of 9080 for the image repo

johren:
This PR is just to try running PR gates while accessing the image repo on port 8080 instead of port 9080.

### DIFF
--- a/test/config/install_centos65_minimum.json
+++ b/test/config/install_centos65_minimum.json
@@ -2,7 +2,7 @@
     {"name": "Graph.InstallCentOS",
         "options": {"defaults": {
         "version": "6.5",
-        "repo": "http://172.31.128.1:9080/repo/centos/6.5",
+        "repo": "http://172.31.128.1:8080/repo/centos/6.5",
         "rootPassword": "1234567",
         "hostname": "rackhdnode",
         "users": [{"name": "rackhduser",
--- a/test/config/install_centos70_minimum.json
+++ b/test/config/install_centos70_minimum.json
@@ -2,7 +2,7 @@
     {"name": "Graph.InstallCentOS",
         "options": {"defaults": {
         "version": "7",
-        "repo": "http://172.31.128.1:9080/repo/centos/7.0",
+        "repo": "http://172.31.128.1:8080/repo/centos/7.0",
         "rootPassword": "1234567",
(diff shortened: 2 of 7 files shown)


PengTian0:
Could you please explain the reason for accessing image repo on port 8080?
I think it's better to make the port as a config parameter of stack_config.json and the payloads are generated according to the ports of  stack_config.json

johren:
This is just to try out the other port to see if it improves the performance of the OS install.  This is not currently intended to be checked in.  The CIT tests were using port 8080 while the new FIT tests are using port 9080.

johren:
test this please

anhou:
@johren I'm a little surprised that 8080 was used for southbound OS installation file downloading, I didn't notice previously. I think 9080 is recommended for southbound in original purpose, but it's no problem from me that we could have a try 8080 based on it's a factor for change from CIT to FIT. it's better change back to 9080 if it's not the cause of recent OS install problem

anhou:
@johren I also found that https://github.com/RackHD/RackHD/blob/master/test/config/install_centos70_minimum.json divorce with https://github.com/RackHD/RackHD/blob/master/example/samples/install_centos_7_payload_minimal.json which is used in previous test, I'm not sure the history that why FIT use another sets of payload, but from the words 'minmal' or 'minimum' itself, it means that it wants to use the least keys to validate os installation , and these payload examples also will be provided to users/developers, and also it's a base alignment when users/developers report os installation bugs

iceiilin:
I would prefer to use "{{ file.server }}" in payload instead of hard code "IP:port" in payload. The previous one will generate correct url in rackhd using data in /opt/monorail/config.json. you may refer to http://rackhd.readthedocs.io/en/latest/rackhd/static_file_server.html?highlight=file.server

amymullins:
Adding to @anhou 's feedback, was specifically tied to the payload in the example ( https://github.com/RackHD/RackHD/blob/master/example/samples/install_centos_7_payload_minimal.json ) in order to maintain consistency with what is documented.  Until that change was made, the documentation and what was supported got out of sync quickly.  

@johren, is it possible to go back to leveraging the payload from the examples directory?

jimturnquist:
Closing:  @johren

PR #818, opened 2017-06-26
Use wget to download the files to speed up tests

sunnyqianzhang:
Before the  code change, it took nearly half an hour to download on-http static files from the website. With the code change, the time is reduced to around 700s.  Also the cases with problem is fixed in the PR. @nortonluo

### DIFF
--- a/test/tests/imageserver/test_os_image_service.py
+++ b/test/tests/imageserver/test_os_image_service.py
@@ -38,10 +38,10 @@ class test_os_image_service(fit_common.unittest.TestCase):
         try:
             if os.path.exists(mountpoint) is False:
                 command = "mkdir " + mountpoint,
-                subprocess.check_output(command,shell=True)
+                subprocess.check_output(command, shell=True)
             # use fuseiso tool to mount without root privilege. user need install fuseiso before the test.
             command = "fuseiso " + file_name + " " + mountpoint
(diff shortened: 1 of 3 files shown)


test/tests/imageserver/test_os_image_service.py
```diff
@@ -322,26 +320,22 @@ def test_create_os_repo_from_store(self):
             self.assertTrue(self._mount_local_os_repo(file_name, os_name), "Could not mount ISO")
             self.assertTrue(self._compare_repo(os_name, os_version), "Fileserver compare failed!")
             self._release(file_name, os_name)
-        self.test_delete_all_images()
-
+    """
```
nortonluo: Annotate These test?

test/config/imageservice_config.json
```diff
@@ -5,7 +5,7 @@
         "osname": "CentOS",
         "version": "7.0",
         "linktype": "http",
-        "url": "http://buildlogs.centos.org/rolling/7/isos/x86_64/CentOS-7-x86_64-DVD.iso"
+        "url": "http://10.62.59.209/static/images/ubuntu-16.04-server-amd64.iso"
```
nortonluo: I think we can use another imageservice_config.json  instead of this one in github. There is no need to modify the original on in github

PengTian0:
:+1:

PR #820, opened 2017-06-27
rac-5340 Implement the immediate and  delayed rediscovery test for FIT

manfrednde:
- Add a single function to get IPMI user from node catalog
- Added the test to preload unidentified compute sku


derrickostertag:
@jimturnquist

PR #821, opened 2017-06-29
RAC-5129 Removed all references to CIT USER_AUTH_PORT

gpaulos:
RAC-5129 Removed all referenced to CIT USER_AUTH_PORT


gpaulos:
@jimturnquist @patelb10 @hohene

PR #824, opened 2017-07-03
RAC-4091 remove test of obmsetting from rackhd_stack_init.py

patelb10:
This PR is followup on 
https://github.com/RackHD/RackHD/pull/822
we added utility autoCreateObm:true  in config file. This will remove test of apply obm setting from stack_init


patelb10:
@gpaulos @jimturnquist @hohene

PR #825, opened 2017-07-10
RAC-5603 update sel logs test to exclude empty 'message' field

gpaulos:
We have determined that an empty SEL 'message' field is normal under certain circumstances. Test was asserting on an empty message, which is a false failure.

In the event of an empty SEL 'message' field, the test will now issue an INFO message containing SensorType and SensorNumber rather than an assert.


gpaulos:
Paired with Beena
@jimturnquist @patelb10 @hohene @johren

PR #826, opened 2017-07-11
RAC-5604 Centos 6.5 maximum payload for regression test

patelb10:
Maximum payload file for CentOS 6.5 OS install in regression testing for Quanta T41.
Paired with @gpaulos


patelb10:
@jimturnquist @hohene

PR #829, opened 2017-07-13
RAC-5635: Add docker hub dummy credentials

derrickostertag:
Add docker hub dummy credentials in credentials_default.json
@manfrednde @jimturnquist


manfrednde:
+1

PR #836, opened 2017-07-19
RAC-5253 support on-web-ui CI/CD

changev:
add on-web-ui build release stuffs.

@panpan0000 @PengTian0


debian/control
```diff
@@ -20,6 +20,7 @@ Depends: ${misc:Depends},
          on-syslog,
          on-taskgraph,
          on-tftp,
+         on-web-ui,
```
panpan0000: as we talk, I think rackhd.deb does NOT depend on independent services?

PR #842, opened 2017-07-28
RAC-4832: RackHD password reset tests

diontje:
RAC-4832: RackHD password reset tests
AutoTest refactored for Fit
Tests the ability for each user type (Admin, ReadOnly, Operator) to change/reset their password based on their permissions.

### DIFF
--- a/test/tests/security/test_api_security_passwd_resets.py
+++ b/test/tests/security/test_api_security_passwd_resets.py
@@ -14,7 +14,9 @@ Copyright 2015, DellEMC LLC
                In both cases, all user roles will have established accounts in both endpoints and
                will possess tokens for role base access and permission.
 
-           EX: python run_tests.py -stack 2 -config dellemc-test/config-mn/ -test tests/security/test_api_security_passwd_resets.py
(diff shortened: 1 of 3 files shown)


test/tests/security/test_api_security_passwd_resets.py
```diff
@@ -0,0 +1,133 @@
+'''
+Copyright 2015, DellEMC LLC
+
```
derrickostertag: Copyright incorrect

test/tests/security/test_api_security_passwd_resets.py
```diff
@@ -0,0 +1,133 @@
...
+logs = flogging.get_loggers()
+
+
+# Helper functions
+def createUsername():
+    return exrex.getone('[a-zA-Z]{1}[a-zA-Z0-9._\-]{1,}')
+
+
+@attr(regression=False, smoke=True, security=True)
+class TestCase01(fit_common.unittest.TestCase):
```
derrickostertag: Change **regression=false** to **regression=True**

This is a good regression test.

PR #843, opened 2017-07-28
RAC-5752 - Correct tests to support TaskCollection compliance changes

larry-dean:
* This change is compatible the current TaskCollection


larry-dean:
@jimturnquist

PR #846, opened 2017-08-02
add script to run FIT

PengTian0:
Add script to run FIT.
The steps of the script:
1. set up config for FIT
2. create virtual environment 
3. run test and collect test report under workspace/xunit-reports

The PR is the first step of job : refactor pipeline scripts.
@panpan0000 @changev @anhou @iceiilin

### DIFF
--- a/test/runFIT.sh
+++ b/test/runFIT.sh
@@ -1,0 +1,186 @@
+#!/bin/bash -e
+
+############################################
+# Global variable
+############################################
+RACKHD_TEST_DIR="$( cd -P "$( dirname "$0" )" && pwd )"
+
+#############################################
+#
+# Usage
+############################################
+Usage(){
+    set +x
+    echo "Function: This script is used to set up environment for FIT and run FIT."
+    echo "Usage: $0 [OPTIONS]"
+    echo "  OPTIONS:"
+    echo "    Mandatory options:"
+    echo "      -p, --PASSWORD: password of current user which is used for FIT to log into host by, it's required."
+    echo "    Optional options:"
+    echo "      -g, --TEST_GROUP: test group of FIT, such as imageservice, smoke"
+    echo "      -s, --TEST_STACK: target test stack of FIT, such as docker, vagrant..."
+    echo "      -v, --TEST_LOG_LEVEL: log level of FIT, 0 is the least verbose for log output , 9 is full on logging including info/debug etc. A good mix of console/file output is at level 4 or 5."
+    echo "      -w, --WORKSPACE: The directory of workspace( where the test report will be )"
+    echo "      -e, --TEST_EXTRA_OPTIONS: The extra options of FIT, such as '-extra myconfig.json -config my-dir -nodeid abcd'"
+    set -x
+}
+
+
+#############################################
+#
+#  Create the virtual env for FIT  
+#
+############################################
+setupVirtualEnv(){
+    pushd ${RACKHD_TEST_DIR}
+    virtual_env_name=FIT
+    rm -rf .venv/$virtual_env_name
+    ./mkenv.sh $virtual_env_name
+    source myenv_$virtual_env_name
+    popd
+}
+
+deactivateVirtualEnv(){
+    deactivate
+}
+
+####################################
+#
+# 1. Modify FIT config files , to  using actual DHCP Host IP instead of 172.31.128.1
+#
+##################################
+setupTestsConfig(){
+    echo "SetupTestsConfig ...replace the 172.31.128.1 IP in test configs with actual DHCP port IP"
+    RACKHD_DHCP_HOST_IP=$(ifconfig | awk '/inet addr/{print substr($2,6)}' |grep 172.31.128)
+    if [ "$RACKHD_DHCP_HOST_IP" == "" ]; then
+         echo "[Error] There should be a NIC with 172.31.128.xxx IP in your OS."
+         exit -2
+    fi
+    pushd ${RACKHD_TEST_DIR}/config
+    sed -i "s/\"username\": \"vagrant\"/\"username\": \"${USER}\"/g" credentials_default.json
+    sed -i "s/\"password\": \"vagrant\"/\"password\": \"${PASSWORD}\"/g" credentials_default.json
+    popd
+    pushd ${RACKHD_TEST_DIR}
+    find ./ -type f -exec sed -i -e "s/172.31.128.1/${RACKHD_DHCP_HOST_IP}/g" {} \;
+    popd
+}
+
+####################################
+#
+# Collect the test report
+#
+##################################
+collectTestReport()
+{
+    if [ -n "$WORKSPACE" ]; then
+        mkdir -p ${WORKSPACE}/xunit-reports
+        mv ${RACKHD_TEST_DIR}/*.xml ${WORKSPACE}/xunit-reports
+        echo "Test reports are xml files, under here: ${WORKSPACE}/xunit-reports"
+    else
+        echo "Test reports are xml files, under here: ${RACKHD_TEST_DIR}"
+    fi
+}
+
+
+####################################
+#
+# Start to run FIT tests
+#
+##################################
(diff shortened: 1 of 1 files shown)


test/runFIT.sh
```diff
@@ -0,0 +1,164 @@
...
+    popd
+}
+
+####################################
+#
+# 1. Modify FIT config files , to  using actual DHCP Host IP instead of 172.31.128.1
+#
+##################################
+setupTestsConfig(){
+    echo "SetupTestsConfig ...replace the 172.31.128.1 IP in test configs with actual DHCP port IP"
```
panpan0000: Talked with @PengTian0, it's not elegant to replace every 172.31.128.1  . And it's possible that the RackHD interface to FIT is not 172.31.128.xx (if running in diff docker/vm)
and  there's a story RAC-5418 to track the enhancement for FIT, to improve the hard code rackhd IP.

test/runFIT.sh
```diff
@@ -0,0 +1,164 @@
...
+#
+# Start to run FIT tests
+#
+##################################
+runFIT() {
+    set +e
+    netstat -ntlp
+    pushd ${RACKHD_TEST_DIR}
+    echo "########### Run FIT Smoke Test #############"
+    python run_tests.py ${TEST_GROUP} -stack docker_local_run -v 4 -xunit
```
panpan0000: because the stack = docker_local_run
Please clear state the comments in the usage  :   this script only supports test RackHD running in docker with specific config.json .

test/runFIT.sh
```diff
@@ -0,0 +1,164 @@
+#!/bin/bash -ex
+
+############################################
+# Global variable
+############################################
+RACKHD_TEST_DIR="$( cd -P "$( dirname "$0" )" && pwd )"
```
panpan0000: why only "dirname $0" not working ?

test/runFIT.sh
```diff
@@ -0,0 +1,164 @@
...
+#
+# Usage
+############################################
+Usage(){
+    set +x
+    echo "Function: This script is used to set up environment for FIT and run FIT."
+    echo "Usage: $0 [OPTIONS]"
+    echo "  OPTIONS:"
+    echo "    Mandatory options:"
+    echo "      -w, --WORKSPACE: The directory of workspace( where the code will be cloned to and staging folder), it's required"
```
panpan0000: if put runFIT.sh as a common script,  the WORKSPACE is not mandatory .. because what you use WORKSPACE is just to copy the xunit reports.

test/runFIT.sh
PengTian0: If go to the directory RackHD/test and run ./runFIT.sh, the ```dirname $0``` returns ```.````

test/runFIT.sh
PengTian0: Thanks, fixed.

test/runFIT.sh
PengTian0: Thanks, fixed.

PengTian0:
test this please

hohene:
There are a few more options that may be useful here - TEST_GROUP is the group name, where we also may need -extra XXXX for specific config files.  Or -config XXX to point at a different config directory.  (that may be less likely, but useful if we want to make this a bit more flexible and able to call different test envs).  
You could make just an extra "OPTIONs" variable where we fill in a string of options so you don't have to cover all.  ie.  OPTIONS = "-extra myconfig.json -config my-dir -nodeid abcd"  then just pass thru to run_tests.py.

test/runFIT.sh
```diff
@@ -0,0 +1,176 @@
...
+    set +x
+    echo "Function: This script is used to set up environment for FIT and run FIT."
+    echo "Usage: $0 [OPTIONS]"
+    echo "  OPTIONS:"
+    echo "    Mandatory options:"
+    echo "      -p, --PASSWORD: password of current user which is used for FIT to log into host by, it's required."
+    echo "    Optional options:"
+    echo "      -g, --TEST_GROUP: test group of FIT, such as imageservice, smoke"
+    echo "      -s, --TEST_STACK: target test stack of FIT, such as docker, vagrant..."
+    echo "      -v, --TEST_LOG_LEVEL: log level of FIT, 0 is the smallest, 9 is the biggest"
```
panpan0000: 0 is the most verbose for log output , 9 for most succinct .

PengTian0: Thanks, fixed.

PengTian0:
@hohene , thanks for your review.
I already add a extra option according to your comments.

test/runFIT.sh
```diff
@@ -0,0 +1,182 @@
...
+    echo "Function: This script is used to set up environment for FIT and run FIT."
+    echo "Usage: $0 [OPTIONS]"
+    echo "  OPTIONS:"
+    echo "    Mandatory options:"
+    echo "      -p, --PASSWORD: password of current user which is used for FIT to log into host by, it's required."
+    echo "    Optional options:"
+    echo "      -g, --TEST_GROUP: test group of FIT, such as imageservice, smoke"
+    echo "      -s, --TEST_STACK: target test stack of FIT, such as docker, vagrant..."
+    echo "      -v, --TEST_LOG_LEVEL: log level of FIT, 0 is the most verbose for log output , 9 for most succinct"
+    echo"       -w, --WORKSPACE: The directory of workspace( where the test report will be )"
```
sunnyqianzhang: Minor error here, need one space between echo and "

test/runFIT.sh
```diff
@@ -0,0 +1,182 @@
...
+####################################
+#
+# Collect the test report
+#
+##################################
+collectTestReport()
+{
+    if [ -n "$WORKSPACE" ]; then
+        mkdir -p ${WORKSPACE}/xunit-reports
+        cp ${RACKHD_TEST_DIR}/*.xml ${WORKSPACE}/xunit-reports
```
sunnyqianzhang: If we moved *xml instead of copying *xml,  would it make the workspace a little cleaner?

test/runFIT.sh
```diff
@@ -0,0 +1,182 @@
...
+
+##############################################
+#
+# Set up test environment and run test
+#
+#############################################
+runTests(){
+    setupTestsConfig
+    setupVirtualEnv
+    runFIT " --sm-amqp-use-user guest"
```
sunnyqianzhang: what if we remove " --sm-amqp-use-user guest"? Seems  it isn't used anywhere

PengTian0: It's used by RackHD FIT test.

test/runFIT.sh
PengTian0: Thanks, fixed.

test/runFIT.sh
PengTian0: Thanks, fixed.

test/runFIT.sh
```diff
@@ -0,0 +1,182 @@
...
+    set +x
+    echo "Function: This script is used to set up environment for FIT and run FIT."
+    echo "Usage: $0 [OPTIONS]"
+    echo "  OPTIONS:"
+    echo "    Mandatory options:"
+    echo "      -p, --PASSWORD: password of current user which is used for FIT to log into host by, it's required."
+    echo "    Optional options:"
+    echo "      -g, --TEST_GROUP: test group of FIT, such as imageservice, smoke"
+    echo "      -s, --TEST_STACK: target test stack of FIT, such as docker, vagrant..."
+    echo "      -v, --TEST_LOG_LEVEL: log level of FIT, 0 is the most verbose for log output , 9 for most succinct"
```
hohene: Actually 0 is the least verbose.  you will get no output other than pass/fail with log level 0.
9 is full on logging including info/debug etc.  A good mix of console/file output is at level 4  or 5. Note that logs files are also being auto created into the log_output directory for every run.  We should look to copy those off at some point.

sunnyqianzhang:
:+1:

PengTian0:
test this please

PR #848, opened 2017-08-02
WIP:Make taskgraph stand alone service

tannoa2:


docker/monorail/config.json
```diff
@@ -40,6 +32,13 @@
           "routers": "southbound-api-router"
       }
     ],
+    "taskGraphEndpoint": {
+    "address": "0.0.0.0",
```
anhou: so the intention is removing southbound-api-router and move to taskGraphEndpoint?

geoff-reid: Correct

lanchongyizu:
@tannoa2 could you resolve the conflict?

PR #860, opened 2017-08-16
fix for empty Created field in SEL records

geoff-reid:
This fixes a failure when running FIT on an infrasim virtual Dell r630.  
@RackHD/committers


johren:
Looks okay to me.  Will approve after hound violation is addressed.

PR #861, opened 2017-08-17
Remove footprint benchmark from test repo

iceiilin:
**Background**
As @gpaulos suggested, test cases used in benchmark is obsolete as it uses API 1.1. And from user's experience, they would prefer to use benchmark as an independent tool to measure RackHD footprint, rather than measuring its performance only in some specific test cases.

Besides, benchmark tool is relying on the configurations set for CIT previously, which is out of date as well when CIT is gradually deprecated.

**Solution**
This PR moves benchmark out from RackHD/test. It will be in on-tools.


iceiilin:
test this please

PR #862, opened 2017-08-18
Fix: the step replacing 172.31.128.1 with 172.31.128.250 is not needed by ova

PengTian0:
The script runFIT.sh will be used by docker, ova, source_code and so on.
The step to replacing 172.31.128.1 with 172.31.128.250 is not needed by ova.
So the PR removes it

@panpan0000 @anhou @iceiilin

### DIFF
--- a/test/runFIT.sh
+++ b/test/runFIT.sh
@@ -51,17 +51,9 @@ deactivateVirtualEnv(){
 ##################################
 setupTestsConfig(){
     echo "SetupTestsConfig ...replace the 172.31.128.1 IP in test configs with actual DHCP port IP"
-    RACKHD_DHCP_HOST_IP=$(ifconfig | awk '/inet addr/{print substr($2,6)}' |grep 172.31.128)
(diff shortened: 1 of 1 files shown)


PengTian0:
test this please

panpan0000:
But for docker/source_code, how can you handle the IP difference ?
say, the host runs with 172.31.128.250?

PengTian0:
It's handled in share library FIT: 
https://github.com/RackHD/on-build-config/blob/8f0a1663516dd540e5df2053ea21172f1fcb7c27/src/pipeline/fit/FIT.groovy#L3

If the test needs to replace the ip, the test should call the function "configControlInterfaceIp" before calling function "run"

panpan0000:
talking with @PengTian0 ,  it was not elegant to strictly hard-code the southbound IP in FIT configs/codes, there should be a flexibility to pass in the southbound IP to FIT as a parameter or config file..etc.
in another word, the "sed" operation was  also a workaround.. 
so either way, moving the "sed" out from fit.sh to fit.groovy doesn't matter.
the key is to enhance FIT..
so +1

PR #869, opened 2017-08-23
Add alpha concourse testing

keedya:


keedya:
@hohene @geoff-reid @johren

PR #871, opened 2017-08-24
move fit config modification to on-build-config lib

changev:


changev:
test this please

changev:
test this please

panpan0000:
@PengTian0  to review

packer/ansible/roles/setuptools/tasks/main.yml
```diff
@@ -4,3 +4,6 @@
 
 - pip: name=setuptools extra_args='--upgrade'
   sudo: yes
+
+- pip: name=virtualenv extra_args='--upgrade'
+  sudo: yes
```
PengTian0: Why upgrade virtualenv in this PR?

PengTian0:
Please update the ```Usage``` according to your change

packer/ansible/roles/setuptools/tasks/main.yml
changev: in ova venv version is too low.
And with such venv,  pip install  in ```vnc_record.sh``` will throw error.

PengTian0: I think it's better to modify it in PR for build_ova.sh

changev:
test this please

packer/ansible/roles/setuptools/tasks/main.yml
changev: but without the ova with upgraded virtualenv, current ova post test won't work.

one pr or two pr is both ok for me ~  but I think this is not a urgent topic, merge this PR and let feature branch work is the urgent affairs

panpan0000:
👍 

@iceiilin  to review

PR #881, opened 2017-08-31
RAC-5911 Added 404 response for rackhd manager serial interface

patelb10:
Paired with @pathakvaidehi2391 @jimturnquist


jimturnquist:
This Jenkins failure is OK.  The failure won't occur when running with the other half of RAC-5911.

PR #885, opened 2017-09-07
RAC-6078 refactor build ova for build template

changev:
support ova cache build refactor

@PengTian0 @panpan0000


packer/build_ova.sh
```diff
@@ -121,9 +127,12 @@ postProcess(){
   createChecksum
   gpgSigning
   chmod a=r "$OVA"*
-  if [ $TARGET_DIR != `pwd` ]; then
+  set +e
```
panpan0000: can you add comment here why set +e?
because it's possible that "TARGET_DIR can OUTPUT_PATH are the same location, and the mv will fail"

changev: no~ because when build ova there's no vmdk files

PR #886, opened 2017-09-08
RAC-6079: Refactor vagrant cache build shell script

PengTian0:
Implement a shell script build_ovf.sh for vagrant cache build pipeline.
@panpan0000 @changev @anhou @iceiilin


packer/box_handler.sh
```diff
@@ -0,0 +1,87 @@
...
+        echo "[ERROR] packer is not installed . please install it."
+        exit 5
+    fi
+    vagrant_path=$( which vagrant )
+    if [ -z "$vagrant_path" ];then
+        echo "[ERROR] vagrant is not installed . please install it."
+        exit 5
+    fi
+    jq_path=$(which jq)
+    if [ ! -n "$(which jq)" ] ; then
```
panpan0000: are you going to use   ```if [ -z $jq_path ] ; then``` like what you did before?

packer/build_ovf.sh
```diff
@@ -0,0 +1,113 @@
...
+    echo "      --TARGET_DIR: The target directory to put ovf and log file"
+}
+
+#########################################
+#
+#  build ovf
+#
+#########################################
+build(){
+    pushd $DIR
```
panpan0000: do you want to use $1/$2/$3 instead of global variable ..?

packer/build_ovf.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash -e
```
panpan0000: ```ovf``` is a standard packaging format . it's not only used for Virtualbox/Vagrant , but also for VMWare( but I believe you will have another script to build the cached/templated RackHD Vmware image )
 I think it's more proper to rename this script to build_base_vagrant_box.sh or something more clear ?

packer/box_handler.sh
PengTian0: Thanks, fixed.

PengTian0:
@panpan0000 Thanks.
Already rename the build_ovf.sh to "build_vagrant_base_box.sh" according to your comments.

PR #892, opened 2017-09-25
Added docker-compose pull command under README.md

ajeetraina:
Under RackHD for MacOS section, one need to first pull the docker images and then execute docker-compose up


panpan0000:
in theory, docker-compose up will do pull+up ?

ajeetraina:
If the --recursive flag is provided during the `git clone
https://github.com/rackhd/RackHD` , then it mightn't be necessary but if
not, then it fails to pull the overall repository.

PR #898, opened 2017-10-09
fix docker-compose to build dependencies with the right tag

keedya:
@geoff-reid  @tannoa2 @dalebremner @tgelter


tgelter:
Thanks @keedya !

PR #900, opened 2017-11-01
rac-6421-turndown-some-amqp-debug

stuart-stanley:
There are some debug messages in early AMQP that are only usefull if debugging the amqp code itself. Set their debug level to 8 (not normally visible)

Do not merge until the deliberate test failure is removed!


RackHDCI:
The Copyright Check has failed!
The files listed below do not have a valid copyright statement:
- test/stream-monitor/.flake8

PR #902, opened 2017-11-06
Updates to support running smoke against real hardware

hohene:
Updates for bare-metal regression stacks.
- added separate init script rackhd_stack_init_baremetal.py
  - verifies the hardware inventory before smoke starts
  - skips any power on reset of the nodes before smoke starts
- update sel_alert_poller_tests to skip performing full SEL injection tests against real hardware
- update system tests to skip power-reset type tests
- updated coreos repository in install_coreos_minimum.json
- added stackType and stack_nodes into stack_config.json for running manually


hohene:
@keedya @johren

PR #903, opened 2017-11-10
Fix the issue of visiting SMI firmware update service

lanchongyizu:
**Background**
smi-service API could be invoked through each individual smi-service or smi-service-gateway which could be controlled by config file. During development of previous stories, we found it was failed to invoke API through gateway but successful through individual smi-services. We need to resolve this problem.
https://rackhd.atlassian.net/browse/RAC-6424

**Done**
Add extra_hosts: service-registry for consul.

**Reviewers**
@rahmanmuhamad  @RackHD/corecommitters @nortonluo

### DIFF
--- a/docker/dell/docker-compose.yml
+++ b/docker/dell/docker-compose.yml
@@ -1,3 +1,5 @@
+# Copyright 2017, Dell EMC, Inc.
+
 version: '2'
 
 services:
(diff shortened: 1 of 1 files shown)


RackHDCI:
The Copyright Check has failed!
The files listed below do not have a valid copyright statement:
- docker/dell/docker-compose.yml

PR #904, opened 2017-11-10
Changed reset_type to ResetType to match Redfish standard

keedya:
depends on: https://github.com/RackHD/on-http/pull/809


keedya:
@hohene @dalebremner @johren @geoff-reid  to review

PR #909, opened 2017-11-20
RAC-6412: fix original FIT test cases for ucs-service

PengTian0:
1. Add ucs option to "docker_local_run" which is the stack used to run FIT.
2. Fix error : the api /catalog?identifier=switch-B is not supported.

@pengz1 @anhou @geoff-reid @tannoa2


PengTian0:
Jenkins failure can be ignore because it's disabled now.

anhou:
ignore Jenkins result and merge

PR #910, opened 2017-11-22
RAC-6490: add fit test cases for ucs-service poller

PengTian0:
Add test cases for ucs-service poller

@anhou @pengz1 @geoff-reid


pengz1:
@bbcyyb

PR #911, opened 2017-11-24
Move UCS API test to ucs-service-repo, remove it from FIT

pengz1:
As independent service, UCS-service should have its own function test. Thus API test will be done in UCS-service repo, remove it from FIT.


pengz1:
@PengTian0 @anhou

PR #916, opened 2017-12-05
[RAC-6407] fix ucs debian fit bug

HQebupt:
During the ucs-service debian fit test, there is no ucs.fan poller in the test environment here.
So the fit test has one failer.

```
curl -k GET   'https://1.2.3.4:7080/pollers?identifier=sys%2Fchassis-3&classIds=equipmentFanStats'   -H 'ucs-host: 10.62.xx.xx'   -H 'ucs-password: ****'   -H 'ucs-user: ucspe'

{
  "equipmentFanStats": []
} 
```
Can we remove the if judgement ?

@Kevin Yu @pengz1


pengz1:
You should clarify two things: fan poller doesn't exist in UCSPE or it doesn't exist in this kind of nodes.

If it only doesn't exist in UCSPE but exists in bare mental nodes, you should consider skip it instead of removing it in case bare mental tests require it.

HQebupt:
It makes sense. Thanks. I will skip the test, and close the issue.

panpan0000:
:+0

PR #920, opened 2017-12-14
Add retry to testing ucs poller data cache

PengTian0:
@pengz1 @anhou


test/tests/ucs/test_rackhd20_ucs_pollers.py
```diff
@@ -193,6 +194,13 @@ def test_api_20_verify_poller_default_cache(self):
                 poller_id = poller_dict[poller]["poller_id"]
                 poller_data = test_api_utils.get_poller_data_by_id(poller_id)
                 poll_len = len(poller_data)
+                timeout = 120
```
pengz1: For poller it would be better use 60 seconds.

